PR #3, opened 2025-07-22
fix: add ios permissions check

vzhovnitsky:
Reason: On iOS, if a user cancels the camera permission request when CameraView.launchScanner() is called, the system will never prompt for permissions again. This leaves users unable to use the QR code scanner functionality without manually going to Settings > Privacy & Security > Camera to re-enable permissions.


ex3ndr:
Thanks!

PR #6, opened 2025-07-28
fix: android barcode scanner

dvlkv:


ex3ndr:
I think it should be iOS only, since we have webapp too

dvlkv:
Technically dismissScanner is not called on web by the implementation, but I updated PR for the clarification
Screenshot 2025-07-29 at 12 12 23 AM

ex3ndr:
Thanks for your contribution!

Issue #34, opened 2025-08-22
button for "Enter URL manually" does not work


turbocrime:
the button for "Enter URL manually" to pair with the desktop cli does nothing on android

this affects both locations on the onboarding screen, and the settings menu

this is important to me because without play services, the barcode scanner also does not work: https://github.com/slopus/happy/issues/36

PR #35, opened 2025-08-22
fix: `Modal.prompt` should render on android platform

turbocrime:
when the `ModalProvider` attempts to display a modal on android, it renders nothing, because it refuses to render unless `Platform.OS === 'web'`

correcting this fixes https://github.com/slopus/happy/issues/34

### DIFF
--- a/sources/modal/ModalProvider.tsx
+++ b/sources/modal/ModalProvider.tsx
@@ -1,5 +1,4 @@
 import React, { createContext, useContext, useState, useEffect, useCallback } from 'react';
-import { View, Platform } from 'react-native';
 import { ModalState, ModalConfig, ModalContextValue } from './types';
 import { Modal } from './ModalManager';
 import { WebAlertModal } from './components/WebAlertModal';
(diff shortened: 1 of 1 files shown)


ex3ndr:
I have seen similar problem too, we have a dedicated prompt alert and custom modal should be removed altogether, can you try this instead?

turbocrime:
i'm not sure what you mean. this PR fixes issue #34, which was my goal.

the modal config is still `{ type: 'prompt' }` and nothing is changed except a condition in `ModalProvider` which prevented rendering the component.

without this PR, the behavior of `Modal.prompt` is currently broken on android.

turbocrime:
actually, i believe the rendering of any modal via `ModalProvider` is currently broken on android, this was just the first modal i encountered.

turbocrime:
it seems like most modals will render with `Alert.alert` or `Alert.confirm`, so this may not surface anywhere else. but `Alert.prompt` is not available on android, so there must be some kind of component solution.

ex3ndr:
Oof didn't know it is not available on android!


Steve Korshakov

Sent via ***@***.***>


On Fri, Aug 22, 2025 at 3:51 PM, the letter L ***@***.******@***.***>> wrote:
[https://avatars.githubusercontent.com/u/134443988?s=20&v=4]turbocrime left a comment (slopus/happy#35)<https://github.com/slopus/happy/pull/35#issuecomment-3215879958>

it seems like most modals will render with Alert.alert or Alert.confirm, so this may not surface anywhere else. but Alert.prompt is not available on android, so there must be some kind of component solution.

PR #37, opened 2025-08-22
fix: staged file paths were being missing first character

kamal:
## Before
Screenshot 2025-08-22 at 4 06 43 PM

## After

Screenshot 2025-08-22 at 4 05 51 PM

### DIFF
--- a/sources/sync/gitStatusFiles.ts
+++ b/sources/sync/gitStatusFiles.ts
@@ -117,7 +117,7 @@ function parseGitStatusFiles(
         
         const indexStatus = line[0]; // Staged status
         const workingStatus = line[1]; // Working directory status
-        const fileName = line.slice(3); // File path
+        const fileName = line.slice(2).trim(); // File path
 
         // Handle renamed files (format: "R  old_name -> new_name")
         let filePath = fileName;


GrocerPublishAgent:
What is your git version? I'm trying to understand why this is working differently on your computer. On my laptop with Git 2.35 I need to slice(3). I don't doubt your system needs slice(2), I'm surprised.

For example, when I run the same command that Happy is running, this is what I get
```
user@work expo-mobile-app % git status --porcelain
M  PRIVACY.md
```
So I need to take 3 characters away to get the filename. But your system clearly only needs to remove 2.

Can you stage a file in a random git repository and then run `git status --porcelain` and show us the output. I am wondering if we need to make the parser a little bit more robust if there are various formats running around.

kamal:
Here's what my git outputs

```
kamal@kmahyuddin-mbp happy % git status --porcelain
 M sources/components/GitStatusBadge.tsx
kamal@kmahyuddin-mbp happy % git --version
git version 2.47.1
```

I did some digging and it looks like the stable formatting that's machine parseable is to use `git status --porcelain=v2`. If you want, I can rework the parser to work with this format.

kamal:
looks like more robust git parsing is in main now. i can close out this PR.

PR #53, opened 2025-08-26
Let the voice assistant receive incoming messages from Claude

lava:
The voice assistant was only being notified about permission requests but not about new incoming messages from Claude. Added logic in the applyMessages function to detect new agent-text messages and notify the voice assistant via sendTextMessage.

This enables the voice assistant to read out Claude's responses when they arrive.

### DIFF
--- a/sources/sync/storage.ts
+++ b/sources/sync/storage.ts
@@ -2,7 +2,7 @@ import { create } from "zustand";
 import { useShallow } from 'zustand/react/shallow'
 import { Session, Machine, GitStatus } from "./storageTypes";
 import { createReducer, reducer, ReducerState } from "./reducer/reducer";
-import { Message } from "./typesMessage";
+import { Message, AgentTextMessage } from "./typesMessage";
 import { NormalizedMessage } from "./typesRaw";
 import { isMachineOnline } from '@/utils/machineUtils';
 import { applySettings, Settings } from "./settings";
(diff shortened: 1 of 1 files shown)


ex3ndr:
We already adding them to the context, but we are not using "sendText" since it triggers a response, we think only when result is received it should trigger an answer from voice assistant, anything in between user can just ask and have an access to the messages. What was the problem in your case?

lava:
Maybe I misunderstood the current behavior, but I was experiencing a lot of conversations like this:

```
User (voice): Ask claude to figure out when this file was last changed.
Voice Assistant (text): Figure out when this file was last changed.
Voice Assistant (voice): I've asked claude to figure out when the file was changed, i'll get back to you with the response
...
Claude (text): The file was changed yesterday.
...
User (voice): So what was the answer?
Voice Assistant (voice): I haven't received a response yet, i'll notify you when claude has responded.
```

Ie., the voice assistant was not aware of new chat messages that arrived after starting the voice session. If  I understand you correctly, the expected behavior would be that the assistant doesnt pro-actively respond with the changes but should be able to produce them after the explicit question?

After applying this PR, the voice assistant was able to give the answer as it arrived.

ex3ndr:
Try to update to the latest CLI, we are sending extra event from it to notify when job was finished, it works only when controlling from the app tho, so switch to remote control to make it work on latest CLI.

lava:
@ex3ndr After updating to today's versions of the CLI and web interface, I cannot reproduce the issue anymore, so I'm closing this PR. Thanks for your pointer!

PR #74, opened 2025-09-03
feat: add Simplified Chinese language pack

pyflmcp:


GrocerPublishAgent:
Thank you so much for taking the time to translate the app to Chinese! 非常感谢！🙏

I noticed you used both 您 (formal) and 你 (informal) pronouns in different places, and this is actually a great learning opportunity for me! My Chinese is pretty limited (我的中文水平... 半桶水晃荡晃荡的 😅), so I'm curious about the pronoun choices.

From apps I've used like WeChat, I mostly see 你 used consistently, but I'm wondering if there's a nuance I'm missing? Like, did you intentionally use 您 in certain contexts (maybe security warnings or account settings) for a specific reason? 

As a developer tool, I was thinking we'd want to sound friendly and approachable throughout, but I realize there might be considerations I don't understand. For example:
- Do certain types of messages (like security warnings) feel better with 您?

I've also wondered if we could just avoid the pronoun question entirely in some places:
- "确定要登出吗？" vs "你确定要登出吗？"
- "请备份密钥" vs "请备份你的密钥"

I'm just trying to learn so I can better understand how to approach Chinese translations in the future! Would love to hear your thoughts on this if you have time.

Thanks again for the contribution!

PR #127, opened 2025-09-20
fix: allow hiding inactive sessions from view as an experimental setting

GrocerPublishAgent:
https://github.com/beachead requested being able to hide the inactive sessions from view


ex3ndr:
1) this should be sync setting not local one
2) should be scoped to a view not alter logic in storage

PR #135, opened 2025-09-23
Use wizard for new session

denysvitali:
## Problem Statement

The "Create Session" is broken in many parts. On mobile, the agent config isn't appearing and on desktop (web) the path settings aren't applied.

## Solution

With this change, we introduce a "Create Session" wizard that guides the user towards creating a session.

## Result

image

image

image

image

image

image

---

Note: Vibe Coded w/ Claude Opus 4.1 / Claude Sonnet 4.0


sabirmgd:
thanks @denysvitali this is amazing , kindly let's get this merged, guys.
I had to start all sessions from terminal and is very weird. 

image

ex3ndr:
Someone will look into it!

bra1nDump:
@denysvitali wondering if we can compact this a bit. I think the current launcher works a little better if you start sessions with the same configuration often.

Maybe we can focus on the functional changes of this PR first and experiment with the UX separately.

sabirmgd:
guys can we merge this :/

huntsyea:
Why have we not merged this? It is needed.

kierr:
LGTFM

nelsonblaha:
I did a targeted fix of just the machine selection at https://github.com/slopus/happy/pull/253 before seeing this

tiann:
I think this is over‑engineered. An action that originally could be completed with two taps on a single screen now requires switching between so many screens, which is very inconvenient.

ahundt:
Thanks for the inspiration, @denysvitali I started with your wizard and improved upon the ui. I agree with @tiann in part so I tried to make it much easier to use.

This new wizard has one tap selectable profiles to start typing the first message of your session: https://github.com/slopus/happy/pull/272

PR #253, opened 2025-12-02
Fix machine selection bug in web container

nelsonblaha:
## Problem

When using happy-web to start a new session, clicking the machine name to change it and then selecting a different machine would cause the selection to revert back to the original machine.

## Root Cause

Module-level callback pattern had race conditions in web environments where `router.back()` could complete before state updates propagated.

## Solution

Pass `selectedMachineId` explicitly via route params when navigating back from machine picker. This ensures the selection is guaranteed to be available when the component mounts.

## Changes

- Updated `sources/app/(app)/new/index.tsx` to read machine selection from route params
- Updated `sources/app/(app)/new/pick/machine.tsx` to pass machine ID via route params
- Updated `metro.config.js` to exclude test files from web bundle (prevents vitest error)
- Added `machineSelection.spec.ts` with 9 tests covering bug reproduction and fix validation

## Testing

- ✅ All 9 new tests passing
- ✅ Manually tested in browser - machine selection now works correctly
- ✅ No existing tests broken (same 32 pre-existing failures)

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -2,4 +2,14 @@ const { getDefaultConfig } = require("expo/metro-config");
 
 const config = getDefaultConfig(__dirname);
 
+// Exclude test files from bundling
+config.resolver = {
+  ...config.resolver,
+  sourceExts: [...(config.resolver?.sourceExts || [])],
+  blockList: [
+    /.*\.spec\.ts$/,
+    /.*\.test\.ts$/,
+  ]
+};
+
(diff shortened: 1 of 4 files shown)


rrnewton:
Nice, I've done a scan of the code but haven't tested the patch yet myself. When we get to monorepo + e2e testing it will be nice to have an e2e test for this kind of thing.

PR #262, opened 2025-12-03
fix: retry message fetch when session encryption not ready

### LINKED ISSUE #260: Messages don't load for new sessions due to race condition
## Summary

When a new session is started from the CLI, message history may not load on the mobile app due to a race condition between session encryption initialization and message fetching.

## Severity
High

## Affected
Sessions started from CLI that user quickly navigates to on mobile

## Root Cause

The `fetchMessages` function requires session encryption to be initialized (via `fetchSessions`), but if the user navigates to a session before `fetchSessions` completes, `fetchMessages` silently returns early without fetching messages. The `InvalidateSync` pattern only retries on thrown errors, not silent returns.

## Flow That Fails

```
CLI creates new session
       ↓
Server sends "new-session" WebSocket update
       ↓
Mobile receives update, calls sessionsSync.invalidate()
       ↓
User navigates to the session (e.g., from push notification)
       ↓
onSessionVisible(sessionId) is called
       ↓
fetchMessages() runs BUT encryption isn't ready yet
       ↓
getSessionEncryption(sessionId) returns NULL
       ↓
fetchMessages() returns early (silent, no error thrown)
       ↓
isLoaded stays FALSE, user sees loading spinner forever
       ↓
fetchSessions() eventually completes, sets up encryption
       ↓
But fetchMessages was already "completed" - no automatic retry
```

## Key Code Locations

- **`sources/sync/sync.ts:1402-1406`** - Early return when encryption not ready
- **`sources/sync/sync.ts:188-194`** - `onSessionVisible` creates the fetch sync
- **`sources/sync/encryption/encryption.ts:90-92`** - `getSessionEncryption` returns null if not initialized
- **`sources/utils/sync.ts`** - `InvalidateSync` uses `backoff` for retries, but only on thrown errors

## Reproduction

1. Start a new Claude Code session from CLI
2. Quickly open the session on mobile (e.g., via push notification)
3. Messages may not load, showing loading spinner indefinitely

## Potential Fixes

1. **Throw an error instead of returning** - Change the early return to throw, triggering `backoff` retry
2. **Wait for session encryption** - Add `await sessionsSync.awaitQueue()` before fetching messages
3. **Re-invalidate after sessions load** - After `fetchSessions` completes, re-trigger `fetchMessages` for visited sessions
4. **Queue message fetches** - If encryption isn't ready, queue the fetch to run after initialization

kmizzi:
## Summary

Fixes the race condition where messages don't load for new sessions when the user quickly navigates to them.

- When a new session is created and user navigates to it quickly, `fetchMessages()` may run before `fetchSessions()` completes
- Previously, if session encryption wasn't ready, `fetchMessages()` silently returned early
- This left the session stuck with `isLoaded: false` forever (loading spinner)
- Now we throw an error, which triggers the `backoff` retry mechanism in `InvalidateSync`
- Messages will load once session encryption becomes available

## Test plan

- [x] TypeScript type checking passes (`yarn typecheck`)
- [x] Existing sync-related tests pass
- [ ] Manual test: Start session from CLI, quickly open on mobile - messages should load

Fixes #260

### DIFF
--- a/sources/sync/sync.ts
+++ b/sources/sync/sync.ts
@@ -1398,11 +1398,12 @@ class Sync {
     private fetchMessages = async (sessionId: string) => {
         log.log(`💬 fetchMessages starting for session ${sessionId} - acquiring lock`);
 
-        // Get encryption
+        // Get encryption - may not be ready yet if session was just created
+        // Throwing an error triggers backoff retry in InvalidateSync
         const encryption = this.encryption.getSessionEncryption(sessionId);
-        if (!encryption) { // Should never happen
-            console.error(`Session ${sessionId} not found`);
-            return;
+        if (!encryption) {
+            log.log(`💬 fetchMessages: Session encryption not ready for ${sessionId}, will retry`);
+            throw new Error(`Session encryption not ready for ${sessionId}`);
         }
 
         // Request


bra1nDump:
Thanks @kmizzi! Merging this and #263.

Sorry for the long review time - we're resuming active work on the project. Aiming to roll out a new version this weekend.

Contributions welcome! We're discussing plans in Discord: https://discord.gg/aVhXZfYh

PR #270, opened 2025-12-07
Fix first message not sent to model on new session creation

tiann:
### Symptoms
  When creating a new session with prefilled text, the first message appears in the UI but the agent never responds. User must copy and resend the message manually.

  When Claude starts the binary via spawn, it is more likely to be triggered because spawn is time-consuming.

  ### Root Cause
  Race condition between app sending message and agent WebSocket connection.

  ### Fix

  The app now waits for `agentStateVersion > 0` before sending the first message. This field starts at 0 when a session is created and increments when the CLI connects and calls `updateAgentState()`, serving as a reliable signal that the agent's WebSocket is ready to receive messages.

  ### Timeline Diagrams

  **Before fix (message lost)**
  ```
  App                              Server                      Agent
  ──────────────────────────────────────────────────────────────────────────────
  machineSpawnNewSession ────────► spawn daemon ─────────────► Agent starting…
  refreshSessions (session in DB)                                │ (connecting)
  sendMessage(input) ───────────► store in DB ✓                 │ not connected
                                │ broadcast via WebSocket ────► (no recipient)
  navigate to session                                               │
                                                                      ▼
                                                              Agent connects
                                                              updateAgentState()
                                                              (first message missed)
```

  **After fix (message delivered)**
  
  ```
  App                              Server                      Agent
  ──────────────────────────────────────────────────────────────────────────────
  machineSpawnNewSession ────────► spawn daemon ─────────────► Agent starting…
  refreshSessions (session in DB)                                │ (connecting)
    └─ agentStateVersion = 0                                     │
                                                                   │
  sendMessage(input) ──┐                                          │
                       │ waitForAgentReady()                      │
                       │  - watches agentStateVersion             │
                       │  - 10s timeout fallback                  │
                       │                                          ▼
                       │                                    WebSocket connects
                       │                                    updateAgentState()
                       │                              ◄──── agentStateVersion = 1
                       └─ resumes ───────────────► broadcast message ─────────► receives ✓
  navigate to session
  Agent responds
  ```
  
 ### Testing

  To reproduce the bug, a 5-second delay was added in `happy-cli/src/claude/runClaude.ts` before the WebSocket connects:

  ```typescript
  // DEBUG: Simulate slow CLI startup to reproduce first-message-lost bug
  // The delay must be BEFORE sessionSyncClient() because that's where WebSocket connects
  // Remove this after testing!
  await new Promise(resolve => setTimeout(resolve, 5000));

  // Create realtime session
  const session = api.sessionSyncClient(response);
```

#### Steps to verify:

1. Start daemon with the delay in place
2. Create a new session from mobile app (e.g., via spawn)
3. Send a message immediately after session creation

#### Expected behavior:

- Without fix: Message appears in UI but agent never responds (lost during 5s window)
- With fix: App waits ~5s for agentStateVersion > 0, then message is delivered and agent responds

### DIFF
--- a/sources/sync/sync.ts
+++ b/sources/sync/sync.ts
@@ -41,6 +41,8 @@ import { UserProfile } from './friendTypes';
 import { initializeTodoSync } from '../-zen/model/ops';
 
 class Sync {
+    // Spawned agents (especially in spawn mode) can take noticeable time to connect.
+    private static readonly SESSION_READY_TIMEOUT_MS = 10000;
 
     encryption!: Encryption;
     serverID!: string;
@@ -295,6 +297,11 @@ class Sync {
         const normalizedMessage = normalizeRawMessage(localId, localId, createdAt, content);
         if (normalizedMessage) {
             this.applyMessages(sessionId, [normalizedMessage]);
+        }
+
+        const ready = await this.waitForAgentReady(sessionId);
+        if (!ready) {
+            log.log(`Session ${sessionId} not ready after timeout, sending anyway`);
         }
 
         // Send message with optional permission mode and source identifier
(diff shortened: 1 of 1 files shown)


bra1nDump (approved):
Beautiful stuff!

PR #272, opened 2025-12-07
feat: New Session Page w/ one touch Profiles, adds z.ai / deepseek / microsoft AI Backend Support

ahundt:
# feat: New Session Page with Profile System and Alternative AI Backend Support

companion to https://github.com/slopus/happy-cli/pull/107 (replaces https://github.com/slopus/happy-cli/pull/86)

## Motivation

Users want to use Happy with alternative AI backends (OpenAI, DeepSeek, Anthropic direct API) and configure environment variables from their mobile device. Previously, all AI configuration required manual CLI setup. This PR enables:

- **Switch Claude Code backends** e.g. anthropic, z.ai's GLM4.6, and deepseek on claude code, OpenAI Codex, with microsoft and openai backends from your phone.
- **Secure API key management** - Configure API keys once in a profile, sync to CLI automatically
- **Create sessions on your phone, continue on your desktop terminal** - tmux integration keeps sessions alive when switching devices
- **No more CLI fumbling** - Savable / selectable session profiles, Permission modes, working directories, all with pre-set reasonable defaults you don't have to change unless you want a specially configured setup.

## New Session Wizard

One tap profiles to select different cli tools and ai backends, tap once then type your message and hit send!

happy-new-session-0-select-profile-2025-12-07
happy-new-session-1-dir-2025-12-07

## Editable Profiles

happy-new-session-3-edit-profile-2025-12-07
happy-new-session-4-edit-profile-2025-12-07

### 1. Profile System with Alternative AI Backend Support
- Built-in profiles for Claude, OpenAI, DeepSeek, Gemini, Azure, and more
- Custom profile creation with environment variable configuration
- Cross-device profile sync with end-to-end encryption
- Profile compatibility warnings when CLI lacks required tools

### 2. Environment Variable Configuration
- New `EnvironmentVariableCard` and `EnvironmentVariablesList` components
- Security-aware display: shows `${VAR}` syntax without exposing secrets
- Bash parameter expansion support: `${VAR:-default}` syntax
- Retrieve actual values from remote machine for verification

### 3. tmux Session Integration (when tmux is available)
- Enable/disable tmux per session with checkbox
- Defaults to first existing session when name is empty
- Session persistence across device switches

### 4. Agent-Specific Permission Modes
- Permission mode selector in wizard (plan, default, auto, yolo)
- Auto-reset permission mode when switching between Claude/Codex
- Cross-agent permission mapping for compatibility

### 5. Working Directory Selection
- Inline path selector with search, favorites, and recent history
- Collapsible sections for All/Recent/Favorites
- Machine online status indicators

### 6. CLI Detection & Availability Warnings
- Automatic non-blocking CLI detection per machine
- Dismissible warnings (per-machine or global)
- Profile compatibility indicators based on available CLIs

### 7. Internationalization
- 900+ new translation strings across 7 languages (en, ru, pl, es, pt, ca, zh-Hans)
- Complete coverage for new wizard and profile management UI

### 8. UI/UX Polish
- Session view header no longer disappears on desktop resize
- SearchableListSelector border radius consistency
- Theme spacing/sizing improvements
- Sticky AgentInput at bottom of wizard

## Stats
- **158 commits** | **54 files changed** | **+9,636 / -879 lines**

## Related

- Current Companion CLI PR: [happy-cli#107 ](https://github.com/slopus/happy-cli/pull/107) (provides backend support for profiles and tmux)
- Outdated Companion CLI PR: [happy-cli#86 ](https://github.com/slopus/happy-cli/pull/86) (provides backend support for profiles and tmux)

### DIFF
--- a/sources/components/MessageView.tsx
+++ b/sources/components/MessageView.tsx
@@ -10,6 +10,7 @@ import { ToolView } from "./tools/ToolView";
 import { AgentEvent } from "@/sync/typesRaw";
 import { sync } from '@/sync/sync';
 import { Option } from './markdown/MarkdownView';
+import { useSetting } from "@/sync/storage";
 
 export const MessageView = (props: {
   message: Message;
@@ -88,9 +89,15 @@ function AgentTextBlock(props: {
   message: AgentTextMessage;
   sessionId: string;
 }) {
+  const experiments = useSetting('experiments');
   const handleOptionPress = React.useCallback((option: Option) => {
     sync.sendMessage(props.sessionId, option.title);
   }, [props.sessionId]);
+
+  // Hide thinking messages unless experiments is enabled
+  if (props.message.isThinking && !experiments) {
+    return null;
+  }
 
   return (
     <View style={styles.agentMessageContainer}>
--- a/sources/sync/reducer/reducer.ts
+++ b/sources/sync/reducer/reducer.ts
@@ -123,6 +123,7 @@ type ReducerMessage = {
     createdAt: number;
     role: 'user' | 'agent';
     text: string | null;
+    isThinking?: boolean;
     event: AgentEvent | null;
     tool: ToolCall | null;
     meta?: MessageMeta;
@@ -631,12 +632,14 @@ export function reducer(state: ReducerState, messages: NormalizedMessage[], agen
             for (let c of msg.content) {
                 if (c.type === 'text' || c.type === 'thinking') {
                     let mid = allocateId();
+                    const isThinking = c.type === 'thinking';
                     state.messages.set(mid, {
                         id: mid,
                         realID: msg.id,
                         role: 'agent',
                         createdAt: msg.createdAt,
-                        text: c.type === 'text' ? c.text : `*Thinking...*\n\n*${c.thinking}*`,
+                        text: isThinking ? `*Thinking...*\n\n*${c.thinking}*` : c.text,
+                        isThinking,
                         tool: null,
                         event: null,
                         meta: msg.meta,
@@ -862,12 +865,14 @@ export function reducer(state: ReducerState, messages: NormalizedMessage[], agen
             for (let c of msg.content) {
                 if (c.type === 'text' || c.type === 'thinking') {
                     let mid = allocateId();
+                    const isThinking = c.type === 'thinking';
                     let textMsg: ReducerMessage = {
                         id: mid,
                         realID: msg.id,
                         role: 'agent',
                         createdAt: msg.createdAt,
-                        text: c.type === 'text' ? c.text : `*Thinking...*\n\n*${c.thinking}*`,
+                        text: isThinking ? `*Thinking...*\n\n*${c.thinking}*` : c.text,
+                        isThinking,
                         tool: null,
                         event: null,
                         meta: msg.meta,
@@ -1114,6 +1119,7 @@ function convertReducerMessageToMessage(reducerMsg: ReducerMessage, state: Reduc
             createdAt: reducerMsg.createdAt,
             kind: 'agent-text',
             text: reducerMsg.text,
+            ...(reducerMsg.isThinking && { isThinking: true }),
             meta: reducerMsg.meta
         };
     } else if (reducerMsg.role === 'agent' && reducerMsg.tool !== null) {
--- a/sources/sync/typesMessage.ts
+++ b/sources/sync/typesMessage.ts
@@ -46,6 +46,7 @@ export type AgentTextMessage = {
     localId: string | null;
     createdAt: number;
     text: string;
+    isThinking?: boolean;
     meta?: MessageMeta;
 }
 


bra1nDump:
Hey @ahundt, during our call last week I mentioned I should be able to merge this & the CLI change the past weekend, which unfortunately I could not make happen.

We've also talked about gating the frontend changes, do you think you will have some bandwidth to do that?

ahundt:
@bra1nDump ok I've merged main into this branch and tested the feature flag and everything looks good to me with that degree of testing, same as https://github.com/slopus/happy-cli/pull/107

ahundt:
@bra1nDump also, the front end gating aka feature flags are added and tested and keep the main branch functionality by default

leeroybrun:
Hi @ahundt @bra1nDump ,

Were the UI changes of the new session fields wanted and expected? 

Because in my opinion, it gives too much attention to the "Machine & Folder" compared to the text input. In my opinion, the previous UI was cleaner and clearer, but I don't know if this was a wanted change.

Screenshot 2026-01-12 at 21 52 16

Also the new session wizard and the profiles settings do not seem really unified with the rest of the UI in my opinion, which was much more "clean", no borders, dark gray vs hash black, etc. Also, don't know if that is only on my fork, but the icon to select preferences in the profiles seems broken (no icon for the selected permission).

Screenshot 2026-01-12 at 21 54 26
Screenshot 2026-01-12 at 21 54 58

bra1nDump:
I wanted to merge the changes because @ahundt has been working on this for a while and I have been stalling a review / merge.

I have provided feedback on this UI wise to @ahundt.

We certainly need to follow up with a redesign of profiles. Some issues - nesting, repeated content, strange scroll behaviors, paddings etc.

I want to keep the current prod session creation ui unchanged and only have profiles affect users who toggle the flag on for now. Changing the current ui was not very intentional. Later ideally we would transition to a system where profiles will get replaced with "environments" like in Claude, codex, cursor web apps. But we need to think more about this problem, open to suggestions here

leeroybrun:
Ok I understand. Happy - no pun intended - to work a bit on this if that helps, at least to restore the correct and cleaner previous "new session" UI when the feature flag is not enabled. 

Regarding the profiles and "new session" UI with profiles: in my opinion, the "Profiles" could be a simple knob in the "new session" UI like the rest, and when users click on it, they can select a different profile, no? This makes the feature easily accessible without changing the whole "new session" UI, which was very clean and beautiful previously, in my opinion.

Screenshot 2026-01-12 at 22 16 37

bra1nDump:
Yep I agree with that approach. If you can help with restoring ui to the previous state for non enabled users and improving the profile picker (possibly moving to a separate pill as you suggest) that would be amazing! @ahundt and I can help test

ahundt:
Hey thanks for the merge! One caveat, @leeroybrun, with the prior ui you provided in your screenshot, which just says 'mac', one cannot tell the actual machine that is in use if they have multiple machines, particularly if they are over DNS or another mechanism with a long path.

bra1nDump:
you can rename machines. maybe we need to surface this more easily

leeroybrun:
@ahundt @bra1nDump I'm currently working on this, will provide a status update later today or tomorrow, hopefully

bra1nDump:
thank you!

ahundt:
Awesome, thanks!

leeroybrun:
@ahundt @bra1nDump I'm almost done, just putting the final touches and doing a thorough final review. 

I was able to unify everything with the rest of the app's design and UI, and address a few issues along the way. 

I will create PRs as soon as I'm done and provide an easy way for you to test and review.

ahundt:
Hopefully the starting point was useful, thanks for picking it up!

ahundt:
@leeroybrun are there updated prs addressing the outstanding items related to this PR? I saw the current PR with fixes was closed. 

Also one design decision I had made was to not assume tmux, but if it could be a hard dep I think usability could improve quite a lot, at least for Claude code integrations, though with some major tradeoffs, by having happy just directly type into Claude code message fields via tmux. One big advantage might be not needing to swap between remote and local modes explicitly. Deeper hooks integration could probably help mitigate the downsides too.

leeroybrun:
@ahundt yes, there had been a lot of work in the meantime, the issues should be fully addressed in https://github.com/slopus/happy/pull/487

About tmux: not sure we should make that a hard dep, and not sure this is the solution for Claude remote <-> local switchng, but I've improved things a bit in that regard, and also added the ability to install deps easily from the UI, so could simply surface that for tmux next to the tmux settings. 

Also, a lot of Claude remote <-> local switching issues have been resolved directly in the new PR as well, so they should be less of an issue. And now users can use "happy attach" to re-attach locally to a remote-started happy session, to then switch to Claude local control if they want to.

Not using Claude SDK anymore and using tmux to type into Claude seems really brittle to me, and we lose all the advantages of Claude SDK and the deeper integrations it gives us. In the end, our local/remote switching should work now, and should be less of an issue, but we can revisit later if needed.

PR #308, opened 2025-12-30
feat: add pagination for loading older chat messages

leeroybrun:
Implements pagination in the chat message list, allowing users to load older messages as they scroll. Updates the ChatList component to handle loading state and triggers loading via scroll events. Adds pagination state and logic to the sync layer, including tracking the oldest loaded message and whether more messages are available.


bra1nDump:
This looks like a good idea - pagination for older messages is definitely needed for long sessions.

A bit hesitant to merge though - worried about hidden implications with the sync layer and message ordering. Would need to test this thoroughly.

Will revisit when I have more time to dig into it.

Issue #304, opened 2025-12-29
Claude Code Clarification / Plan Questions aren't handled


mediavariance:
Love the project! Maybe this is in the works, if not I would love to fix it:

When Claude code asks clarifying or planning questions, happy doesn't display them and instead spits out the json object containing the questions and potential answers.

WeZZard:
Basically the AskUserTool and ExitPlanMode tool support.

SumeraMartin:
Is anyone currently working on this, or may I start?

bra1nDump:
tracking in #375 - the askUserQuestion display has been fixed, verifying the full flow works end-to-end. closing here.

PR #317, opened 2025-12-31
feat: add Claude Code AskUserQuestion tool view with radio/checkbox UI

SaneBow:
- Add AskUserQuestionView component for interactive question prompts
- Support single-select (radio) and multi-select (checkbox) options
- Hide PermissionFooter for AskUserQuestion (uses custom submit UI)
- Pass sessionId to tool view components
- Add translations for submit button

ScreenShot 2025-12-31 at 18 28 43@2x

This should address #304


bra1nDump:
Hell yes! Is there a CLI part to this or this just works?

SaneBow:
> Hell yes! Is there a CLI part to this or this just works?

It works with current CLI, with some workaround. Better solution would be change the CLI as well.

PR #325, opened 2026-01-01
fix(ios): Enable HTTP connections to local networks including Tailscale

### LINKED ISSUE #319: Error: Request failed with status code 404
```
$ happy
Error: Request failed with status code 404
```

craigvandotcom:
## Problem

When using self-hosted Happy Server instances accessible via Tailscale or local networks, the iOS app fails to connect with "failed to connect" errors, even though the server is accessible in Safari and other browsers.

This occurs because iOS App Transport Security (ATS) blocks HTTP connections to local network IP addresses by default, including:
- RFC1918 private IPs (10.x.x.x, 172.16.x.x, 192.168.x.x)  
- Tailscale CGNAT range (100.64.0.0/10)

## Solution

Add `NSAllowsLocalNetworking: true` to the iOS App Transport Security configuration in `app.config.js`.

This is the [official Apple-recommended approach](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowslocalnetworking) for apps that need to communicate with local network servers over HTTP.

## Changes

- Added `NSAppTransportSecurity` configuration with:
  - `NSAllowsLocalNetworking: true` - Allows HTTP to local networks
  - `NSAllowsArbitraryLoads: false` - Maintains security by still requiring HTTPS for public internet

## Security

This change maintains security by:
- Only allowing HTTP for local network addresses
- Still requiring HTTPS for all public internet connections
- Following Apple's documented best practices for local network communication

## Testing

Tested with self-hosted Happy Server on:
- ✅ Local WiFi (192.168.x.x)
- ✅ Tailscale (100.x.x.x)
- ✅ Localhost (127.0.0.1)

Fixes #319

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -34,7 +34,11 @@ export default {
             infoPlist: {
                 NSMicrophoneUsageDescription: "Allow $(PRODUCT_NAME) to access your microphone for voice conversations with AI.",
                 NSLocalNetworkUsageDescription: "Allow $(PRODUCT_NAME) to find and connect to local devices on your network.",
-                NSBonjourServices: ["_http._tcp", "_https._tcp"]
+                NSBonjourServices: ["_http._tcp", "_https._tcp"],
+                NSAppTransportSecurity: {
+                    NSAllowsLocalNetworking: true,
+                    NSAllowsArbitraryLoads: false
+                }
             },
             associatedDomains: variant === 'production' ? ["applinks:app.happy.engineering"] : []
         },


antoinekm (approved):

mike2153:
Sorry but this has been quoted as a fix for launching the CLI tool resulting in a 404. I don't believe this would fix that unless I am mistaken. The cli issue is connecting to the happy server?

AventusM:
> Sorry but this has been quoted as a fix for launching the CLI tool resulting in a 404. I don't believe this would fix that unless I am mistaken. The cli issue is connecting to the happy server? 


Would think so since my attempts to set up self hosted version fails due to this. Perhaps they have similar on their end.

craigvandotcom:
thats my mistake - this fix was unrelated to the 404

mikeyward:
Were you able to work around it? I encountered the same issue (self-hosted Happy server, connecting over tailscale), and when I encountered the error, I set it up to operate via SSH, to dodge the iOS ATS issue, but am still experiencing silent failures. It seems like the authentication links / qr codes are not respecting the custom server and directing straight to the Happy cloud version.

craigvandotcom:
I actually landed up changing to https with tailscale. seems to be working
mostly fine. The occasional process unexpectedly quit.

GrocerPublishAgent:
Hey thanks for fixing this. I'm not going to merge this right away only because I think we are trying to do an over the air update of the app this week. Changing the native permissions will require submitting to Apple and Google for review.

PR #327, opened 2026-01-02
feat: ability to swipe to archive/delete sessions from the top-level page

danielamitay:
Something I have long wanted was the ability to quickly/easily clean up terminal sessions from the top level page. Right now the only way to do so is to tap into a session and then tap again into the settings page for it.

This pull request simply adds swipe-to-delete functionality

| Archive | Archive Confirmation | Delete | Delete Confirmation |
|---|---|---|---|
|  |  |  |  |


bra1nDump:
Thank you!

PR #339, opened 2026-01-06
feat(i18n): add Japanese language support

manatto:
- Add 'ja' to SupportedLanguage type and SUPPORTED_LANGUAGES config
- Create Japanese translation file with 850+ translation keys
- Register Japanese translations in the i18n module


bra1nDump:
Thank you!!

PR #364, opened 2026-01-11
fix(web): alert modals not clickable when shown above expo-router modals

leeroybrun:
On web, when an alert modal is shown while an expo-router modal (like "new session") is open, the alert's OK button was unclickable. This happened because expo-router sets pointer-events: none on body and pointer-events: auto on its own overlay, which intercepted clicks meant for our alert modal.

For example, we could never click "OK" and close that modal:
Screenshot 2026-01-11 at 16 36 27

Fix:
- Add pointerEvents: 'auto' to modal container to enable interactions
- Stop event propagation on web to prevent clicks from reaching expo-router's overlay

### DIFF
--- a/sources/modal/components/BaseModal.tsx
+++ b/sources/modal/components/BaseModal.tsx
@@ -9,6 +9,13 @@ import {
     Platform
 } from 'react-native';
 
+// On web, stop events from propagating to expo-router's modal overlay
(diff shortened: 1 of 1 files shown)


GrocerPublishAgent:
Alright, I cherry-picked this commit onto main after resolving conflicts. a57f73520f147a84feb9b73c591985d2b59ba64f

PR #366, opened 2026-01-12
Add message history navigation with arrow keys

aronchick:
## Summary
Implements shell-like history navigation for the agent input field, allowing users to navigate through previous messages using arrow keys (similar to terminal history or Claude Code).

## Features
- Navigate through message history with Up/Down arrow keys (web only)
- Draft preservation: current input is saved when first navigating into history
- Draft restoration: saved draft is restored when navigating back past all history
- Automatic cursor positioning at end of loaded text
- History state reset on message send
- Non-interfering: only active when autocomplete suggestions are not visible

## Technical details
- Created `useUserMessageHistory` hook in `sources/hooks/`
- Collects all user messages from all sessions, sorted by timestamp
- Uses on-demand history fetching to avoid stale data
- Integrates with existing AgentInput keyboard navigation

## Test plan
- [ ] Test arrow up/down navigation through message history on web
- [ ] Verify draft preservation when navigating into history
- [ ] Verify draft restoration when navigating back
- [ ] Test that history resets after sending a message
- [ ] Verify no interference with autocomplete functionality

### DIFF
--- a/.github/workflows/no-ai-authors.yml
+++ b/.github/workflows/no-ai-authors.yml
@@ -1,0 +1,40 @@
+name: Block AI Author Attribution
+on:
+  push:
+    branches: ["**"]
+  pull_request:
+    branches: ["**"]
+jobs:
+  check-authors:
+    name: Verify no AI commit authors
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


GrocerPublishAgent:
This is pretty neat. I'll try to test it out on tuesday.

PR #408, opened 2026-01-21
fix(markdown): implement column-first layout for consistent table cell widths

### LINKED ISSUE #294: Markdown table cells don't resize together on mobile
## Description

Markdown table rendering has layout issues on mobile - cells in the same column don't resize together when content varies in length. This causes misaligned columns and inconsistent table appearance.

## How to Reproduce

1. Run the app on iOS or Android
2. Navigate to Settings → Dev → messages-demo
3. Observe the tables with varying cell content lengths
4. Notice that columns don't align properly when one cell has longer content

## Current Behavior

- Each cell has a minimum width (100px) but expands independently based on content
- Cells in different rows of the same column can have different widths
- Long content causes individual cells to expand, breaking column alignment

## Expected Behavior

- All cells in the same column should have the same width
- Column width should be determined by the widest cell in that column
- Table should be horizontally scrollable when content exceeds screen width

## Technical Context

The table is rendered using React Native's `View` and `ScrollView` components in `sources/components/markdown/MarkdownView.tsx`. The current implementation uses `minWidth` on cells but doesn't synchronize widths across rows.

## Potential Solutions

1. **Pre-calculate column widths**: Measure all cell content before rendering and set fixed widths per column

2. **Use `onLayout` callbacks**: Collect cell widths after initial render, then re-render with synchronized widths

3. **CSS Table display (web only)**: On web, could use actual CSS `display: table` which handles this natively

4. **Third-party libraries**: 
   - `react-native-table-component`
   - `react-native-data-table`
   - Custom implementation using `FlatList` with fixed column widths

5. **Flexbox with fixed ratios**: Calculate percentage widths based on header content and apply uniformly

## Files Involved

- `sources/components/markdown/MarkdownView.tsx` - `RenderTableBlock` function
- `sources/components/markdown/parseMarkdownBlock.ts` - Table parsing logic

## Additional Notes

- The horizontal scrolling now works correctly (fixed in recent commit)
- Long press to copy also works without blocking scroll gestures
- This issue is specifically about column width synchronization

hyacz:
Fixes #294 - Changed table rendering from row-first to column-first layout to ensure cells in the same column have consistent widths.

Changes:
- Render each column as a vertical container (flip axis approach)
- Columns contain their header cell and all data cells
- Column width determined by widest cell content naturally
- Updated styles: tableContent (row), tableColumn, tableCell borders

This approach uses Flexbox's natural behavior instead of pre-calculating widths or requiring multiple render passes.

| Before | After |
|:--- |:--- |
| before | after |

Generated with [Claude Code](https://claude.ai/code) via [Happy](https://happy.engineering)

### DIFF
--- a/expo-app/sources/components/markdown/MarkdownView.tsx
+++ b/expo-app/sources/components/markdown/MarkdownView.tsx
@@ -228,9 +228,9 @@ function RenderSpans(props: { spans: MarkdownSpan[], baseStyle?: any }) {
     </>)
 }
 
-// TODO: Table rendering has layout issues on mobile - cells don't resize together when content varies.
-// See https://github.com/slopus/happy/issues/294 for details and potential fixes.
-// To repro: Navigate to dev/messages-demo and observe tables with long cell content.
+// Table rendering uses column-first layout to ensure consistent column widths.
+// Each column is rendered as a vertical container with all its cells (header + data).
+// This ensures that cells in the same column have the same width, determined by the widest content.
 function RenderTableBlock(props: {
     headers: string[],
     rows: string[][],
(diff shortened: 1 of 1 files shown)


hyacz:
The PR has been modified to fit the structure of the mono-repo.

bra1nDump:
thank you!

PR #476, opened 2026-01-26
feature(cli): add --chrome and --no-chrome flags for Chrome mode (Claude)

gaius-codius:
Add support for Claude --chrome in Happy CLI:
- Add chromeMode setting to persist default preference
- Add --chrome flag to enable Claude chrome browser access for a session
- Add --no-chrome flag to disable Chrome even if default is on
- Flag overrides always take precedence over settings
- Update help text with Chrome examples

-Future todo: Add toggle in mobile app settings page to launch claude with --chrome flag.

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -10,7 +10,7 @@
 import chalk from 'chalk'
 import { runClaude, StartOptions } from '@/claude/runClaude'
 import { logger } from './ui/logger'
-import { readCredentials } from './persistence'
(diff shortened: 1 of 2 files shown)


GrocerPublishAgent:
Thanks for contributing a bug fix! Happy is back 🤠.

I'm going to go ahead and get this merged so the CLI works for you in the next release this week, but I think we should iterate on this because I think I saw AIBackendProfile might be the location to store agent specific settings.

PR #497, opened 2026-01-30
fix: prevent orphan Claude processes when using native binary

deflating:
## Summary
- Fixes orphan Claude Code processes when installed as native binary
- Adds SIGTERM/SIGINT signal forwarding in launcher and SDK

## Problem
When Claude Code is installed as a native binary (via native installer, not npm), the launcher spawns it as a child process. If the parent process receives SIGTERM/SIGINT (e.g., when switching modes or aborting), it would exit without forwarding the signal to the child. This leaves Claude binary processes running as orphans (PPID becomes 1), consuming memory indefinitely.

## Solution
Added signal handlers in two places:
1. `scripts/claude_version_utils.cjs` - forwards SIGTERM/SIGINT to spawned binary child
2. `src/claude/sdk/query.ts` - adds signal handlers alongside existing cleanup logic

## Test plan
- [ ] Start Happy with native binary Claude installation
- [ ] Send message, switch to remote mode (triggers abort)
- [ ] Verify Claude process terminates (check `ps aux | grep claude`)
- [ ] Previously: orphaned process with PPID=1; Now: process should be gone

### DIFF
--- a/packages/happy-cli/scripts/claude_version_utils.cjs
+++ b/packages/happy-cli/scripts/claude_version_utils.cjs
@@ -497,6 +497,17 @@ function runClaudeCli(cliPath) {
             stdio: 'inherit',
             env: process.env
         });
+
+        // Forward termination signals to child process
+        // Without this, killing the launcher orphans the Claude binary (PPID becomes 1)
+        const forwardSignal = (signal) => {
+            if (!child.killed) {
+                child.kill(signal);
+            }
+        };
+        process.on('SIGTERM', () => forwardSignal('SIGTERM'));
+        process.on('SIGINT', () => forwardSignal('SIGINT'));
+
         child.on('exit', (code) => {
             process.exit(code || 0);
         });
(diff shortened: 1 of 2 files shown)


HirokiKobayashi-R:
We've been running a similar fix in our fork and can confirm this resolves the orphan process issue when the launcher is killed. A couple of observations from our experience:

**The `child.killed` guard has a subtle issue.** `child.killed` becomes `true` after the first `child.kill()` call — not when the child actually exits. This means if the child traps the first signal (e.g., SIGINT for graceful shutdown), subsequent signals from the parent will be silently dropped because `child.killed` is already `true`.

In our fork we addressed this by tracking exit state ourselves:

```js
let childExited = false;
const forwardSignal = (signal) => {
    if (!childExited) {
        try {
            child.kill(signal);
        } catch {
            // Child may have exited between our check and the kill call
        }
    }
};
child.on('exit', () => { childExited = true; });
```

The `try-catch` handles a race condition where the child exits between the `!childExited` check and the `child.kill()` call — without it, you get an unhandled ESRCH error.

**Also worth considering:**
- Adding `SIGHUP` to the forwarded signals. When a terminal disconnects (e.g., SSH drops), the parent receives SIGHUP but the child won't unless it's forwarded.
- Cleaning up signal handlers in the `child.on('exit')` callback to avoid leaking listeners if the parent process continues running after the child exits.

The core approach is solid though — signal forwarding is the right fix here. Without it, the orphaned Claude binary keeps holding stdin, which is especially problematic during remote-to-local mode switches.

PR #508, opened 2026-02-01
fix: preserve empty cells in markdown table parsing

### LINKED ISSUE #506: Markdown tables do not render cleanly - empty cells cause column misalignment
## Description

Markdown tables with empty cells do not render correctly in the Happy app. Empty cells are removed during parsing, causing column misalignment.

## Steps to Reproduce

1. In a conversation, have Claude output a markdown table with empty cells:

```markdown
| A | B | C |
|---|---|---|
| 1 |   | 3 |
|   | 2 |   |
```

2. Observe that the table columns are misaligned

## Expected Behavior

Empty cells should be preserved, and columns should remain properly aligned.

## Actual Behavior

Empty cells are filtered out by the parser, causing cells to shift left and break the table structure.

## Root Cause

In `packages/happy-app/sources/components/markdown/parseMarkdownBlock.ts`, lines 28-31 and 42-45:

```typescript
.filter(cell => cell.length > 0)  // This removes empty cells!
```

This filter removes empty cells that result from legitimate empty table content, not just pipe artifacts.

## Proposed Fix

Replace the filter with logic that only trims leading/trailing empty cells from pipe artifacts, while preserving intentionally empty cells in the middle of rows.

lucharo:
## Summary

- Fix markdown table rendering when cells contain empty values
- Tables with empty cells were displaying incorrectly due to the parser filtering out all empty strings
- Add `trimPipeArtifacts()` helper to only remove leading/trailing empty strings from pipe characters
- Pad rows to match header count for consistent column alignment
- Improve table styling with better padding and responsive width

## Test plan

- [ ] Render a table with empty middle cells: `| A | | C |`
- [ ] Verify columns remain aligned
- [ ] Verify empty cells display correctly (not shifted)
- [ ] Check tables with uneven row lengths

Fixes #506

### DIFF
--- a/packages/happy-app/sources/components/markdown/MarkdownView.tsx
+++ b/packages/happy-app/sources/components/markdown/MarkdownView.tsx
@@ -504,10 +504,9 @@ const style = StyleSheet.create((theme) => ({
         borderColor: theme.colors.divider,
         borderRadius: 8,
         overflow: 'hidden',
-        alignSelf: 'flex-start',
     },
     tableScrollView: {
-        flexGrow: 0,
+        flexGrow: 1,
     },
     tableContent: {
         flexDirection: 'row',
@@ -521,11 +520,12 @@ const style = StyleSheet.create((theme) => ({
         borderRightWidth: 0,
     },
     tableCell: {
-        paddingHorizontal: 12,
-        paddingVertical: 8,
+        paddingHorizontal: 16,
+        paddingVertical: 10,
         borderBottomWidth: 1,
         borderBottomColor: theme.colors.divider,
         alignItems: 'flex-start',
+        minWidth: 40,
     },
     tableCellFirst: {
         borderTopWidth: 0,
--- a/packages/happy-app/sources/components/markdown/parseMarkdownBlock.ts
+++ b/packages/happy-app/sources/components/markdown/parseMarkdownBlock.ts
@@ -1,6 +1,26 @@
 import type { MarkdownBlock } from "./parseMarkdown";
 import { parseMarkdownSpans } from "./parseMarkdownSpans";
 
+/**
+ * Trims empty strings that result from leading/trailing pipe characters,
+ * while preserving intentionally empty cells in the middle of the row.
+ *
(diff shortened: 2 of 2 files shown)


HirokiKobayashi-R:
We've been running this fix in our fork and can confirm it correctly handles tables with empty cells. The `trimPipeArtifacts` approach is the right solution — the previous `.filter(cell => cell.length > 0)` was too aggressive and couldn't distinguish between pipe-boundary empty strings and intentionally empty cells.

A few notes from our testing:

1. **The row padding logic (lines 67-69) is a good defensive addition.** In practice, some markdown generators produce rows with fewer cells than the header. Without padding, the table would render with misaligned columns. We've seen this with LLM-generated markdown in particular.

2. **Edge case we verified:** tables where the first or last data cell is intentionally empty (e.g., `| | B | C |` or `| A | B | |`). The `trimPipeArtifacts` function correctly handles this because it only removes the boundary empty strings from `split('|')`, not the cell-position ones. The docstring example at line 8 documents this clearly.

3. **Styling changes are a nice touch.** The increased padding (`paddingHorizontal: 16`, `paddingVertical: 10`) and `minWidth: 40` on cells makes tables more readable, especially on mobile where columns were previously collapsing too aggressively. Removing `alignSelf: 'flex-start'` from the table container and setting `flexGrow: 1` on the scroll view allows tables to fill the available width rather than shrinking to content, which is a better default for responsive layouts.

Clean fix with good test coverage in the accompanying test file. Would be great to see this merged.

tools-bra1ndump:
Thanks for jumping on this — you were first to identify the right fix (trimming pipe artifacts instead of filtering all empties). We ended up going with #730 which builds on the same idea but also handles blank lines between rows (common in LLM output) and includes tests. Closing in favor of that. Appreciate the contribution!

bra1nDump:
Thanks for jumping on this — you were first to identify the right fix (trimming pipe artifacts instead of filtering all empties). We ended up going with #730 which builds on the same idea but also handles blank lines between rows (common in LLM output) and includes tests. Closing in favor of that. Appreciate the contribution!

PR #553, opened 2026-02-07
fix: PTY proxy for local mode stdin corruption

### LINKED ISSUE #423: Terminal input stuck after exit (raw mode not restored)
## Bug Description

After exiting happy-cli (via Ctrl+C or normal exit), the terminal input becomes stuck at whitespace. Users need to run `stty sane` or `reset` to restore normal terminal behavior.

## Root Cause

The terminal raw mode is enabled in `src/claude/claudeRemoteLauncher.ts:64`:
```typescript
process.stdin.setRawMode(true);
```

The cleanup code that restores terminal state is in a `finally` block at line 446-448:
```typescript
finally {
    if (process.stdin.isTTY) {
        process.stdin.setRawMode(false);
    }
}
```

However, when SIGINT/SIGTERM is caught, the `cleanup()` function in `src/claude/runClaude.ts:406` calls `process.exit(0)` directly:
```typescript
const cleanup = async () => {
    // ... cleanup code that does NOT restore terminal state
    process.exit(0);  // Immediately terminates, skipping finally blocks
};
process.on('SIGINT', cleanup);
```

`process.exit()` terminates the process immediately, **bypassing** the `finally` block in `claudeRemoteLauncher.ts`. The terminal is left in raw mode.

## Steps to Reproduce

1. Run `happy`
2. Wait for remote mode to start (raw mode enabled)
3. Press Ctrl+C to exit
4. Try typing in terminal - input appears stuck/doesn't echo

## Proposed Fix

Add terminal state restoration to the `cleanup` function in `src/claude/runClaude.ts` before calling `process.exit()`:

```typescript
const cleanup = async () => {
    try {
        // ... existing cleanup code ...

        // Restore terminal state before exit
        if (process.stdin.isTTY) {
            process.stdin.setRawMode(false);
        }

        logger.debug('[START] Cleanup complete, exiting');
        process.exit(0);
    } catch (error) {
        // Also restore terminal in error path
        if (process.stdin.isTTY) {
            process.stdin.setRawMode(false);
        }
        logger.debug('[START] Error during cleanup:', error);
        process.exit(1);
    }
};
```

## Environment

- macOS (Darwin 25.2.0)
- Node.js with node-pty

### LINKED ISSUE #527: Moving from remote mode to local mode removes the history and harm the writing in local mode
Tech stack:
Happy app on iPhone17 pro, iOS 26.2
Apple M4 Max, macOS version: 26.2
Claude Code v2.1.29 via iTerm

attached is a screen shot showing the issue:

Image

mfazekas:
After local→remote→local mode switch, Node.js children spawned with `stdio:'inherit'` stop receiving stdin.

Fix: spawn local-mode children via `node-pty` in their own PTY (child never inherits parent's fd 0), and give Ink a separate `/dev/tty` fd so it never touches `process.stdin`.

<details>
<summary>Investigation details</summary>

Native binaries (e.g. `cat`) using raw `read()` are unaffected — only Node.js children break.

The remote mode code and Ink both manipulate `process.stdin` (flowing mode, `setEncoding`, `setRawMode`) between child spawns. After cleanup (`pause()`, `removeAllListeners()`), something in the libuv TTY handle state remains dirty, preventing subsequent Node.js children from reading stdin.

E2e tests reproducing the issue: [`test/pty-proxy-e2e`](https://github.com/mfazekas/happy/tree/test/pty-proxy-e2e) ([test file](https://github.com/mfazekas/happy/blob/test/pty-proxy-e2e/packages/happy-cli/src/claude/terminalSwitch.e2e.test.ts))

Code that touches `process.stdin` between spawns:
- [claudeRemoteLauncher.ts:62-66](https://github.com/slopus/happy/blob/3ed8b121580b52e9c56a1973a92e80bf7f1a220b/packages/happy-cli/src/claude/claudeRemoteLauncher.ts#L62-L66)
- [Ink App.tsx:161 (v6.6.0)](https://github.com/vadimdemedes/ink/blob/v6.6.0/src/components/App.tsx#L161)

</details>

Fixes #527
Fixes #423

### DIFF
--- a/packages/happy-cli/src/claude/claudeLocal.test.ts
+++ b/packages/happy-cli/src/claude/claudeLocal.test.ts
@@ -1,14 +1,19 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { claudeLocal } from './claudeLocal';
 
-// Use vi.hoisted to ensure mock functions are available when vi.mock factory runs
-const { mockSpawn, mockClaudeFindLastSession } = vi.hoisted(() => ({
-    mockSpawn: vi.fn(),
-    mockClaudeFindLastSession: vi.fn()
+const { mockPtySpawn, mockClaudeFindLastSession, mockNetCreateServer } = vi.hoisted(() => ({
+    mockPtySpawn: vi.fn(),
+    mockClaudeFindLastSession: vi.fn(),
+    mockNetCreateServer: vi.fn(),
 }));
 
-vi.mock('node:child_process', () => ({
-    spawn: mockSpawn
+vi.mock('node-pty', () => ({
+    default: { spawn: mockPtySpawn },
+    spawn: mockPtySpawn,
+}));
+
+vi.mock('node:net', () => ({
+    createServer: mockNetCreateServer,
 }));
 
 vi.mock('@/ui/logger', () => ({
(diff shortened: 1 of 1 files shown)


speakman:
Hey @mfazekas! Found the root cause of the PTY exhaustion on macOS.

**TL;DR**: node-pty v1.1.0 has an off-by-one bug in `pty_posix_spawn()` that leaks 1 PTY master fd per `pty.spawn()` call. The upstream fix is in [microsoft/node-pty#882](https://github.com/microsoft/node-pty/pull/882) (merged 2026-01-28), available in `1.2.0-beta.9+`.

I opened a PR against your branch with the fix: https://github.com/mfazekas/happy/pull/1

**Changes:**
1. Upgrade node-pty from `1.1.0` → `1.2.0-beta.11` (includes the leak fix)
2. Fix abort handler to kill the process group (`-pid`, SIGTERM) instead of just the launcher PID (SIGHUP), so the Claude binary grandchild is also terminated

The bug is in the `low_fds` cleanup loop in `src/unix/pty.cc` — it uses `for (; count > 0; count--)` which skips `low_fds[0]` when `count == 0` (the normal case). The fix changes it to `for (size_t i = 0; i <= count; i++)`.

Issue #443, opened 2025-11-25
Happy crashes when switching input devices to local mode


flohump:
## Description

Happy crashes completely when switching input from a remote device (e.g., iPhone via Universal Control) back to the main computer and entering local mode.

## Environment
- Happy version: 0.11.2
- Claude Code version: 2.0.14
- OS: macOS (Darwin 24.6.0)
- Device setup: MacBook Pro M4 with iPhone input via Universal Control/Continuity

## Steps to Reproduce
1. Start Happy on Mac
2. Begin typing from iPhone (via Universal Control or similar remote input method)
3. Press space twice to trigger local mode switch
4. Happy crashes completely and exits

## Expected Behavior
Happy should gracefully handle the local mode switch even when input device state changes mid-session.

## Actual Behavior
Happy crashes immediately with no error message when attempting to switch to local mode after device input switch.

## Root Cause
When the input device switches mid-session, the TTY state becomes inconsistent. The code calls \`process.stdin.setRawMode()\` without error handling, which throws an exception when stdin is in an inconsistent state due to the device switch.

## Affected Code Locations
- \`dist/index-sSOy3f2x.mjs:2503\` (setRawMode true)
- \`dist/index-sSOy3f2x.mjs:2798\` (setRawMode false)
- \`dist/index-DmJ8WyYo.cjs:2525\` (setRawMode true)
- \`dist/index-DmJ8WyYo.cjs:2820\` (setRawMode false)
- \`dist/runCodex-BnjA1TX6.mjs:951\` (setRawMode true)
- \`dist/runCodex-DQPZNHzH.cjs:953\` (setRawMode true)

Note: The runCodex files already have try-catch when setting to false (lines 1310-1313 and 1312-1314 respectively), but not when setting to true. The main index files have no error handling at all.

## Proposed Fix
Wrap all \`setRawMode()\` calls in try-catch blocks to gracefully handle TTY state inconsistencies:

\`\`\`javascript
if (process.stdin.isTTY) {
  try {
    process.stdin.setRawMode(true); // or false
  } catch (err) {
    logger.debug(\`Failed to set raw mode: \${err.message}\`);
  }
}
\`\`\`

This fix allows Happy to continue operating even when TTY state is inconsistent due to input device switching.

## Workaround
Until fixed, avoid pressing space twice when typing from a remote device, as this triggers the local mode switch that causes the crash.

## Status
[...]

Issue #430, opened 2026-01-12
Competing processes when switching to local mode


fberlakovich:
When switching from local mode to remote mode (or vice versa), the terminal flickers between two states and. Keystrokes are sometimes captured by one mode, sometimes by the other, making the session unusable.

## Environment

  - happy-coder: 0.13.0
  - Claude Code: 2.1.1 (installed via homebrew)
  - Node.js: v24.12.0
  - macOS: 15.1 (Darwin 25.1.0)
  - Terminal: iTerm2

I am not sure whether maybe I am just holding it wrong. 
It seems that two processes are competing for the input/output.
Either way, the following local fix seems to solve the issue

```
  diff --git a/scripts/claude_version_utils.cjs b/scripts/claude_version_utils.cjs
  index 2184917..1a6b638 100644
  --- a/scripts/claude_version_utils.cjs
  +++ b/scripts/claude_version_utils.cjs
  @@ -497,6 +497,18 @@ function runClaudeCli(cliPath) {
               stdio: 'inherit',
               env: process.env
           });
  +
  +        // Forward signals to child process so it gets killed when parent is killed
  +        // This prevents orphaned Claude processes when switching between local/remote modes
  +        const forwardSignal = (signal) => {
  +            if (child.pid && !child.killed) {
  +                child.kill(signal);
  +            }
  +        };
  +        process.on('SIGTERM', () => forwardSignal('SIGTERM'));
  +        process.on('SIGINT', () => forwardSignal('SIGINT'));
  +        process.on('SIGHUP', () => forwardSignal('SIGHUP'));
  +
           child.on('exit', (code) => {
               process.exit(code || 0);
           });
```

In case it is not just a me-problem I am happy to create a PR.

Akeboshiwind:
I have this problem too!
I just wasn't able to articulate exactly what the problem was well enough to file an issue so thanks 😄

fberlakovich:
I am not entirely sure whether my patch is a good fix, as I've seen cases now where the session completely terminates when switching to local mode. It could also be completely unrelated though.

Akeboshiwind:
Trying out on a fork and it seems fine for now.
Will watch out for sessions terminating though, thanks :)

sferich888:
I can also [report](https://github.com/slopus/happy-cli/pull/127#issuecomment-3784202869) that this PR fixes the issues I was seeing related to this!

bra1nDump:
[...]

Issue #424, opened 2026-01-22
After switching back to local mode, the input is starting to getting incomplete.


ryuan-cw:
Description
When using happy-cli with the Opus 4.5 model remotely, switching back to local mode doesn't respond to the full input on local machine.

Environment
happy version: 0.13.0
Using Claude Code from native installer
2.1.17 (Claude Code)
Model: Opus 4.5
OS: Mac
Terminal: iterm2

ryuan-cw:
After i double pressed space to switch to local mode, the input to claude code is incomplete and unrecognizable.

hyacz:
I encountered a similar problem. It seems that when I first switched from Local to Remote, Claude Code wasn't properly terminated and became an orphan process. After switching back to Local, it would compete with the new Claude Code process for stdin.

ryuan-cw:
when is happy going to make a new release???

NvL-WEAP:
Love happy code, but this kills the workflow. #423 suggests a fix.

bra1nDump:
tracking this in #613 - same root cause (input handler persists after remote→local switch). closing here to keep discussion in one place.

PR #570, opened 2026-02-11
fix: restore terminal state after remote→local mode switch

### LINKED ISSUE #301: Bug Report: Terminal input corruption after switching from remote to local mode
## Environment

| Component | Version |
|-----------|---------|
| happy-coder | 0.13.0 |
| Claude Code | 2.0.76 |
| Node.js | v22.19.0 |
| macOS | 26.2 (Build 25C56) |
| Terminal | xterm-24bit |
| Shell | zsh |

## Summary

After switching from remote mode (mobile) back to local mode, terminal input becomes severely corrupted. Characters are dropped/eaten, the status line flashes erratically, and using slash commands (`/`) causes Claude Code CPU to spike to 100%, making the terminal completely unresponsive.

## Steps to Reproduce

1. Start happy-coder in local mode: `happy`
2. Send a message from the mobile app (triggers automatic switch to remote mode)
3. Press double-space to switch back to local mode
   - **Note**: Often requires pressing space 3-5+ times before the switch actually works
4. Attempt to type in local mode

## Expected Behavior

- Double-space switches cleanly to local mode
- Terminal displays chat history normally
- Typing works normally with all characters appearing

## Actual Behavior

1. **Multiple spaces required**: Double-space often doesn't work; need 3-5+ presses
2. **Missing chat history**: Terminal doesn't refresh properly, missing conversation context
3. **Character loss**: Typed characters are dropped/eaten intermittently
4. **Status line flashing**: Duration and context window values flash back and forth
5. **Slash command crash**: Pressing `/` causes Claude Code CPU to spike to 100%
6. **Complete unresponsiveness**: Terminal becomes frozen, requiring `kill` of both Claude Code and happy-coder processes

## Additional Symptom

On exit, an EIO error is emitted:

```
EIO: i/o error, read
      fd: 11,
 syscall: "read",
   errno: -5,
    code: "EIO"
```

## Root Cause Analysis

After investigating the source code, several issues were identified in the remote→local mode transition:

### 1. stdin state not properly reset after remote mode

In `claudeRemoteLauncher` finally block (~line 2980-2992 in `index-B3gQr6vs.mjs`):

```javascript
} finally {
  permissionHandler.reset();
  process.stdin.off("data", abort);
  if (process.stdin.isTTY) {
    process.stdin.setRawMode(false);
  }
  if (inkInstance) {
    inkInstance.unmount();
  }
  // ...
}
```

**Problems:**

- `stdin.pause()` is never called - stdin remains in "flowing" mode
- `setEncoding("utf8")` set during remote mode setup (~line 2693) is never reset
- Keypress listeners may not be fully removed

Compare to remote mode **setup** (~line 2689-2695):

```javascript
if (hasTTY) {
  process.stdin.resume();
  if (process.stdin.isTTY) {
    process.stdin.setRawMode(true);
  }
  process.stdin.setEncoding("utf8");  // This persists!
}
```

### 2. Race condition with Ink unmount

When `inkInstance.unmount()` is called, Ink's internal stdin handling may not fully clean up before `claudeLocal` spawns the new Claude Code process with `stdio: ["inherit"]`.

### 3. Async cleanup not awaited

In the main cleanup handler (~line 5210-5215):

```javascript
stopCaffeinate();
happyServer.stop();    // Not awaited
hookServer.stop();     // Not awaited
cleanupHookSettingsFile(hookSettingsPath);
logger.debug("[START] Cleanup complete, exiting");
process.exit(0);       // Exits while async operations still running
```

This causes the EIO error on exit.

## Attempted Fixes (Unsuccessful)

Added to the finally block:

```javascript
process.stdin.removeAllListeners("keypress");
process.stdin.pause();
```

This did not resolve the issue, suggesting the problem may be deeper - possibly in how Claude Code handles inherited stdin when the terminal is in an inconsistent state, or a race condition with Ink.

## Suggested Investigation Areas

1. **Ink stdin handling**: Check if Ink fully releases stdin control on unmount
2. **Claude Code terminal setup**: How does Claude Code initialize terminal when resuming a session with inherited stdio?
3. **Stream encoding**: The `setEncoding("utf8")` call may affect how Claude Code reads raw terminal input
4. **Timing**: Add delays or proper awaits between Ink unmount and Claude spawn

## Workaround

Currently none - the remote→local switch is essentially unusable. Users must avoid using mobile mode or restart happy-coder after each mobile interaction.

## Log Reference

Relevant log entries showing the mode switch:

```
[01:02:37.816] [loop] Iteration with mode: remote
[01:02:58.064] [remote]: Switching to local mode via double space
[01:02:58.065] [remote]: doSwitch
[01:02:58.066] [remote]: launch finally
[01:02:58.076] [loop] Iteration with mode: local
[01:02:58.079] [local]: launch
[01:02:58.079] [ClaudeLocal] Will resume existing session: ...
```

The switch appears to complete in logs, but terminal state is corrupted.

tomstetson:
## Summary

Fixes the most-reported bug in Happy CLI: terminal input corruption after switching from remote mode (mobile) back to local mode via double-space. Users experience dropped characters, phantom mode switches, and an unresponsive Claude session.

- Fixes #301, #422, #423, #424, #430, #443, #527

## Root cause analysis

Five independent bugs compound to corrupt stdin during the remote→local transition:

| # | Bug | Effect |
|---|-----|--------|
| 1 | **Wrong cleanup order** — raw mode disabled *before* Ink unmounts | Ink expects raw mode during teardown; disabling it first causes Ink to leave the terminal in a half-raw state |
| 2 | **`setEncoding("utf8")` never reset** | This is a sticky, one-way call on the Node.js public API. Once set, stdin delivers strings instead of Buffers to all subsequent consumers, breaking Claude's input handling |
| 3 | **`process.exit()` bypasses finally blocks** | Signal handlers (SIGINT/SIGTERM) call `process.exit()` which skips the finally block that would restore the terminal — leaves it in raw mode |
| 4 | **No signal forwarding to binary children** | When spawning Claude as a binary (e.g. Homebrew install), the parent process doesn't forward signals — child becomes orphan competing for stdin |
| 5 | **stdin left in flowing mode** | No `pause()` after remote cleanup, so stdin data events keep firing into the void |

### Compared to PR #458 (pkill-based process cleanup)

PR #458 addresses the orphan process symptom by adding `HAPPY_SESSION_ID` environment tracking and `pkill`-based cleanup with escalating SIGTERM→SIGKILL. While it correctly identifies the orphan problem, it:

- **Doesn't fix the core stdin corruption** (bugs 1-3 above) — the wrong cleanup order, sticky encoding, and signal handler gaps
- Adds `pkill -f "HAPPY_SESSION_ID=..."` with hardcoded timeouts (100ms → 500ms) which is platform-dependent and fragile
- Registers signal handlers 4+ times ("multiple times to ensure they stick") which is a code smell
- Adds a raw mode toggle hack in `RemoteModeDisplay.tsx` with `setTimeout(100ms)` that papers over the real ordering bug

### Compared to PR #553 (node-pty proxy)

PR #553 takes a nuclear approach: replace `stdio: 'inherit'` with a full PTY proxy via `node-pty`, giving Ink a separate `/dev/tty` fd. This correctly isolates stdin but:

- **Adds a native dependency** (`node-pty`) requiring compilation on every platform — the comment thread already found a PTY fd leak bug in node-pty v1.1.0 (microsoft/node-pty#882)
- Rewrites 200+ lines of `claudeLocal.ts` including IPC socket plumbing to replace fd 3
- Much larger blast radius and review surface for what is fundamentally a cleanup ordering bug

### This PR's approach

Minimal, targeted fixes at each root cause — **94 lines added, 8 removed, zero new dependencies**:

1. **`restoreStdin()` utility** (new file) — idempotent function that properly restores stdin: disable raw mode, pause, reset encoding via `_readableState`, remove orphaned listeners. Every operation wrapped in try-catch.

2. **Fix cleanup order** in `claudeRemoteLauncher.ts` — unmount Ink *first* (while raw mode is still active), *then* call `restoreStdin()`. This is the core fix. Also removed dead code (`process.stdin.off('data', abort)` where `abort` was never a data listener).

3. **Defensive guard** in `claudeLocal.ts` — call `restoreStdin()` before spawning child with inherited stdio. Catches edge cases where remote cleanup was incomplete.

4. **Signal handler protection** in `runClaude.ts` — call `restoreStdin()` at the top of the cleanup function, before any async work or `process.exit()`. Prevents terminal being left in raw mode when signals bypass finally blocks.

5. **Signal forwarding** in `claude_version_utils.cjs` — forward SIGINT/SIGTERM/SIGHUP to binary child processes. Clean up handlers on child exit. Prevents orphan Claude processes.

## Files changed

| File | Lines | Change |
|------|-------|--------|
| `packages/happy-cli/src/utils/restoreStdin.ts` | +53 | **New** — shared stdin restoration utility |
| `packages/happy-cli/src/claude/claudeRemoteLauncher.ts` | +12/-6 | Fix cleanup order, use `restoreStdin()` |
| `packages/happy-cli/src/claude/claudeLocal.ts` | +5/-2 | Defensive `restoreStdin()` before spawn |
| `packages/happy-cli/src/claude/runClaude.ts` | +6 | `restoreStdin()` in signal handler |
| `packages/happy-cli/scripts/claude_version_utils.cjs` | +18 | Signal forwarding to binary children |

## Test plan

- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Full build passes (`yarn build`)
- [x] All 273 existing tests pass (`vitest run`)
- [ ] Manual: Start happy → send message from mobile (enter remote mode) → double-space to switch to local → type a sentence with spaces → verify all characters appear, no phantom mode switches
- [ ] Manual: Ctrl-C twice in remote mode → restart → verify terminal is clean
- [ ] Manual: Rapid switching: remote→local→remote→local
- [ ] Manual: `ps aux | grep claude` after mode switches — verify no orphan processes
- [ ] Manual: Test on both macOS and Linux

### DIFF
--- a/packages/happy-cli/scripts/claude_version_utils.cjs
+++ b/packages/happy-cli/scripts/claude_version_utils.cjs
@@ -497,7 +497,25 @@ function runClaudeCli(cliPath) {
             stdio: 'inherit',
             env: process.env
         });
+
+        // Forward termination signals to the child process.
+        // Without this, the child becomes an orphan competing for stdin
+        // when the parent is killed (e.g., during remote→local mode switch).
+        const forwardSignal = (signal) => {
+            if (!child.killed) {
+                child.kill(signal);
+            }
+        };
+        const signals = ['SIGINT', 'SIGTERM', 'SIGHUP'];
+        for (const sig of signals) {
+            process.on(sig, () => forwardSignal(sig));
+        }
+
         child.on('exit', (code) => {
+            // Clean up signal handlers to avoid leaks
+            for (const sig of signals) {
+                process.removeAllListeners(sig);
+            }
             process.exit(code || 0);
         });
     }
--- a/packages/happy-cli/scripts/claude_version_utils.cjs
+++ b/packages/happy-cli/scripts/claude_version_utils.cjs
@@ -501,9 +501,18 @@ function runClaudeCli(cliPath) {
         // Forward termination signals to the child process.
         // Without this, the child becomes an orphan competing for stdin
         // when the parent is killed (e.g., during remote→local mode switch).
+        // Note: we track exit ourselves rather than using child.killed, because
+        // child.killed becomes true after the first kill() call — not when the
+        // child actually exits. We need to keep forwarding signals (e.g., a
+        // second SIGTERM) in case the child trapped the first one.
+        let childExited = false;
         const forwardSignal = (signal) => {
-            if (!child.killed) {
-                child.kill(signal);
+            if (!childExited) {
+                try {
+                    child.kill(signal);
+                } catch {
+                    // Child already gone
+                }
             }
         };
         const signals = ['SIGINT', 'SIGTERM', 'SIGHUP'];
@@ -512,6 +521,7 @@ function runClaudeCli(cliPath) {
         }
 
         child.on('exit', (code) => {
+            childExited = true;
             // Clean up signal handlers to avoid leaks
             for (const sig of signals) {
                 process.removeAllListeners(sig);
--- a/packages/happy-cli/src/claude/claudeLocal.ts
+++ b/packages/happy-cli/src/claude/claudeLocal.ts
@@ -9,6 +9,7 @@ import { claudeFindLastSession } from "./utils/claudeFindLastSession";
 import { getProjectPath } from "./utils/path";
 import { projectPath } from "@/projectPath";
 import { systemPrompt } from "./utils/systemPrompt";
+import { restoreStdin } from "@/utils/restoreStdin";
 
 /**
  * Error thrown when the Claude process exits with a non-zero exit code.
@@ -177,8 +178,10 @@ export async function claudeLocal(opts: {
 
     // Spawn the process
     try {
-        // Start the interactive process
-        process.stdin.pause();
+        // Ensure stdin is fully clean before handing it to the child process.
+        // This guards against edge cases where remote mode cleanup was incomplete
+        // (e.g., encoding left as utf8, raw mode still active, flowing mode).
+        restoreStdin();
         await new Promise<void>((r, reject) => {
             const args: string[] = []
 
--- a/packages/happy-cli/src/claude/claudeRemoteLauncher.ts
+++ b/packages/happy-cli/src/claude/claudeRemoteLauncher.ts
@@ -6,6 +6,7 @@ import React from "react";
 import { claudeRemote } from "./claudeRemote";
 import { PermissionHandler } from "./utils/permissionHandler";
 import { Future } from "@/utils/future";
+import { restoreStdin } from "@/utils/restoreStdin";
 import { SDKAssistantMessage, SDKMessage, SDKUserMessage } from "./sdk";
 import { formatClaudeMessageForInk } from "@/ui/messageFormatterInk";
 import { logger } from "@/ui/logger";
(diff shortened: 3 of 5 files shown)


leave-a-message:
Hit this twice on macOS (M1). Had to kill the terminal both times to recover. Would love to see this land — thanks for the fix @tomstetson.

surfaceowl:
Really looking forward to this being merged.  For those looking for a temp workaround from your messed up Claude terminal  where raw mode not restored, consider trying this: <br><br><strong>From the broken Claude terminal after exiting Happy.... </strong><br>-hit `CNTL+C` until you exit to shell <br>-run `reset` (a linux/macOS terminal util that fully reinitializes your terminal state)<br>-run `claude --resume`

PR #591, opened 2026-02-13
fix: improve tmux utility behavior and fix -e flag quoting

### LINKED ISSUE #556: Bug: tmux -e flag wraps env var values in quotes, breaking path resolution
## Summary

When the daemon spawns a Claude Code session inside tmux via `spawnInTmux()`, all environment variable values passed through the tmux `-e` flag are wrapped in double quotes. Since `runCommand()` uses Node.js `spawn()` with `shell: false`, these quotes become **literal characters** in the env var values. This breaks path resolution for critical variables like `HOME`, causing authentication failure, webhook timeout, zombie tmux windows, and eventual fallback to non-tmux spawn.

## Root Cause

In `packages/happy-cli/src/utils/tmux.ts` (line 536-539):

```typescript
const escapedValue = value
    .replace(/\\/g, '\\\\')
    .replace(/"/g, '\\"')
    .replace(/\$/g, '\\$')
    .replace(/`/g, '\\`');
createWindowArgs.push('-e', `${key}="${escapedValue}"`);
```

This produces arguments like:
```
tmux new-window -e HOME="/Users/username" -e PATH="/usr/bin:/usr/local/bin" ...
```

Since `runCommand()` (line 278) uses `spawn()` with `shell: false`, there is **no shell to interpret and strip the quotes**. The arguments are passed directly to the tmux process as-is. tmux faithfully sets `HOME` to the literal string `"/Users/username"` (with quote characters included).

## Cascade Failure

1. `HOME` becomes `"/Users/username"` (with literal `"` characters)
2. `configuration.happyHomeDir` resolves to `"/Users/username"/.happy` (invalid path)
3. `readCredentials()` in `persistence.ts` can't find `~/.happy/access.key`
4. `authAndSetupMachineIfNeeded()` in `auth.ts` shows interactive auth prompt
5. Claude Code is stuck on auth prompt — webhook never fires
6. 15s timeout expires in `daemon/run.ts`
7. Zombie tmux window left behind (process + window not cleaned up)
8. Next spawn attempt hits "index N in use" error due to zombie window
9. Falls back to regular (non-tmux) spawn

## Reproduction

```bash
# This demonstrates the issue:
tmux new-window -e 'HOME="/Users/username"' 'echo $HOME'
# Output: "/Users/username"  (with literal quotes!)

# Correct behavior:
tmux new-window -e 'HOME=/Users/username' 'echo $HOME'
# Output: /Users/username  (clean value)
```

From daemon logs:
```
[DAEMON RUN] spawnInTmux returned pid 11749
[DAEMON RUN] Session webhook timeout for PID 11749 (tmux)
[DAEMON RUN] Failed to spawn session in tmux: Session webhook timeout for PID 11749 (tmux)
[DAEMON RUN] Falling back to regular spawn
```

Running `tmux capture-pane` on the zombie window reveals:
```
? How would you like to authenticate?
  Log in with a browser
  Use an API key
  Use Claude Max / Claude Pro
```

### Fix 1: Remove quoting and escaping in `tmux.ts`

Since `runCommand()` uses `spawn(shell: false)`, no shell escaping or quoting is needed. Arguments are passed directly to tmux as array elements:

```typescript
// BEFORE (line 536-539):
const escapedValue = value
    .replace(/\\/g, '\\\\')
    .replace(/"/g, '\\"')
    .replace(/\$/g, '\\$')
    .replace(/`/g, '\\`');
createWindowArgs.push('-e', `${key}="${escapedValue}"`);

// AFTER:
createWindowArgs.push('-e', `${key}=${value}`);
```

tmux `-e` format is `KEY=VALUE` — no quotes around the value.

### Fix 2: Reduce env vars passed via `-e`

In `packages/happy-cli/src/daemon/run.ts` (lines 466-470), the entire `process.env` (~38 variables) is passed via `-e` flags. This is redundant because tmux windows inherit the environment from the tmux server. Only `extraEnv` variables need to be explicitly set:

```typescript
// BEFORE:
const tmuxEnv: Record<string, string> = {};
for (const [key, value] of Object.entries(process.env)) {
  if (value !== undefined) {
    tmuxEnv[key] = value;
  }
}
Object.assign(tmuxEnv, extraEnv);

// AFTER:
const tmuxEnv: Record<string, string> = { ...extraEnv };
```

### Fix 3: Clean up zombie windows on timeout

The timeout handler in `daemon/run.ts` (lines 508-520) only deletes the awaiter — it doesn't kill the zombie process or tmux window. This causes "index in use" errors on subsequent spawn attempts:

```typescript
const timeout = setTimeout(async () => {
  pidToAwaiter.delete(tmuxResult.pid!);
  // Kill zombie process
  try { process.kill(tmuxResult.pid!, 'SIGTERM'); } catch (e) { /* already exited */ }
  // Kill zombie tmux window
  try {
    await tmux.executeTmuxCommand(['kill-window', '-t', `${sessionName}:${windowName}`]);
  } catch (e) { /* best effort */ }
  pidToTrackedSession.delete(tmuxResult.pid!);
  resolve({ type: 'error', errorMessage: `Session webhook timeout for PID ${tmuxResult.pid} (tmux)` });
}, 30_000); // Consider increasing from 15s — shell init in tmux adds overhead
```

## Environment

- macOS 14.7.6 (Darwin 24.6.0)
- tmux 3.6a
- Node.js 25.4.0
- happy-cli built from `main` branch

HirokiKobayashi-R:
## Summary

- Fixes tmux `-e` flag wrapping env var values in quotes, breaking path resolution
- Adds multi-attempt window creation strategy for compatibility across tmux versions
- Prevents duplicate `-t` flag when caller already provides it
- Adds `list-panes` fallback for PID extraction
- Includes 5 new test cases covering all compatibility scenarios

> **Attribution:** This fix was originally authored by [`lusya123`](https://github.com/lusya123) and cherry-picked from their fork. Full credit to the original author.

## Changes

- `packages/happy-cli/src/utils/tmux.ts` — Compatibility improvements for env passing and window targeting
- `packages/happy-cli/src/utils/tmux.test.ts` — 5 new test cases

## Test plan

- [x] New tests cover: modern tmux, no-pid-flags fallback, session-env fallback, legacy tmux, and list-panes PID extraction
- [x] Existing tmux tests continue to pass

Closes #556

### DIFF
--- a/packages/happy-cli/src/utils/tmux.test.ts
+++ b/packages/happy-cli/src/utils/tmux.test.ts
@@ -5,7 +5,7 @@
  * They do NOT require tmux to be installed on the system.
  * All tests mock environment variables and test string parsing only.
  */
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import {
     parseTmuxSessionIdentifier,
     formatTmuxSessionIdentifier,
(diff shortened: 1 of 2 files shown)


HirokiKobayashi-R:
Closing in favor of #614 which addresses the same core issues with a more targeted approach. The test coverage from this PR could be submitted separately if useful.

PR #681, opened 2026-02-20
fix(security): E2E encrypt vendor API tokens client-side

rootux:
## Summary

Vendor API tokens (Anthropic, OpenAI, Gemini) stored via `happy connect` are currently encrypted server-side using `HANDY_MASTER_SECRET`. This means the server operator can decrypt and read any stored API key — contradicting the project's "zero-knowledge encryption" claims.

This PR moves encryption to the client using a purpose-specific key derived from the user's `machineKey` (HMAC-SHA256 → AES-256-GCM), making the server a dumb blob store that cannot decrypt vendor tokens.

### Changes

**CLI (`packages/happy-cli`):**
- **`encryption.ts`**: Add `deriveVendorEncryptionKey()` — derives a purpose-specific AES-256 key from `machineKey` via HMAC-SHA256 to avoid domain overlap with session encryption
- **`api.ts`**: `registerVendorToken()` encrypts tokens client-side before sending; `getVendorToken()` decrypts after receiving, with legacy fallback for pre-migration tokens
- **`daemon/run.ts`**: Daemon self-serves vendor tokens from local credentials instead of receiving plaintext tokens through the server relay (fixes remote/mobile session flow)
- **`commands/connect.ts`**: Updated help text to reflect E2E encryption

**Server (`packages/happy-server`):**
- **`connectRoutes.ts`**: Removed `encryptString`/`decryptString` calls from all vendor token endpoints — POST stores the opaque blob as-is, GET returns it as-is

### Security model before/after

```
BEFORE: CLI → plaintext token → Server encrypts with HANDY_MASTER_SECRET → DB
        CLI ← plaintext token ← Server decrypts with HANDY_MASTER_SECRET ← DB

AFTER:  CLI encrypts with machineKey → encrypted blob → Server stores as-is → DB
        CLI decrypts with machineKey ← encrypted blob ← Server returns as-is ← DB
```

### Test plan

**33 tests across 3 files**, all passing:

#### `encryption.test.ts` — 7 unit tests (pure crypto, no mocks)
| Test | What it verifies |
|------|-----------------|
| `deriveVendorEncryptionKey` produces 32-byte key | Output is correct AES-256 key size |
| Deterministic output | Same machineKey always produces same derived key |
| Domain separation from machineKey | Derived key ≠ input machineKey |
| Different machineKeys → different derived keys | Key uniqueness per device |
| HMAC label domain separation | Fixed label produces consistent, non-trivial output |
| Round-trip encrypt → decrypt | `encryptWithDataKey` + `decryptWithDataKey` recovers original token |
| Wrong key → null | Decrypting with different derived key returns null |

#### `api.test.ts` — 9 new tests (mocked axios + encryption)

**`registerVendorToken` (4 tests):**
| Test | What it verifies |
|------|-----------------|
| Sends encrypted blob, not raw plaintext | Request body contains base64 blob, not the API key |
| Calls `deriveVendorEncryptionKey` with machineKey | Correct key derivation from credentials |
| Calls `encryptWithDataKey` with serialized token | Token is JSON-serialized then encrypted |
| Throws on non-2xx response | Error propagation on server failure |

**`getVendorToken` (5 tests):**
| Test | What it verifies |
|------|-----------------|
| Returns decrypted token from E2E blob | Happy path: decrypt → JSON parse → return |
| Falls back to legacy JSON parsing | When `decryptWithDataKey` returns null (pre-migration token) |
| Returns null for `{ token: null }` | No token registered for vendor |
| Returns null on 404 | Server endpoint missing or vendor not found |
| Handles decryption throw gracefully | Inner try/catch falls through to legacy path |

#### `connectRoutes.test.ts` — 8 integration tests (Fastify `app.inject()`)

**`POST /v1/connect/:vendor/register` (4 tests):**
| Test | What it verifies |
|------|-----------------|
| Stores token blob as-is | Server does NOT modify/decrypt the client-encrypted blob |
| Upsert overwrites on second POST | Re-registration replaces previous token |
| Stores independently per vendor | OpenAI and Anthropic tokens don't collide |
| Rejects without auth (401) | Authentication required |

**`GET /v1/connect/:vendor/token` (4 tests):**
| Test | What it verifies |
|------|-----------------|
| Returns blob exactly as stored | Server does NOT decrypt before returning |
| Returns `{ token: null }` when none exists | No token registered |
| Returns 401 without auth | Authentication required |
| Ownership isolation | Different user cannot read another user's token |

### How to run tests

```bash
# CLI tests (skip globalSetup build with --config override if needed)
cd packages/happy-cli
npx vitest run src/api/encryption.test.ts src/api/api.test.ts

# Server tests
cd packages/happy-server
npx vitest run sources/app/api/routes/connectRoutes.test.ts
```

### Migration

Users with existing stored tokens should re-run `happy connect <vendor>` to re-encrypt with their client key. The `getVendorToken()` method includes a legacy fallback that gracefully handles pre-migration server-encrypted tokens.

### Future considerations (not in this PR)

- **Multi-device schema**: `machineKey` is per-device, so multi-device users would need per-device token storage (`accountId_vendor_machineId` unique constraint)
- **GitHub tokens**: `githubConnect.ts` still uses server-side encryption — separate PR if desired

### Discussion

See https://github.com/slopus/happy/discussions/680 for full security analysis and review.

### DIFF
--- a/packages/happy-cli/src/api/api.test.ts
+++ b/packages/happy-cli/src/api/api.test.ts
@@ -2,16 +2,19 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { ApiClient } from './api';
 import axios from 'axios';
 import { connectionState } from '@/utils/serverConnectionErrors';
+import { encryptWithDataKey, decryptWithDataKey, deriveVendorEncryptionKey } from './encryption';
 
 // Use vi.hoisted to ensure mock functions are available when vi.mock factory runs
-const { mockPost, mockIsAxiosError } = vi.hoisted(() => ({
+const { mockPost, mockGet, mockIsAxiosError } = vi.hoisted(() => ({
     mockPost: vi.fn(),
+    mockGet: vi.fn(),
     mockIsAxiosError: vi.fn(() => true)
 }));
 
 vi.mock('axios', () => ({
     default: {
         post: mockPost,
+        get: mockGet,
         isAxiosError: mockIsAxiosError
     },
     isAxiosError: mockIsAxiosError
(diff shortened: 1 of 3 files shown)


rootux:
Does this hurt performance in any way?                                                                                      
                                                        
⏺ Good question on performance. Short answer: no meaningful impact. Here's why:                                                      
                                                                                                                                     
  - deriveVendorEncryptionKey — HMAC-SHA256 on 32 bytes, takes microseconds                                                          
  - encryptWithDataKey / decryptWithDataKey — AES-256-GCM on a small token (~500-2000 bytes), also microseconds                      
  - Daemon token self-serve — this does add one HTTP call (getVendorToken) during session spawn, but it replaces the existing flow   
  where the mobile app fetched the token first. Net latency is roughly the same, just the initiator changes. And it falls back to the
   RPC token if the call fails.                                                                                                      
                                                                                                                                     
  These are all one-shot operations during happy connect or session spawn — not hot paths.

PR #688, opened 2026-02-21
fix(happy-cli): create per-request MCP transport for SDK >= 1.26.0

### LINKED ISSUE #634: MCP tools fail after first request (SDK v1.26+ stateless transport reuse)
## Bug

MCP tools (e.g. `change_title`) fail on every call after the initial `initialize` handshake. The first request succeeds but all subsequent requests throw:

```
Error: Already connected to a transport. Call close() before connecting to a new transport, or use a separate Protocol instance per connection.
```

## Root Cause

The MCP server in `happy-coder` creates a single `McpServer` instance and attempts to call `mcp.connect(transport)` with a new transport for each HTTP request. In `@modelcontextprotocol/sdk` v1.26+, `Protocol.connect()` throws if `_transport` is already set ([source](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts)).

The current code (simplified):

```javascript
const mcp = new McpServer({ name: "Happy MCP", version: "1.0.0" });
mcp.registerTool("change_title", ...);

let transport = new StreamableHTTPServerTransport({ sessionIdGenerator: void 0 });
await mcp.connect(transport);

  // This throws on the 2nd request because mcp._transport is already set
  const reqTransport = new StreamableHTTPServerTransport({ sessionIdGenerator: void 0 });
  await mcp.connect(reqTransport);  // 💥 "Already connected to a transport"
  await reqTransport.handleRequest(req, res);
});
```

## Fix

Use the factory pattern from the SDK's own [`simpleStatelessStreamableHttp.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/simpleStatelessStreamableHttp.ts) example — create a new `McpServer` + transport per request:

```javascript
const createMcpServer = () => {
  const s = new McpServer({ name: "Happy MCP", version: "1.0.0" });
  s.registerTool("change_title", ...);
  return s;
};

  const mcp = createMcpServer();
  const transport = new StreamableHTTPServerTransport({ sessionIdGenerator: void 0 });
  await mcp.connect(transport);
  await transport.handleRequest(req, res);
  res.on("close", () => {
    transport.close();
    mcp.close();
  });
});
```

## Environment

- `happy-coder`: 0.13.0
- `@modelcontextprotocol/sdk`: 1.26.0 (installed via `^1.22.0`)
- Node: v22
- OS: Ubuntu (WSL2)

## Workaround

I have a local patch script that applies this fix to the compiled dist file after install. Happy to share if useful.

davidrimshnick:
## Summary

- `@modelcontextprotocol/sdk@1.26.0` (released Feb 4, 2026) added a guard that throws when a stateless `StreamableHTTPServerTransport` is reused across requests
- The Happy MCP server was creating a single transport at startup and reusing it for all incoming requests, causing HTTP 500 on every request after the first
- This creates a fresh `McpServer` + `StreamableHTTPServerTransport` per request, following the SDK's own `simpleStatelessStreamableHttp.ts` example

Fixes #634

## Test plan

- [x] Added unit tests in `startHappyServer.test.ts` that verify:
  - Server starts and returns correct url and tool names
  - Single MCP initialize request returns 200
  - **Multiple sequential requests all return 200** (core regression — would have been 500 with the old single-transport approach on SDK >= 1.26.0)
- [ ] Manual: install `happy-coder` with `@modelcontextprotocol/sdk@1.26.0+`, start a session, and verify `mcp__happy__change_title` works on repeated calls

### DIFF
--- a/packages/happy-cli/src/claude/utils/startHappyServer.test.ts
+++ b/packages/happy-cli/src/claude/utils/startHappyServer.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Tests for Happy MCP server
+ * Verifies that the per-request transport pattern works correctly,
+ * especially with @modelcontextprotocol/sdk >= 1.26.0 which forbids
+ * reusing a stateless StreamableHTTPServerTransport across requests.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest'
+import { startHappyServer } from './startHappyServer'
+import http from 'node:http'
+
+function createMockClient() {
+    const messages: unknown[] = [];
+    return {
+        sessionId: 'test-session-123',
+        sendClaudeSessionMessage: (msg: unknown) => {
+            messages.push(msg);
+        },
+        _messages: messages,
+    } as any;
+}
+
+function postJSON(url: string, body: object): Promise<{ status: number; body: string }> {
+    return new Promise((resolve, reject) => {
+        const data = JSON.stringify(body);
+        const req = http.request(url, {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json',
+                'Accept': 'application/json, text/event-stream',
+            },
+        }, (res) => {
+            let body = '';
+            res.on('data', (chunk) => body += chunk);
+            res.on('end', () => resolve({ status: res.statusCode!, body }));
+        });
+        req.on('error', reject);
+        req.write(data);
+        req.end();
+    });
+}
+
+const MCP_INITIALIZE = {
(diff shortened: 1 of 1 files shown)


bra1nDump:
hey @davidrimshnick, sorry we missed this PR — you had the right fix and we should have merged it. we ended up applying the same approach (fresh transport per request) independently. really appreciate the work, and sorry again for not getting back to you sooner 🙏

PR #691, opened 2026-02-21
fix(happy-cli): strip CLAUDECODE env var from daemon-spawned sessions

### LINKED ISSUE #682: Daemon inherits CLAUDECODE env var, causing all remote sessions to crash
## Bug Description

When `happy daemon start` is run from within a Claude Code session (e.g., via VS Code terminal or Claude Code CLI), the daemon process inherits the `CLAUDECODE=1` environment variable. This variable is then propagated to all child session processes spawned by the daemon.

Claude Code detects `CLAUDECODE=1` and refuses to start with the message: *"Claude Code cannot be launched inside another Claude Code session"*, exiting with a non-zero code. This causes every mobile/remote session to immediately fail with **"Process exited unexpectedly"**.

## Root Cause

1. Claude Code sets `CLAUDECODE=1` in the environment of all child processes to prevent nested sessions
2. When `happy daemon start` is run inside a Claude Code session, the daemon inherits `CLAUDECODE=1`
3. The daemon passes `process.env` (including `CLAUDECODE=1`) to all spawned session processes (see `spawnHappyCLI()` in the daemon code)
4. When the session process tries to launch Claude Code via `query()`, Claude Code detects `CLAUDECODE=1` and immediately exits
5. The exit triggers `new Error("Claude Code process exited with code ...")` which is logged as `{}` (because `JSON.stringify(new Error(...))` produces `{}` — Error objects have no enumerable properties)

## Reproduction Steps

1. Open a Claude Code session (terminal, VS Code, etc.)
2. Run `happy daemon start` from within that session
3. Try to start a remote session from the mobile app
4. Observe: session immediately shows "Process exited unexpectedly"

## Evidence

Daemon log shows the cryptic error:
```
[remote]: launch error {}
```

The `{}` is the serialized `Error` object — the actual message (`"Claude Code process exited with code 1"`) is lost due to `JSON.stringify()`.

Checking the daemon's environment confirms the leak:
```bash
cat /proc/<daemon-pid>/environ | tr '\0' '\n' | grep CLAUDECODE
# Output: CLAUDECODE=1
```

## Workaround

Start the daemon from a clean shell (e.g., a tmux session) with the env var unset:

```bash
unset CLAUDECODE CLAUDE_CODE_ENTRYPOINT && happy daemon start-sync
```

## Suggested Fix

In the daemon startup code (or in `spawnHappyCLI()`), explicitly delete `CLAUDECODE` and `CLAUDE_CODE_ENTRYPOINT` from the environment before spawning session processes:

```javascript
const env = { ...process.env, ...extraEnv };
delete env.CLAUDECODE;
delete env.CLAUDE_CODE_ENTRYPOINT;

const happyProcess = spawnHappyCLI(args, {
  cwd: directory,
  detached: true,
  stdio: ["ignore", "pipe", "pipe"],
  env
});
```

## Additional Issue: Error Logging

The logger's `logToFile()` method uses `JSON.stringify(arg)` for non-string arguments. Since `Error` objects have no enumerable properties, they serialize to `{}`, making debugging extremely difficult. Consider using `err.message` or `err.stack` instead.

## Environment

- happy-coder: v0.13.0
- Claude Code: v2.1.50 (native binary)
- OS: Ubuntu Linux (x64)
- Node.js: v22.22.0

davidrimshnick:
## Summary

- When `happy daemon start` is run from within a Claude Code session (VS Code terminal, Claude Code CLI, etc.), the daemon inherits `CLAUDECODE=1` in its environment
- This was passed through to all child session processes via `process.env`
- Claude Code detects `CLAUDECODE=1` and refuses to start with "cannot be launched inside another Claude Code session", causing every remote session to immediately fail with "Process exited unexpectedly"
- Fix: strip `CLAUDECODE` from `process.env` before spawning session processes

Fixes #682

## Test plan

- [x] Added unit tests verifying `CLAUDECODE` is stripped while other env vars are preserved
- [x] Added test verifying `extraEnv` still merges correctly after stripping
- [ ] Manual: start daemon from within a Claude Code session, verify remote sessions work

### DIFF
--- a/packages/happy-cli/src/daemon/run.ts
+++ b/packages/happy-cli/src/daemon/run.ts
@@ -495,14 +495,18 @@ export async function startDaemon(): Promise<void> {
             '--started-by', 'daemon'
           ];
 
-          // TODO: In future, sessionId could be used with --resume to continue existing sessions
-          // For now, we ignore it - each spawn creates a new session
+          // Strip CLAUDECODE from env — if the daemon was started from
+          // within a Claude Code session, child processes inherit it and
+          // Claude refuses to launch ("cannot be launched inside another
+          // Claude Code session").
+          const { CLAUDECODE: _stripped, ...cleanEnv } = process.env;
+
           const happyProcess = spawnHappyCLI(args, {
             cwd: directory,
             detached: true,  // Sessions stay alive when daemon stops
             stdio: ['ignore', 'pipe', 'pipe'],  // Capture stdout/stderr for debugging
             env: {
-              ...process.env,
+              ...cleanEnv,
               ...extraEnv
             }
           });
--- a/packages/happy-cli/src/daemon/stripClaudeCodeEnv.test.ts
+++ b/packages/happy-cli/src/daemon/stripClaudeCodeEnv.test.ts
@@ -1,0 +1,64 @@
+/**
+ * Tests for CLAUDECODE environment variable stripping (regression #682).
+ *
+ * When the daemon is started from within a Claude Code session, the
+ * CLAUDECODE=1 env var is inherited. If passed to child session processes,
+ * Claude Code refuses to launch with "cannot be launched inside another
+ * Claude Code session". The daemon must strip it before spawning.
+ */
+
+import { describe, it, expect } from 'vitest';
+
+describe('CLAUDECODE env stripping (issue #682)', () => {
+    it('should remove CLAUDECODE from env while preserving other vars', () => {
+        const env: Record<string, string> = {
+            HOME: '/home/user',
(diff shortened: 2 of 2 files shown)


davidrimshnick:
> **Merge note:** This PR modifies the non-tmux spawn path in `daemon/run.ts` (replaces the TODO comment with CLAUDECODE env stripping), which overlaps with #1 on the fork (`feat/resume-session-from-mobile`) that removes the same TODO and adds `--resume` args. If both are merged, the second will have a trivial conflict — just keep both the `--resume` args and the CLAUDECODE stripping.

PR #694, opened 2026-02-21
fix: create MCP server per request (SDK v1.26+ compat)

### LINKED ISSUE #634: MCP tools fail after first request (SDK v1.26+ stateless transport reuse)

sparrowfm:
## Summary

- Fixes MCP tools (e.g. `change_title`) failing on every call after the first `initialize` handshake
- Uses the per-request factory pattern from the SDK's own [`simpleStatelessStreamableHttp.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/simpleStatelessStreamableHttp.ts) example
- Creates a fresh `McpServer` + `StreamableHTTPServerTransport` per HTTP request and cleans up on connection close

## Root cause

`@modelcontextprotocol/sdk` v1.26+ throws `"Already connected to a transport"` if `Protocol.connect()` is called when `_transport` is already set. The previous code created a single `McpServer` and `StreamableHTTPServerTransport` at startup and reused them for all requests, breaking after the first one.

## Test plan

- [ ] Start a remote session from the Happy mobile app
- [ ] Verify `change_title` MCP tool works on the first message
- [ ] Send multiple messages — verify the tool continues to work on subsequent calls
- [ ] Verify no `"Already connected to a transport"` or `"Server already initialized"` errors in logs

Fixes #634

### DIFF
--- a/packages/happy-cli/src/claude/utils/startHappyServer.ts
+++ b/packages/happy-cli/src/claude/utils/startHappyServer.ts
@@ -33,53 +33,54 @@ export async function startHappyServer(client: ApiSessionClient) {
     };
 
     //
-    // Create the MCP server
+    // Create a fresh MCP server + transport per request.
+    // @modelcontextprotocol/sdk v1.26+ throws "Already connected to a
+    // transport" if connect() is called twice on the same McpServer instance.
+    // The stateless HTTP pattern requires a new server per request — see:
+    // https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/simpleStatelessStreamableHttp.ts
     //
 
-    const mcp = new McpServer({
-        name: "Happy MCP",
-        version: "1.0.0",
-    });
+    const createMcpServer = () => {
+        const s = new McpServer({
+            name: "Happy MCP",
+            version: "1.0.0",
+        });
 
-    mcp.registerTool('change_title', {
-        description: 'Change the title of the current chat session',
-        title: 'Change Chat Title',
-        inputSchema: {
-            title: z.string().describe('The new title for the chat session'),
-        },
-    }, async (args) => {
-        const response = await handler(args.title);
-        logger.debug('[happyMCP] Response:', response);
-        
-        if (response.success) {
-            return {
-                content: [
-                    {
-                        type: 'text',
-                        text: `Successfully changed chat title to: "${args.title}"`,
-                    },
-                ],
-                isError: false,
-            };
-        } else {
-            return {
-                content: [
-                    {
-                        type: 'text',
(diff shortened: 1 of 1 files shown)


bra1nDump:
hey @sparrowfm, sorry we missed this — you nailed the fix and we should have caught it earlier. we ended up applying the same per-request transport approach independently. thanks for the contribution and apologies for the late response 🙏

PR #706, opened 2026-02-21
feat(sync): add force re-sync messages debug tool in dev menu

seibe:
## Summary

Adds a developer menu action that wipes the local message cache for a specific session and re-fetches all messages from the server. Useful for diagnosing sync issues without a full app reinstall.

### Changes

- `storage.ts`: New `clearSessionMessages(sessionId)` method to remove a session's cached messages from the Zustand store
- `sync.ts`: New `forceResyncMessages(sessionId)` method that stops the existing sync, acquires the message lock, resets `sessionLastSeq` / message queue state, clears the store, then creates a fresh sync and awaits completion
- `dev/index.tsx`: New "Force Re-sync Messages" action in the dev menu with session ID prompt and loading state

## Test plan

- [ ] Dev menu → Force Re-sync Messages → enter session ID → messages reload from server
- [ ] Verify the action properly resets cursor state and re-fetches from seq 0

### DIFF
--- a/packages/happy-app/sources/app/(app)/dev/index.tsx
+++ b/packages/happy-app/sources/app/(app)/dev/index.tsx
@@ -7,13 +7,14 @@ import { ItemList } from '@/components/ItemList';
 import { useRouter } from 'expo-router';
 import Constants from 'expo-constants';
 import * as Application from 'expo-application';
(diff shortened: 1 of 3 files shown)


seibe:
Withdrawing — will reconsider scope.

PR #707, opened 2026-02-21
chore: gitignore .env.dev and remove from tracking

seibe:
## Summary

Removes the tracked `packages/happy-server/.env.dev` file and adds `.env.dev` to `.gitignore`.

The file contained local development configuration values (database URL, master secret, MinIO/S3 credentials, metrics config) that should not be committed to the repository.

## Test plan

- [ ] `.env.dev` is no longer tracked by git
- [ ] New clones do not include the file


seibe:
Withdrawing — will reconsider scope.

PR #730, opened 2026-02-23
fix(happy-app): handle blank lines and empty cells in markdown table parser

### LINKED ISSUE #506: Markdown tables do not render cleanly - empty cells cause column misalignment

Yoowatney:
## Summary

Fixes #506

Two bugs in the markdown table parser:

1. **Blank lines between rows**: LLMs often output tables with `\n\n` between rows. The parser stopped collecting at the first blank line, rendering each row as raw text.
2. **Empty cells filtered out**: `filter(cell => cell.length > 0)` removed interior empty cells (e.g. `| | Header1 | Header2 |` lost the empty first column), causing column misalignment.

### Changes
- Skip blank lines when collecting table rows in `parseTable`
- Add `splitTableRow` helper that strips only leading/trailing pipe artifacts while preserving interior empty cells
- Add 5 unit tests covering both bugs

## Test plan
- [x] 5 new tests pass (blank lines, empty cells, combined case)
- [x] All 444 existing tests pass
- [x] Verified visually in web dev server

### DIFF
--- a/packages/happy-app/sources/components/markdown/parseMarkdownBlock.test.ts
+++ b/packages/happy-app/sources/components/markdown/parseMarkdownBlock.test.ts
@@ -1,0 +1,106 @@
+import { describe, it, expect } from 'vitest';
+import { parseMarkdown } from './parseMarkdown';
+
+describe('parseMarkdownBlock - table parsing', () => {
+
+    it('parses a standard table without blank lines', () => {
+        const md = [
+            '| A | B |',
+            '|---|---|',
+            '| 1 | 2 |',
+        ].join('\n');
+
+        const blocks = parseMarkdown(md);
+        expect(blocks).toHaveLength(1);
+        expect(blocks[0]).toEqual({
+            type: 'table',
+            headers: ['A', 'B'],
+            rows: [['1', '2']],
+        });
+    });
+
+    it('parses a table with blank lines between rows (LLM output)', () => {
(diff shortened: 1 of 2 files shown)


tools-bra1ndump:
Thanks for the clean fix! The blank-line handling is a great catch — LLM output hits that constantly. Tests are solid too. Merging this.

PR #733, opened 2026-02-23
fix: inherit parent permission mode in Task subagent sessions

ascott:
Minimal one-line fix: background Task subagents were having their permission mode reset to default instead of inheriting the parent session's mode (bypassPermissions, acceptEdits, etc.).

### DIFF
--- a/cli/src/claude/claudeRemote.ts
+++ b/cli/src/claude/claudeRemote.ts
@@ -111,7 +111,7 @@ export async function claudeRemote(opts: {
         cwd: opts.path,
         resume: startFrom ?? undefined,
         mcpServers: opts.mcpServers,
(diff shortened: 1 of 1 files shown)


ascott:
> 🤖 _This comment was generated by Claude Code._

The fix for this is already in `main` at line 120 of `packages/happy-cli/src/claude/claudeRemote.ts` using `mapToClaudeMode(initial.mode.permissionMode)`, which properly passes all permission modes through unchanged. This PR takes the same approach slightly differently (direct passthrough vs mapper function) but the behaviour is identical.

Closing as the fix already exists in main. It'll ship in the next stable release — `v0.14.0-0` prerelease already includes it.

Issue #639, opened 2026-02-16
OutgoingMessageQueue head-of-line blocking causes 'one behind' responses with background tasks


nikhilsitaram:
## Description

When background tasks are running (subagents, SessionStart hooks), the mobile app shows responses "one behind" — the result of a background task doesn't appear until the user sends another message, at which point the queued result loads and the new message appears to be ignored.

## Environment

- **happy-coder version:** 0.13.0
- **Platform:** Linux (WSL2 Ubuntu)
- **Node:** v20.20.0

## Expected Behavior

Background task completions should be delivered to the mobile app immediately when they finish, regardless of other queued messages.

## Actual Behavior

1. User sends a message
2. Claude responds, spawning a background subagent (or a SessionStart hook runs a background agent)
3. The background task completes — but the mobile app shows nothing
4. User sends another message
5. The **previous** background result suddenly appears, and the new message seems ignored
6. Pattern repeats — always "one behind"

## Root Cause

The `OutgoingMessageQueue` in `index-B3gQr6vs.mjs` uses strict head-of-line blocking:

### 1. Tool call messages are delayed and blocked (lines ~2835-2843)

```javascript
if (toolCallIds.length > 0) {
  const isSidechain = assistantMsg.parent_tool_use_id !== void 0;
  if (!isSidechain) {
    messageQueue.enqueue(logMessage, {
      delay: 250,        // 250ms delay
      toolCallIds        // blocked until released
    });
    return;
  }
}
```

### 2. Queue processing stops at first unreleased item (lines ~2598-2606)

```javascript
processQueueInternal() {
  this.queue.sort((a, b) => a.id - b.id);
  while (this.queue.length > 0) {
    const item = this.queue[0];
    if (!item.released) {
      break;  // STOPS ENTIRE QUEUE
    }
    // ... send and dequeue
  }
}
```

### 3. Release only happens on tool_result or permission grant (lines ~2761-2763)

```javascript
if (c.type === "tool_result" && c.tool_use_id) {
  ongoingToolCalls.delete(c.tool_use_id);
  messageQueue.releaseToolCall(c.tool_use_id);
}
```

### The cascade

For fast synchronous tool calls, this works fine — the result comes back within 250ms. But for background subagents or SessionStart hooks that run background agents (e.g., memory synthesis), the tool call stays in the queue for seconds or minutes, blocking all subsequent messages behind it.

## Impact

[...]

Issue #625, opened 2026-02-15
CLI doesn't propagate sandboxConfig to session metadata, web app defaults to 'default' permission mode


ZiggyGameDev:
## Bug Description

When `sandboxConfig.enabled: true` is set in `~/.happy/settings.json` along with a profile that has `defaultPermissionMode: "bypassPermissions"`, the **web app** (`app.happy.engineering`) still sends `permissionMode: "default"` with every new session's first messages. This causes permission prompts to appear on the phone/web despite the user configuring bypass mode.

The Android app exhibits the same behavior — both platforms intermittently send `"default"` for new sessions.

## Root Cause

The CLI's `runClaude()` function creates session metadata **without including sandbox configuration**:

```js
// packages/happy-agent/src/index.ts (or dist/index-B3gQr6vs.mjs ~line 5000)
let metadata = {
    path: workingDirectory,
    host: os.hostname(),
    version: packageJson.version,
    os: os.platform(),
    machineId,
    homeDir: os.homedir(),
    happyHomeDir: configuration.happyHomeDir,
    happyLibDir: projectPath(),
    happyToolsDir: resolve(projectPath(), "tools", "unpacked"),
    startedFromDaemon: options.startedBy === "daemon",
    hostPid: process.pid,
    startedBy: options.startedBy || "terminal",
    lifecycleState: "running",
    lifecycleStateSince: Date.now(),
    flavor: "claude"
};
// No sandbox field ^^^
```

Meanwhile, the web app's permission resolution in `packages/happy-app/sources/sync/messageMeta.ts` checks:

```ts
function isSandboxEnabled(metadata): boolean {
    const sandbox = metadata?.sandbox;
    return !!sandbox && typeof sandbox === 'object' && sandbox.enabled === true;
}
```

Since `session.metadata.sandbox` is never set by the CLI, `isSandboxEnabled()` always returns `false`, and the web app falls through to `'default'` for every new session.

## Evidence from Logs

Analyzed all `~/.happy/logs/` files across multiple sessions (v0.13.0). Pattern:

- **439 messages** with `sentFrom` + `permissionMode` metadata
- Both `android` and `web` intermittently send `"default"` (88 and 78 times respectively)
- Every new session starts with `"default"` until the user manually changes it
- The `[loop] Permission mode updated from user message to: default` log confirms the CLI receives `"default"` from the app
- Permission prompts (`sendToAllDevices "Permission Request"`) only occur when `permissionMode: "default"` was sent
[...]

PR #734, opened 2026-02-23
fix: inherit parent permissionMode in Task subagents

### LINKED ISSUE #521: YOLO mode (bypassPermissions) not working - users still asked for permissions
## Problem

Users report that YOLO mode doesn't work properly in the app. When YOLO mode (bypassPermissions) is selected, users are still constantly being asked for permissions.

## Investigation Findings

After extensive investigation of the codebase, I found one bug and added diagnostic logging:

### Bug Found: `permissionHandler.reset()` doesn't reset permission mode

In `packages/happy-cli/src/claude/utils/permissionHandler.ts`, the `reset()` method clears all state except the `permissionMode` field:

```typescript
reset(): void {
    this.toolCalls = [];
    this.responses.clear();
    this.allowedTools.clear();
    this.allowedBashLiterals.clear();
    this.allowedBashPrefixes.clear();
    // BUG: permissionMode is NOT reset!
    ...
}
```

This means if a previous session used a different permission mode, that mode might persist incorrectly into a new session until `handleModeChange` is explicitly called.

### Code Flow Analysis

The expected flow for YOLO mode:
1. App sends message with `meta.permissionMode: 'bypassPermissions'`
2. CLI receives it in `runClaude.ts` and constructs `EnhancedMode`
3. Message is pushed to queue
4. In `claudeRemoteLauncher.ts`, `nextMessage()` calls `permissionHandler.handleModeChange('bypassPermissions')`
5. When a tool call happens, `handleToolCall()` checks `this.permissionMode === 'bypassPermissions'`
6. Returns `{ behavior: 'allow', ... }` immediately - no permission prompt!

### Diagnostic Logging Added

Added debug logging to help diagnose:
- `handleModeChange()` - logs when mode changes
- `handleToolCall()` - logs the current mode when processing a tool call

## Proposed Fix

1. Reset `permissionMode` to `'default'` in the `reset()` method
2. Debug logging to trace the permission flow

## Steps to Reproduce

1. Start a session with YOLO mode selected
2. Send a message that triggers tool calls
3. Observe that permission prompts still appear

## Expected Behavior

When YOLO mode is selected, all tool calls should be automatically approved without showing permission prompts.

## Environment

- Happy CLI
- Mobile app (iOS/Android/Web)

## Files Affected

- `packages/happy-cli/src/claude/utils/permissionHandler.ts`

ascott:
Closes #521

## Problem

When a Task subagent is spawned, the permission mode was reset to `"default"` for any non-`"plan"` mode:

```typescript
permissionMode: initial.mode.permissionMode === "plan" ? "plan" : "default",
```

This means users with `bypassPermissions` set in `~/.claude/settings.json` find that background Task subagents have all tool calls denied — Bash, Read, Grep, Edit, etc. — because the subagent can't respond to interactive permission prompts and the calls just fail silently.

## Root Cause

The condition only preserves `"plan"` mode. All other modes including `"bypassPermissions"` and `"acceptEdits"` are silently downgraded to `"default"` at the subagent session boundary.

## Fix

The fix already landed in `main` (commit `7f178466`) via `mapToClaudeMode()` which properly passes all 7 permission modes through unchanged. This PR adds a guard comment in `claudeRemote.ts` documenting the anti-pattern to prevent future regression, and formally closes #521.

## Reproduction (before fix)

1. Set `"defaultMode": "bypassPermissions"` in `~/.claude/settings.json`
2. Ask Claude to run a background Task subagent that uses Bash or Read
3. Observe: all tool calls denied in the subagent
4. After fix: subagent inherits `bypassPermissions` and runs fully autonomously

## Related

- #625 — web/Android app sending wrong permissionMode on new sessions (same symptom, different entry point)
- #639 — background task response delivery timing (separate but affects same workflows)

### DIFF
--- a/packages/happy-cli/src/claude/claudeRemote.ts
+++ b/packages/happy-cli/src/claude/claudeRemote.ts
@@ -117,6 +117,11 @@ export async function claudeRemote(opts: {
         cwd: opts.path,
         resume: startFrom ?? undefined,
         mcpServers: opts.mcpServers,
+        // IMPORTANT: Use mapToClaudeMode() to translate all 7 PermissionModes to the 4
+        // modes the Claude SDK understands. Do NOT use a ternary like
+        // `=== "plan" ? "plan" : "default"` — that would silently reset bypassPermissions
+        // and acceptEdits to "default", breaking autonomous agent workflows.
+        // See: packages/happy-cli/src/claude/utils/permissionMode.ts
         permissionMode: mapToClaudeMode(initial.mode.permissionMode),
         model: initial.mode.model,
         fallbackModel: initial.mode.fallbackModel,


ascott:
> 🤖 _This comment was generated by Claude Code._

After further investigation, the fix for this is already in `main` as of commit `7f178466` — `mapToClaudeMode()` now properly passes all permission modes through unchanged rather than resetting to `"default"`. It's included in the `v0.14.0-0` prerelease but not yet in stable `v0.13.0`.

Closing this PR as the fix already exists. The relevant tracking issue is #521.

PR #766, opened 2026-02-25
fix(happy-cli): move @types/* packages to devDependencies

HirokiKobayashi-R:
## Summary

Six `@types/*` packages are currently listed in `dependencies` but are only needed at compile time. This moves them to `devDependencies` where they belong:

- `@types/cross-spawn`
- `@types/http-proxy`
- `@types/ps-list`
- `@types/qrcode-terminal`
- `@types/react`
- `@types/tmp`

This reduces the install footprint for end users who `npm install happy-coder` — type definitions are not needed at runtime since TypeScript compiles to plain JavaScript.

> Note: `@types/inquirer` and `@types/node` were already correctly placed in `devDependencies`.

## Test plan

- [ ] `yarn install` resolves without errors
- [ ] `yarn typecheck` (`tsc --noEmit`) passes
- [ ] `yarn build` completes successfully
- [ ] `yarn test` passes


HirokiKobayashi-R:
CI failure is **unrelated to this PR**. The error is a pre-existing type issue in `src/agent/acp/runAcp.ts:748`:

```
error TS2345: Argument of type 'string | null | undefined' is not assignable to parameter of type 'string | undefined'.
```

The CLI Smoke Test has been failing on `main` for the last several commits as well (e.g. `ad2052b7`, `1aa20033`, `16ae1b55`, `599f7115`, `3ef6aa25`).

This PR only moves `@types/*` packages between `dependencies` and `devDependencies` — no source code changes.

PR #775, opened 2026-02-27
fix(happy-app): add type safety for todos array

cjunxiang:
## Summary
- **Fix 1:** Add proper type guard using `Array.isArray()` instead of truthy check to prevent "t.todos.filter is not a function" runtime error in ActiveSessionsGroup.tsx
- **Fix 2:** Add optional chaining (`?.`) when accessing `opts.tool.input.todos` in knownTools.tsx to prevent "undefined is not a function" error
- **Fix 3:** Persist permission mode (yolo) across session syncs in storage.ts - always load saved permission modes from storage, not just on initial load

### Fix 3 Details
The yolo mode was being lost because:
1. User selects yolo mode → saved to persistent storage
2. Session syncs (e.g., after sending a message)
3. Previously: saved permission modes were only loaded on initial app load, so on subsequent syncs the saved yolo mode wasn't restored

The fix ensures saved permission modes are always loaded from storage on every sync. This is safe because the resolution logic prioritizes in-memory state over saved storage:
1. First: `existingPermissionMode` (in-memory, current user selection)
2. Second: `savedPermissionMode` (storage, fallback if in-memory is reset)
3. Third: Server value (session.permissionMode) 
4. Fourth: Default

## Test plan
- [ ] Verify sessions with todos display correctly in ActiveSessionsGroup
- [ ] Verify sessions without todos don't cause errors
- [ ] Check web app console for runtime errors when using TodoWrite tool
- [ ] Select yolo mode, send a message, verify yolo mode persists after sync

### DIFF
--- a/packages/happy-app/sources/components/tools/knownTools.tsx
+++ b/packages/happy-app/sources/components/tools/knownTools.tsx
@@ -1,4 +1,4 @@
-import { Metadata } from '@/sync/storageTypes';
+import { Metadata, TodoItemsSchema } from '@/sync/storageTypes';
 import { ToolCall, Message } from '@/sync/typesMessage';
 import { resolvePath } from '@/utils/pathUtils';
 import * as z from 'zod';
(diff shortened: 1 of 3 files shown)


pamdla:
Error: t.todos.filter is not a function

the same error here , need to be fixed please .

PR #791, opened 2026-02-28
fix(windows-ci): fix Windows build failures and add token-based tmux session matching

ahundt:
## replaced by https://github.com/slopus/happy/pull/792

This was used to figure out why the windows ci was failing and fix it. Three independent fixes in this PR:

### 1. Fix Windows CI smoke test failures

**Root cause:** `npx tsc` fails on Windows in Yarn 1.22 workspace context with:
```
Error: Cannot find module 'D:\a\happy\happy\node_modules\node_modules\typescript\bin\tsc'
```
npx resolves the path with a double-nested `node_modules\node_modules\` which doesn't exist.

**Fix:** Replace `npx tsc` with `tsc` directly in both build scripts. The `typecheck` script already used `tsc` directly and worked correctly on all platforms — this makes `build` consistent.

**Additional CI hardening:**
- `SKIP_HAPPY_WIRE_BUILD=1` during `yarn install` to skip postinstall build on Windows (Yarn 1.22 bug: GitHub Issue #6175 — postinstall runs from wrong directory on Windows)
- Explicit `cd packages/happy-wire && yarn build` step in Windows CI after install, so happy-coder build can find happy-wire's dist files
- Fixed `scripts/postinstall.cjs` to walk up directory tree to find workspace root and run build from package directory directly, avoiding the `yarn workspace` command on all platforms

**Files changed:**
- `packages/happy-wire/package.json`: `npx tsc` → `tsc` in build script
- `packages/happy-cli/package.json`: `npx tsc` → `tsc` in build script
- `scripts/postinstall.cjs`: find workspace root, run `yarn build` from package dir
- `.github/workflows/cli-smoke-test.yml`: skip postinstall + explicit happy-wire build on Windows

### 2. Token-based tmux session matching (HAPPY_SPAWN_TOKEN)

**Root cause:** When daemon spawns a session inside tmux, the tmux shell PID ≠ the node process PID, causing the webhook to never match and session spawning to always timeout.

**Fix:** Generate a random hex token before spawning, pass it via `HAPPY_SPAWN_TOKEN` env var, include it in session metadata on webhook, and match on token first (falling back to PID for non-tmux sessions).

**Files changed:**
- `packages/happy-cli/src/api/types.ts`: add `spawnToken?: string` to Metadata
- `packages/happy-cli/src/utils/createSessionMetadata.ts`: include `HAPPY_SPAWN_TOKEN` from env
- `packages/happy-cli/src/daemon/run.ts`: token-based tracking maps + webhook matching logic

### 3. Clear and actionable session spawn error messages

**Root cause:** Daemon error messages were being swallowed and replaced with a generic "unexpected response" string, hiding the real problem from users.

**Fix:** Propagate the actual `result.errorMessage` from the daemon response; log the raw response structure for any unexpected result types.

**Files changed:**
- `packages/happy-app/sources/app/(app)/new/index.tsx`: throw actual daemon error, not generic string

## CI Results

All checks passing after fixes:

| Job | Result |
|-----|--------|
| smoke-test-linux (20) | ✅ pass |
| smoke-test-linux (24) | ✅ pass |
| smoke-test-windows (20) | ✅ pass |
| smoke-test-windows (24) | ✅ pass |
| typecheck | ✅ pass |

### DIFF
--- a/packages/happy-cli/package.json
+++ b/packages/happy-cli/package.json
@@ -57,7 +57,7 @@
   "scripts": {
     "why do we need to build before running tests / dev?": "We need the binary to be built so we run daemon commands which directly run the binary - we don't want them to go out of sync or have custom spawn logic depending how we started happy",
     "typecheck": "tsc --noEmit",
-    "build": "shx rm -rf dist && npx tsc --noEmit && pkgroll",
+    "build": "shx rm -rf dist && tsc --noEmit && pkgroll",
     "test": "$npm_execpath run build && vitest run",
     "start": "$npm_execpath run build && node ./bin/happy.mjs",
     "cli": "tsx src/index.ts",
--- a/packages/happy-wire/package.json
+++ b/packages/happy-wire/package.json
@@ -30,7 +30,7 @@
   ],
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "build": "shx rm -rf dist && npx tsc --noEmit && pkgroll",
+    "build": "shx rm -rf dist && tsc --noEmit && pkgroll",
     "test": "$npm_execpath run build && vitest run",
     "prepublishOnly": "$npm_execpath run build && $npm_execpath run test",
     "release": "npx --no-install release-it"


ahundt:
this was used to figure out why the windows ci was failing, replaced by https://github.com/slopus/happy/pull/792

PR #818, opened 2026-03-06
fix(mcp): handle OAuth discovery endpoint to prevent 500 from MCP SDK

antoniocuomo99:
## Problem

Claude Code 2.1.70+ probes `/.well-known/oauth-authorization-server` on HTTP MCP servers as part of its OAuth discovery flow.

The MCP TypeScript SDK does not implement this endpoint — instead of returning 404, it throws an unhandled exception that gets caught and turned into a **500 Internal Server Error**.

Claude Code treats the 500 as a fatal error and refuses to connect to the MCP server entirely.

## Fix

Intercept the `/.well-known/oauth-authorization-server` request before passing it to the MCP SDK transport, and return **404** explicitly. This correctly signals to Claude Code that no OAuth is required, and the connection proceeds normally.

This is a client-side workaround until the MCP TypeScript SDK handles this endpoint properly upstream.

## Impact

- Fixes MCP server connectivity with Claude Code >= 2.1.70
- No behavior change for any other endpoints

### DIFF
--- a/packages/happy-cli/src/claude/utils/startHappyServer.ts
+++ b/packages/happy-cli/src/claude/utils/startHappyServer.ts
@@ -86,6 +86,13 @@ export async function startHappyServer(client: ApiSessionClient) {
     //
 
     const server = createServer(async (req, res) => {
+        // Claude Code 2.1.70+ requires HTTP MCP servers to implement OAuth discovery.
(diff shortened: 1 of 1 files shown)


antoniocuomo99:
Closing in favor of #810 which addresses the same issue and includes additional test coverage.

PR #819, opened 2026-03-06
fix(docker): make happy-server Dockerfile buildable & runnable

marxoffice:
**Description**
This patch fixes the Docker build and runtime of *happy‑server* so it can be built and run without manual intervention.

**What changed**

- **Pre‑install patches** – The Dockerfile now copies the required patch files **before** running `yarn install`. This prevents the build from failing due to missing modules.
- **Correct working directory** – The `CMD` is updated to set the correct working directory. This ensures that imports like `@/app` and `path.join(process.cwd(), "prisma", "migrations")` resolve properly when the container runs in standalone mode.

**Result**

After applying this patch, you can simply run:

```bash
docker build -t happy-server -f Dockerfile .
docker run -p 3005:3005 \
  -e HANDY_MASTER_SECRET=<your-secret> \
  -v happy-data:/data \
  happy-server
```

and the container starts successfully, with all imports resolved correctly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ WORKDIR /repo
 
 COPY package.json yarn.lock ./
 COPY scripts ./scripts
+COPY patches ./patches
 
 RUN mkdir -p packages/happy-app packages/happy-server packages/happy-cli packages/happy-agent packages/happy-wire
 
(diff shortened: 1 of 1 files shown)


bra1nDump:
Closing in favor of #911.

We’re going with the source-level migration-path fix instead of changing the container working directory / entrypoint. #911 covers the standalone Docker regressions with the narrower patch set. Thanks for the contribution.

PR #829, opened 2026-03-08
fix(windows): force UTF-8 code page at startup to prevent garbled TUI output

daravi:
## Problem

On Windows, PowerShell and cmd default to legacy code pages (e.g. CP437). Happy's TUI uses UTF-8 box-drawing characters (─, │, ╭, etc.) for its interface, which get misrendered as garbage like `ΓöÇΓöÇΓöÇ` when the terminal isn't in UTF-8 mode.

The workaround was to run `chcp 65001` manually before launching `happy`, but this is not discoverable and breaks the out-of-box experience on Windows.

**Why native `claude` doesn't have this issue:** The native Claude Code installer includes a Windows launcher that sets encoding before startup. Happy, being an npm package wrapping Claude Code, doesn't inherit that setup — its own TUI layer renders first before Claude's encoding setup kicks in.

## Fix

Force UTF-8 (code page 65001) at process startup on Windows — before any TUI rendering occurs:

```ts
if (process.platform === 'win32') {
  try {
    execSync('chcp 65001', { stdio: 'ignore' })
  } catch { /* non-fatal */ }
  process.stdout.setDefaultEncoding('utf8')
  process.stderr.setDefaultEncoding('utf8')
}
```

This is a standard pattern for Node.js CLI tools on Windows and has no effect on non-Windows platforms.

## Note on a related bug

There is a second rendering bug: after typing a long message that wraps in the terminal, the TUI does a full incremental redraw that corrupts the display again (same `ΓöÇ` characters). This fix addresses the initial startup garbling but **not** the line-wrap redraw bug, which appears to be deeper in the TUI/ink rendering layer. That is worth tracking as a separate issue.

## Screenshots

Before (with default PowerShell code page):
> `ΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇ ΓÄ│ΓÄ│ bypass permissions on (shift+tab to cycle)`

After this fix: renders correctly without any manual `chcp` step.

## Test plan
- [ ] Run `happy` on Windows (PowerShell, default code page) — verify TUI renders correctly without running `chcp 65001` first
- [ ] Run `happy` on macOS/Linux — verify no regression (the `win32` guard prevents any impact)

### DIFF
--- a/packages/happy-cli/src/index.ts
+++ b/packages/happy-cli/src/index.ts
@@ -28,11 +28,25 @@ import { handleConnectCommand } from './commands/connect'
 import { handleSandboxCommand } from './commands/sandbox'
 import { spawnHappyCLI } from './utils/spawnHappyCLI'
 import { claudeCliPath } from './claude/claudeLocal'
-import { execFileSync } from 'node:child_process'
+import { execFileSync, execSync } from 'node:child_process'
 import { extractNoSandboxFlag } from './utils/sandboxFlags'
 
 
 (async () => {
+  // On Windows, PowerShell/cmd default to legacy code pages (e.g. CP437) which
+  // misrender the UTF-8 box-drawing characters used by Happy's TUI as garbage
+  // like "ΓöÇΓöÇΓöÇ". Force UTF-8 (code page 65001) at startup so the terminal
+  // renders correctly without requiring users to run `chcp 65001` manually.
+  if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


daravi:
Closing — investigation showed the rendering issues are already fixed in main (ink 6.6.0 incremental rendering fix). The main action needed is publishing a new npm release of happy-coder to pick up ink 6.6.0+.

Issue #564, opened 2026-02-10
Happy MCP server returns 500 on all requests (Linux headless)


alexmorris0n:
## Description

The Happy MCP server (`change_title` tool) starts successfully but returns HTTP 500 Internal Server Error on all requests, making the `mcp__happy__change_title` tool unavailable to Claude Code.

## Environment

- **Happy version:** 0.13.0 (latest)
- **Claude Code version:** 2.1.38
- **MCP SDK version:** 1.26.0 (`@modelcontextprotocol/sdk`)
- **OS:** Linux (Ubuntu, headless server)
- **Node:** v22.22.0

## Steps to reproduce

1. Install Happy CLI globally: `npm install -g happy-coder`
2. Run `happy` on a Linux headless server
3. Happy daemon starts, authenticates, and registers the machine successfully
4. Happy starts its internal MCP server (e.g., `http://127.0.0.1:38905/`)
5. Claude Code is launched with `--mcp-config {"mcpServers":{"happy":{"type":"http","url":"http://127.0.0.1:38905/"}}}`
6. The `mcp__happy__change_title` tool is never available in the Claude Code session

## Diagnosis

The Happy MCP server starts fine (confirmed in logs):
```
[START] Happy MCP server started at http://127.0.0.1:38905/
```

But it returns 500 on all requests, including MCP `initialize`:
```bash
$ curl -sv -X POST http://127.0.0.1:38905/ \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}},"id":1}'

< HTTP/1.1 500 Internal Server Error
< content-type: text/plain; charset=UTF-8
(empty body)
```

The server uses `StreamableHTTPServerTransport` from `@modelcontextprotocol/sdk/server/streamableHttp.js` (in `index-B3gQr6vs.mjs` around line 4828). The error appears to occur inside `transport.handleRequest()` but no error message is returned in the response body.

## Additional context

- The Happy daemon itself works fine — authentication, WebSocket connection, and session tracking all succeed
- There are also persistent file watcher ENOENT errors in the logs (watching for a JSONL file that doesn't exist yet), though this may be unrelated:
  ```
  [FILE_WATCHER] Watch error: ENOENT: no such file or directory, watch '.../.jsonl', restarting watcher in a second
  ```
[...]

PR #830, opened 2026-03-08
fix(cli): close stateless Happy MCP server on response close

liunice:
## Summary

Fix the built-in Happy MCP bridge for stateless HTTP mode by aligning its lifecycle with the MCP SDK example:

- create a fresh `StreamableHTTPServerTransport` per request
- create a fresh `McpServer` per request
- close both on `res.close` instead of reusing them across requests

## Problem

On Linux/NAS environments, Happy's built-in `mcp__happy__change_title` tool could fail with errors like:

- `No matching deferred tools found`
- `No such tool available: mcp__happy__change_title`

The root cause is that the stateless HTTP MCP implementation in `startHappyServer` was reusing a single `StreamableHTTPServerTransport` (and connected MCP server state) across multiple HTTP requests.

## Why this fix

The MCP SDK's stateless Streamable HTTP example creates a new server/transport for each request and performs cleanup on response close. This PR makes Happy follow the same lifecycle pattern.

## Validation

Tested on a Synology NAS / Linux environment where the bug was reproducible:

- before: `mcp__happy__change_title` failed reliably
- after: title updates succeed again
- takeover UI still works as expected after remote control is attached

Also discussed in #564.

### DIFF
--- a/packages/happy-cli/src/claude/utils/startHappyServer.ts
+++ b/packages/happy-cli/src/claude/utils/startHappyServer.ts
@@ -32,64 +32,92 @@ export async function startHappyServer(client: ApiSessionClient) {
         }
     };
 
-    //
-    // Create the MCP server
-    //
+    const createMcpServer = () => {
+        const mcp = new McpServer({
+            name: "Happy MCP",
+            version: "1.0.0",
+        });
 
-    const mcp = new McpServer({
-        name: "Happy MCP",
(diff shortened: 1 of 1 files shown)


bra1nDump:
hey @liunice, sorry we missed this — you identified the right fix. we ended up applying the same per-request transport approach independently. thanks for the contribution and apologies for the late response 🙏

PR #893, opened 2026-03-20
Fix markdown table parsing, selection, and empty cell rendering

chaehyun2:
## Summary
- **Fix column misalignment**: Table parser used `filter(cell.length > 0)` which dropped empty interior cells, causing column shift when the first header cell is empty. Now strips only leading/trailing `|` and preserves all cells.
- **Fix text selection on web**: Table `Text` elements were missing the `selectable` prop, preventing copy & paste on web. All other markdown blocks already had this.
- **Fix empty cell height collapse**: Empty cells had no text content so `lineHeight` didn't apply, making them ~1/3 the height of populated cells. Added `minHeight: 40` to match expected cell height.

## Test plan
- [ ] Render a table with an empty first header cell (e.g. `| | Col A | Col B |`) — columns should align correctly
- [ ] Select and copy table text on web
- [ ] Verify empty cells have the same height as populated cells
- [ ] Verify normal tables (no empty cells) still render correctly

### DIFF
--- a/packages/happy-app/sources/components/markdown/MarkdownView.tsx
+++ b/packages/happy-app/sources/components/markdown/MarkdownView.tsx
@@ -527,6 +527,7 @@ const style = StyleSheet.create((theme) => ({
         borderBottomWidth: 1,
         borderBottomColor: theme.colors.divider,
         alignItems: 'flex-start',
+        minHeight: 40, // padding (8+8) + lineHeight (24) to prevent empty cell collapse
     },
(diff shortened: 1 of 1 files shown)


bra1nDump:
Thanks for the thorough fix here. The parsing part is now covered by #730. The text selection and empty cell height fixes are real though — would you be up for re-submitting those as a separate PR? Closing this one since the parsing overlap would cause merge conflicts.

PR #935, opened 2026-03-29
Auto-approve safe Codex title tool

### LINKED ISSUE #933: happy codex gets stuck on 'Starting task...' after iOS message because hidden change_title approval blocks the session
## Summary
When I run `happy codex` in a local terminal and send the first message from the iOS app, the local CLI stays stuck on `Starting task...` and never progresses.

This appears to happen because Happy injects a `change_title` MCP tool call at the start of the Codex turn, but the approval prompt is never surfaced in either the local CLI or the mobile UI, so the session waits forever.

`happy codex --dangerously-skip-permissions` does not help, because the permission mode from the iOS message is reset back to `default`, and the session still starts with `approval-policy: untrusted`.

This may be related to issue #229, but this report includes a reproducible Codex/iOS flow and concrete logs.

## Reproduction
1. Run `happy codex` locally in a git repo.
2. Open the Happy iOS app and send a simple first message like `hi`.
3. Observe the local CLI.

## Expected
The first Codex turn should start normally, or any required approval should be shown somewhere actionable.

## Actual
The CLI stays on `Starting task...` and the session appears frozen. No approval prompt is shown locally, and there is no usable permission-mode toggle in the iOS app to bypass it.

## Important detail
Running `happy codex --dangerously-skip-permissions` still fails, because the incoming iOS message resets permission mode to `default` and the Codex session starts with `approval-policy: untrusted`.

## Evidence from logs
From `~/.happy/logs/2026-03-30-01-28-38-pid-31257.log`:

- iOS message metadata sets `permissionMode` to `default`
- Happy logs `Permission mode updated from user message to: default`
- Codex session starts with `approval-policy: untrusted`
- Codex calls `mcp__happy__change_title`
- Happy emits an `elicitation_request` asking to allow `change_title`
- That approval is never shown anywhere, and the session remains stuck

Relevant lines:
- line 57: `permissionMode: "default"`
- line 63: `Permission mode updated from user message to: default`
- line 69: `approval-policy: "untrusted"`
- line 91: `function_call ... mcp__happy__change_title`
- line 93: `elicitation_request ... Allow the happy MCP server to run tool "change_title"?`

I also saw the same hidden approval pattern in an earlier session log: `~/.happy/logs/2026-03-30-01-15-58-pid-27395.log` lines 81-83.

## Extra observations
- `happy doctor` reports `Daemon is not running` even while Happy/Codex bridge processes remain alive.
- `~/.happy/settings.json` also showed `onboardingCompleted: false` during debugging.
- `happy auth login --force` in my environment hit an Ink/raw-mode TUI error, though that may be separate from the main bug.

## Environment
- happy-coder: `0.13.0`
- Codex npm package: `0.117.0`
- Platform: macOS darwin arm64
- Node.js: `v22.22.0`
- iOS app used to send the message

## Suggested direction
It looks like one of these should be fixed:
1. Do not require approval for the initial `change_title` tool call in this flow, or
2. Surface the approval reliably in either the CLI or the mobile UI, or
3. Respect `--dangerously-skip-permissions` even when the iOS message metadata includes `permissionMode: default`

Happy to provide more logs if needed.

ohxiaowo:
## Summary
- auto-approve safe Codex-side MCP tool calls such as `change_title` and `save_memory`
- record those auto-approved requests in completed agent state instead of leaving them pending
- add a focused test covering the safe-tool fast path and the normal pending-approval path

## Why
When `happy codex` receives the first mobile message, Happy injects a `change_title` tool call. In the current Codex path that tool call is treated like any other approval-gated action, so if the approval prompt is not surfaced correctly the session gets stuck on `Starting task...`.

Gemini already auto-approves this class of safe tool calls, but Codex did not. This patch makes Codex follow the same safe default for `change_title`-style tools.

Closes #933.

## Testing
- `../../node_modules/.bin/vitest run src/codex/__tests__/permissionHandler.test.ts src/codex/__tests__/executionPolicy.test.ts`

### DIFF
--- a/packages/happy-cli/src/codex/codexAppServerClient.test.ts
+++ b/packages/happy-cli/src/codex/codexAppServerClient.test.ts
@@ -41,6 +41,7 @@ type MockRpcMessage = {
     id?: number;
     method?: string;
     params?: any;
+    result?: any;
 };
 
 function pushJsonLine(stdout: NodeJS.ReadableStream & { push: (chunk: string) => void }, payload: unknown) {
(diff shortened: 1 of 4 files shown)


ohxiaowo:
One clarification on scope, so I do not over-claim what this PR fixes:

This PR is meant to fix the first-turn deadlock caused by the hidden `change_title` approval path in Codex.

What I have now validated locally:
- launch patched `happy codex`
- send the first iOS message (`hi`)
- the local CLI no longer stays stuck on `Starting task...`
- Codex replies normally and the first turn completes

What I am **not** claiming from this PR alone:
- that I have fully proven the visible mobile chat title always changes in the UI after that tool call

So the confidence level here is: this patch fixes the hidden-approval deadlock on the first mobile turn, which was the behavior reported in #933. There may still be separate follow-up work around the broader `change_title` / elicitation handling path, and I am continuing to investigate that separately.

ohxiaowo:
Follow-up update: I pushed one more commit to handle the current app-server `mcpServer/elicitation/request` path as well.

Local validation now covers both pieces:
- the first mobile message no longer gets stuck on `Starting task...`
- `change_title` is handled successfully end-to-end

Confirmed locally:
- the earlier `missing field action` app-server error is gone
- the hidden `change_title` request is auto-approved and executed successfully
- the mobile UI shows the updated title (`General Help`) instead of the original directory name (`resona`)

bra1nDump:
Thank you @ohxiaowo for the thorough fix & testing!

PR #986, opened 2026-04-04
fix(happy-cli,happy-server): avoid false vendor token timeouts during connect

wang-kaopu:
Fixes #983.

## Summary
`happy connect codex` could fail with `timeout of 5000ms exceeded` even when vendor token registration was still progressing normally. The old flow treated registration as a single synchronous HTTP call with a hardcoded 5s client timeout, so it could not distinguish a slow-but-active request from a genuinely stalled one.

This change converts vendor token registration into a small long-task handshake for the `connect register` path. The server now accepts the work, tracks task stage and heartbeat in memory, and exposes task status for polling. The CLI waits on observable progress instead of a fixed request timeout, and reports task stages while waiting.

## What Changed
- switched `registerVendorToken()` to an activity-aware long-task wait loop in the CLI
- changed `POST /v1/connect/:vendor/register` to return `202 Accepted` with task metadata
- added `GET /v1/tasks/:taskId` for task status polling
- added an in-memory long-task store for registration state, stage, and heartbeat tracking
- added CLI progress output for accepted/encrypting/persisting/succeeded stages
- updated API documentation for the new contract
- added regression tests for progress, stall, failure, and heartbeat behavior

## Validation
- dynamic tests passed:
  - CLI: 14 passing
  - Server: 4 passing
- type diagnostics passed for `happy-cli` and `happy-server`

## Known Limitation
The long-task store is intentionally in-memory for this first rollout. That keeps the registration path lightweight for a single-process foreground wait flow, but tasks do not survive server restarts and are not shared across multiple server instances yet.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -86,6 +86,10 @@ Auth flows:
 - `POST /v1/connect/github/webhook`
 - `DELETE /v1/connect/github`
 - `POST /v1/connect/:vendor/register` (`vendor` in `openai | anthropic | gemini`)
+  - Returns `202 Accepted` for long-running registration work with `{ taskId, state, stage, pollAfterMs, heartbeatAt, updatedAt }`
+  - Clients should poll task status until `state` becomes `succeeded` or `failed`
+- `GET /v1/tasks/:taskId`
+  - Returns current long-task status `{ taskId, state, stage, pollAfterMs, heartbeatAt, updatedAt, error? }`
 - `GET /v1/connect/:vendor/token`
 - `DELETE /v1/connect/:vendor`
 - `GET /v1/connect/tokens`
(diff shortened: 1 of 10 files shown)


wang-kaopu:
Follow-up update after addressing Copilot robustness feedback.

Additional changes on top of the original long-task flow:
- defer `connect register` task execution with `setImmediate()` so the route can return `202 Accepted` before synchronous work starts
- sanitize task failure payloads behind a stable `errorCode` instead of returning raw backend error messages
- handle `/v1/tasks/:taskId` `404` in the CLI with a clearer retry message when the in-memory task record is lost
- update the older `happy-app` consumer to support both legacy `{ success: true }` responses and the new long-task polling contract

Validation run locally:
- passed: `cd packages/happy-server && node ../../node_modules/vitest/vitest.mjs run sources/app/api/routes/connectRoutes.test.ts`
- passed: `cd packages/happy-cli && node ../../node_modules/vitest/vitest.mjs run src/api/api.test.ts`
- passed: `cd packages/happy-app && node ../../node_modules/vitest/vitest.mjs run sources/sync/apiServices.spec.ts`
- `tsc --noEmit` still fails in `happy-server`, `happy-cli`, and `happy-app`, but the failures are pre-existing repo-wide issues (`@slopus/happy-wire` resolution, Prisma exports, and unrelated existing type errors), not from these follow-up changes

Current head:
- `1a7a53fe` `fix(happy-cli,happy-server,happy-app): harden connect register task flow`

PR #995, opened 2026-04-06
Fix mobile AskUserQuestion answers in plan mode

kevin1sMe:
## Summary

- Fix mobile `AskUserQuestion` submissions in plan mode so Claude receives structured `answers` instead of an empty selection.
- Pass tool answer data through the permission RPC as `updatedInput`.
- Merge `updatedInput` back into the original Claude tool input before allowing the tool call.

## Problem

When Claude asked a native `AskUserQuestion` multiple-choice question, the mobile app handled submission like this:

1. approve the tool
2. send the selected answer as a normal follow-up text message

That ordering is incorrect for `AskUserQuestion`.

Claude expects the user selections to be returned as part of the tool input itself, using the `answers` field on `AskUserQuestionInput`. Because the tool was approved before the structured answers were attached, Claude completed the tool call with empty answers.

This produced behavior like:

```json
{
  "questions": [...],
  "answers": {}
}
```

and logs like:

```text
User has answered your questions: .
```

The later mobile text message was treated as a normal user message rather than the result of the pending question tool call.

## Fix

- Add `updatedInput?: Record<string, unknown>` to the session permission RPC payload.
- Change `AskUserQuestionView` to submit answers as structured data instead of sending a follow-up chat message.
- Build the answer payload in the shape Claude expects, with the full question text as the key and the selected option labels joined as the value.
- Merge `updatedInput` into the pending Claude tool input in `permissionHandler.ts` before returning `behavior: 'allow'`.

## Reproduction

1. Start a Claude session in `plan` mode.
2. Have Claude ask a native `AskUserQuestion` multiple-choice question.
3. On mobile, select one or more options and submit.
4. Before this fix, Claude may respond as if the user did not select any option.

### Manual

1. Start `happy claude`.
2. Enter a flow where Claude uses `AskUserQuestion` in `plan` mode.
3. Submit an answer from mobile.
4. Confirm Claude continues with the selected answers instead of saying no option was selected.

### Type checking

Ran successfully:

- `yarn workspace happy typecheck`
- `yarn workspace happy-app typecheck`

## Notes

This PR is intentionally scoped to the `AskUserQuestion` answer transport path between the mobile app and Claude.

### DIFF
--- a/packages/happy-app/sources/components/tools/views/AskUserQuestionView.tsx
+++ b/packages/happy-app/sources/components/tools/views/AskUserQuestionView.tsx
@@ -4,7 +4,6 @@ import { StyleSheet, useUnistyles } from 'react-native-unistyles';
 import { ToolViewProps } from './_all';
 import { ToolSectionView } from '../ToolSectionView';
 import { sessionAllow } from '@/sync/ops';
-import { sync } from '@/sync/sync';
 import { t } from '@/text';
 import { Ionicons } from '@expo/vector-icons';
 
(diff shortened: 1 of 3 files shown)


bra1nDump:
Applied in cf645f7f — rebased your changes on top of #1018 (SDK migration) which had landed first and conflicted with your permissionHandler.ts changes. Thanks for the fix @kevin1sMe, the structured answers flow through updatedInput is exactly right.

PR #1008, opened 2026-04-08
fix(sync): prevent concurrent settings loss during sync roundtrip

AaronZ021:
## Summary

Fixes two race conditions in settings synchronization that can cause user settings to be silently lost:

1. **Settings added during POST are dropped**: `updateSettingsSync` clears `pendingSettings` wholesale after a successful POST. If `applySettings()` is called while the POST is in-flight (e.g. user toggles a setting), those new changes are wiped out. This PR snapshots the pending bag before the POST and only removes the keys that were actually sent, preserving anything added concurrently.

2. **Pending settings invisible after fetch**: `fetchSettingsSync` overwrites the local store with server state but does not re-apply locally-pending changes. Until the next `updateSettingsSync` POST completes, the UI reverts to the server value even though the pending change is still queued. This PR re-applies pending settings to the local store immediately after a fetch so the UI stays consistent.

## Changes

- **`updateSettingsSync`**: Snapshot `pendingSettings` before POST; on success, diff against snapshot to preserve concurrently-added keys instead of clearing everything.
- **`fetchSettingsSync`**: After applying server settings, call `applySettingsLocal(this.pendingSettings)` to keep the local store in sync with queued changes.

## Test plan

- [ ] Toggle a setting, then immediately toggle a second setting before the first POST completes — verify both settings are persisted
- [ ] Open the app on two devices, change a setting on device A, then on device B before sync completes — verify no settings are lost
- [ ] Force a settings fetch (e.g. reconnect) while a local change is pending — verify the local UI still reflects the pending change

### DIFF
--- a/packages/happy-app/sources/sync/sync.ts
+++ b/packages/happy-app/sources/sync/sync.ts
@@ -1301,6 +1301,8 @@ class Sync {
         if (Object.keys(this.pendingSettings).length > 0) {
 
             while (retryCount < maxRetries) {
+                // Snapshot what we're about to send so we can detect concurrent changes
+                const sentPending = { ...this.pendingSettings };
                 let version = storage.getState().settingsVersion;
                 let settings = applySettings(storage.getState().settings, this.pendingSettings);
                 const response = await fetch(`${API_ENDPOINT}/v1/account/settings`, {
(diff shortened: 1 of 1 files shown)


bra1nDump:
@AaronZ021 thanks for the fix! Wondering if you were able to test this locally yourself according to your test plan?

PR #1018, opened 2026-04-09
feat: migrate to official claude-agent-sdk and fix session lifecycle bugs

Scoteezy:
## Summary

- **Migrate to official @anthropic-ai/claude-agent-sdk**: Replace custom subprocess wrapper (~800 lines) with the official SDK. `query.ts` shrinks from 418 → 65 lines. Removes `stream.ts`, `utils.ts`, `metadataExtractor.ts`, `prompts.ts`. Fixes race condition in permission handling (toolUseID now comes directly from SDK callback). Plan mode rewritten to use `query.setPermissionMode()` instead of session restart hack. 470/471 tests pass.

- **Allow archiving disconnected sessions**: When a CLI process crashes without sending session-end, the session previously had no way to be archived. Adds `POST /v1/sessions/:id/archive` endpoint, a fallback `sessionArchive()` client function, and shows the Archive button for all sessions — tries RPC kill first, falls back to server archive.

- **Debounce git-status invalidation and retry push-token fetch**: Coalesces rapid git-status invalidations (300ms debounce) to avoid duplicate RPC round-trips. Adds exponential-backoff retry (3 attempts) for push-token fetches.


packages/happy-cli/src/claude/claudeRemote.ts
```diff
@@ -178,6 +184,16 @@ export async function claudeRemote(opts: {
 
                 const systemInit = message as SDKSystemMessage;
 
+                // Session id is still in memory, wait until session file is written to disk
```
bra1nDump: duplicate comment

PR #578, opened 2026-02-11
fix(cli): proxy support with SSL interception compatibility

hztBUAA:
## Background

Fixes #572 - Proxy support for WebSocket and HTTP connections in restricted network environments.

## Problem

In corporate/enterprise networks, traffic often goes through proxy servers. There are two common proxy configurations:

1. **Regular HTTP Proxy** (e.g., Squid default): Uses CONNECT tunnel for HTTPS
2. **SSL Interception Proxy** (e.g., mitmproxy, Squid with ssl_bump): Decrypts and re-encrypts HTTPS traffic

## Root Cause Analysis

We tested different axios proxy configurations and found:

| Configuration | Regular Proxy | SSL Interception |
|--------------|---------------|------------------|
| `httpsAgent: HttpsProxyAgent` + `proxy: false` | ✅ Works | ✅ Works |
| `proxy: { host, port }` (axios native) | ❌ 503 Error | ✅ Works |

**Why the difference?**
- `HttpsProxyAgent` correctly uses HTTP CONNECT tunnel method
- axios native `proxy` config sends direct `GET https://...` requests which regular squid proxies reject

## Solution

Use `HttpsProxyAgent` with `proxy: false` to ensure CONNECT tunneling is used, which works with **both** proxy types.

```typescript
const proxyAgents = getAxiosProxyAgents(url);
await axios.get(url, {
  ...proxyAgents,
  proxy: proxyAgents ? false : undefined  // Disable native proxy, use agent
});
```

### 1. Proxy support with universal compatibility (5c150bb, accc74e, 698f545)
- Use `HttpsProxyAgent` with `proxy: false` for all axios requests
- Works with both regular proxies and SSL interception proxies
- No additional configuration needed for regular proxies

### 2. SIGTERM exit code 143 handling (28b63dd)
- Fixed mode switch crash when mobile sends messages
- Claude CLI catches SIGTERM and exits with code 143 (128+15)
- Now correctly handles both `signal === 'SIGTERM'` and `code === 143`

### 3. Proxy configuration improvements (9730327)
- Fixed HTTP proxy default port: 80 (was incorrectly 1080)
- Added `NO_PROXY=*` wildcard support

## Compatibility Matrix

| Environment | Status | Notes |
|-------------|--------|-------|
| No proxy | ✅ | Returns undefined, no agent created |
| Regular HTTP proxy | ✅ | Uses CONNECT tunnel via HttpsProxyAgent |
[...]

Issue #572, opened 2026-02-11
Mobile app unable to sync with desktop sessions in restricted network environment (HTTP proxy)


hztBUAA:
When using Happy in a network environment that requires HTTP proxy (squid) with domain whitelist, the mobile app can see sessions but cannot sync messages or control the desktop CLI.

## Environment
- Network: Behind HTTP proxy (squid) with domain whitelist
- Desktop: Linux, happy-cli v0.13.0
- Mobile: iOS app

## Symptoms
1. Mobile app can authenticate successfully (QR code scan works)
2. Mobile app can see workspace/sessions listed
3. **Mobile app CANNOT sync any messages from desktop sessions**
4. **Mobile app CANNOT send messages or control sessions**
5. Clicking "View Device" on mobile shows nothing

## Whitelisted Domains
Based on source code analysis:
- `.cluster-fluster.com` (API and WebSocket)
- `.happy.engineering`
- `.expo.dev`, `.expo.io` (mobile updates)
- `.firebasestorage.app`, `fcm.googleapis.com` (push)
- `.anthropic.com` (AI API)

## Questions
1. Is `api.anthropic.com/api/claude_code/organizations` required for mobile sync? We get 403 from Anthropic server.
2. Are there any additional domains needed for mobile-to-desktop sync?
3. Is WebSocket (Socket.IO on `/v1/updates`) supported through HTTP proxies?

This may be related to #462 and #504.

Issue #451, opened 2025-10-14
start failed with proxy env


tomjamescn:
Great job! Thanks.

If there is proxy env var, such as HTTP_PROXY or HTTPS_PROXY, start happy failed.  It show that:
`Error: Request failed with status code 400`

This is a very bad experience that when my internet is behind a proxy.

LightYear512:
I encountered the same thing.

I use binary version of claude code, it not support *_PROXY env in settings.json now because a bug.

So I use HTTP_PROXY or HTTPS_PROXY for claude chat, but now, I only choose one from happy and claude.

So sad.....

362228416:
+1
export https_proxy="http://127.0.0.1:1087"
happy codex
Error: Request failed with status code 400

Issue #94, opened 2025-09-12
Add proxy support


pxlvoid:
I would like to specify the proxy (socks5 / http) through which I can run Claude Code or Codex.

362228416:
+1

PR #1021, opened 2026-04-10
fix(mcp): add NO_PROXY for local MCP servers behind HTTP proxy

LightYear512:
## Summary

- Fixes silent MCP server failure when `HTTP_PROXY` is set without `NO_PROXY`
- Adds `ensureLocalProxyBypass()` utility to inject `127.0.0.1`, `localhost`, and `::1` into `NO_PROXY`/`no_proxy`
- Applied in both `claudeLocal` and SDK `query` spawn paths
- Includes 10 unit tests covering all edge cases

## Problem

In enterprise environments with HTTP proxies, local MCP HTTP servers (bound to `127.0.0.1`) are routed through the proxy and fail silently. The MCP server is marked as "failed" and its tools become invisible — no error is surfaced to the user.

This is a different dimension of the proxy problem from #578 (which handles outbound API/WebSocket connections). This PR fixes the **inbound local** connection path: when Claude Code spawns a child process that needs to reach a local MCP server.

Related issues: #451, #572, #94

## Root Cause

When `HTTP_PROXY`/`HTTPS_PROXY` is set but `NO_PROXY` doesn't include localhost addresses, Node.js routes requests to `127.0.0.1` through the proxy. The proxy can't reach localhost, so the connection fails silently.

## Solution

Before spawning Claude, check if MCP servers are configured and ensure `127.0.0.1`, `localhost`, and `::1` (IPv6 loopback) are in the `NO_PROXY` environment variable.

Key implementation details:
- Writes both `NO_PROXY` (Node.js `http` module) and `no_proxy` (`undici`) variants
- Only appends missing entries (idempotent)
- Only applied when `mcpServers` are configured (minimal intervention)
- In `query.ts`, spreads `process.env` before mutation to avoid polluting the global env
- Covers IPv6 loopback (`::1`) for Node.js 17+ which may resolve `localhost` to `::1`

## Files Changed

| File | Change |
|------|--------|
| `src/claude/utils/proxyBypass.ts` | New — 18 lines, `ensureLocalProxyBypass()` |
| `src/claude/utils/proxyBypass.test.ts` | New — 10 unit tests |
| `src/claude/claudeLocal.ts` | +5 lines — import + call |
| `src/claude/sdk/query.ts` | +9/-1 lines — import + env spread + call |

## Test Plan

- [x] 10 unit tests for `ensureLocalProxyBypass()` (all passing)
  - Empty env, existing NO_PROXY, no_proxy lowercase, both variants, partial presence, IPv6 only, whitespace, idempotency
- [ ] Manual test: set `HTTP_PROXY`, verify MCP tools work
- [ ] Manual test: verify no-op when `NO_PROXY` already includes `127.0.0.1`
- [ ] Manual test: verify no env pollution when no MCP servers configured

### DIFF
--- a/packages/happy-cli/src/claude/claudeLocal.ts
+++ b/packages/happy-cli/src/claude/claudeLocal.ts
@@ -4,6 +4,7 @@ import { createInterface } from "node:readline";
 import { mkdirSync, existsSync } from "node:fs";
 import { randomUUID } from "node:crypto";
 import { logger } from "@/ui/logger";
+import { ensureLocalProxyBypass } from "./utils/proxyBypass";
 import { claudeCheckSession } from "./utils/claudeCheckSession";
 import { claudeFindLastSession } from "./utils/claudeFindLastSession";
 import { getProjectPath } from "./utils/path";
@@ -240,6 +241,10 @@ export async function claudeLocal(opts: {
             const env = {
                 ...process.env,
                 ...opts.claudeEnvVars
+            }
+
+            if (opts.mcpServers && Object.keys(opts.mcpServers).length > 0) {
+                ensureLocalProxyBypass(env);
             }
 
             logger.debug(`[ClaudeLocal] Spawning launcher: ${claudeCliPath}`);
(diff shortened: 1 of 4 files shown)


bra1nDump:
Applied in 4cbbce2f — rebased on top of #1018 (SDK migration) and adapted the query.ts path to use the SDK's `env` option instead of the old spawn env. Also switched from `||` to `??` to respect explicitly empty NO_PROXY values. Tests pass (9/9). Thanks @LightYear512!

PR #1029, opened 2026-04-12
Multi-process server via Socket.IO Redis adapter

bra1nDump:
## What

When `REDIS_URL` is set, `@socket.io/redis-adapter` broadcasts events across server processes via Redis pub/sub. Without it, behavior is identical to today.

Alternative to #917 — same result, ~15 lines instead of ~1,500.

## How

**socket.ts** — attach Redis adapter when available:
```ts
if (process.env.REDIS_URL) {
    const pubClient = new Redis(process.env.REDIS_URL);
    const subClient = pubClient.duplicate();
    io.adapter(createAdapter(pubClient, subClient));
}
```

**eventRouter.ts** — replace local connection Map with Socket.IO rooms:

| Connection type | Room |
|---|---|
| All | `user:{id}` |
| User-scoped | `user:{id}:user-scoped` |
| Session-scoped | `user:{id}:session:{sid}` |
| Machine-scoped | `user:{id}:machine:{mid}` |

Recipient filters become `io.to(rooms).emit()`. No custom backplane, no envelopes, no subscription queueing.

## Test results (minikube, 2 replicas, real LB via `minikube tunnel`)

Clients connect to the cluster LB, not individual pods.

- **Distribution**: k8s LB split 20 clients 10/10 across pods
- **Cross-pod delivery**: session created on pod A, client on pod B receives update — both directions
- **Load**: 50 burst creates, 1100/1100 events, 100% delivery, 216 events/sec
- **Pod kill**: disconnect detected in 49ms, auto-reconnect through LB, surviving pod zero disruption, 20/20 events post-recovery

Run: `./deploy/local.sh` then `node deploy/test-multiprocess.mjs`


bra1nDump:
heads up — this PR broke RPC routing across replicas. the `rpcListeners` Map is per-process, so when the web UI connects to replica B but the daemon is on replica A, `rpc-call` fails with "method not available."

fixed in b42b9c45635bc03a14fc85dc3eab110b6825f19c — RPC registrations now live in Redis with TTL + keep-alive refresh, and calls route cross-replica via `io.to(socketId).emitWithAck()`. standalone mode (no Redis) still uses the in-memory Map as before.

PR #1041, opened 2026-04-13
fix(happy-app): Enter key not working on Windows web app

### LINKED ISSUE #1040: Windows web app: Enter key does not submit message
## Description

On the Happy web app accessed from Windows (browser), pressing `Enter` does not send the message. The only way to submit is clicking the send button with the mouse.

## Expected behavior

`Enter` should submit the message (consistent with standard chat UX on desktop browsers). `Shift + Enter` should insert a newline.

## Actual behavior

`Enter` either inserts a newline or does nothing. No keyboard shortcut sends the message.

## Environment

- Platform: Windows (desktop browser)
- Happy web app

## Related

- #1006 — Same symptom on iPad with external keyboard. Likely shares the same root cause in the input/submit key handling logic.

## Notes

Happy to investigate and submit a PR for this if the maintainers are open to it. Seems like a keyboard event handler fix in the chat input component.

chphch:
## Summary

- Fix Enter-to-send not working on Windows web app (desktop browsers with touch screens)
- Replace `ontouchstart` / `maxTouchPoints` touch detection with `pointer: coarse` media query
- Windows touch-screen laptops now correctly enable Enter-to-send since their primary pointer is `fine` (mouse/trackpad)

## Root Cause

`AgentInput.tsx` line 528 used `'ontouchstart' in window || navigator.maxTouchPoints > 0` to detect touch devices. On Windows laptops with touch screens, `maxTouchPoints > 0` returns `true`, causing Enter-to-send to be disabled even though a physical keyboard is available.

The `(pointer: coarse)` media query checks the **primary** input device rather than touch capability, correctly identifying desktop browsers as non-touch even with a touch screen present.

## Test plan

- [ ] Windows desktop browser (with touch screen) — Enter should send
- [ ] Windows desktop browser (without touch screen) — Enter should send
- [ ] Mobile browser (iOS Safari, Android Chrome) — Enter should insert newline
- [ ] iPad with external keyboard — separate issue (#1006), not affected by this change

Fixes #1040
Related: #1006

### DIFF
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -525,7 +525,9 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
         if (Platform.OS === 'web') {
             // On mobile web (touch devices), Enter should insert a newline since
             // there's no Shift key available. Users send via the send button instead.
-            const isTouchDevice = typeof window !== 'undefined' && ('ontouchstart' in window || navigator.maxTouchPoints > 0);
+            // Use pointer:coarse media query instead of ontouchstart/maxTouchPoints
+            // to avoid false positives on Windows touch-screen laptops with keyboards.
+            const isTouchDevice = typeof window !== 'undefined' && window.matchMedia('(pointer: coarse)').matches;
             if (agentInputEnterToSend && event.key === 'Enter' && !event.shiftKey && !isTouchDevice) {
                 if (props.value.trim()) {
                     if (isSendBlocked) {


chphch:
**Update**: After a system restart, the Enter key works fine again on Windows web. I was unable to reliably reproduce the original issue.

The fix itself is still technically an improvement — `window.matchMedia('(pointer: coarse)')` is a more accurate way to detect touch-primary devices than `ontouchstart`/`maxTouchPoints`, which can return true on Windows laptops with touch screens even when a keyboard is the primary input.

That said, since I can't confirm this was the actual root cause, feel free to close this if you'd prefer to wait until someone reports a reproducible case. Or merge it as a defensive improvement — up to you.

bra1nDump:
thanks for the fix!

PR #1064, opened 2026-04-14
fix(cli): restart daemon when launching Happy from a different project root

iops-leo:
Happy could reuse an already-running daemon that had been started from a different working directory than the current `happy` launch. When a remote/mobile client then opened a child folder relative to the current project root, the daemon could resolve that path against stale cwd context, which matches the symptom where the session UI looked normal but the spawned chat never produced replies. This change persists the daemon startup cwd and restarts the daemon when Happy is launched from a different project root so relative remote paths stay anchored to the folder the user actually launched from.

## What changed
- persist `startedFromCwd` in local daemon state
- compare daemon startup cwd vs current `process.cwd()` in `ensureDaemonRunning()`
- restart the daemon when the cwd changed even if the CLI version matches
- add focused tests for unchanged cwd, changed cwd, legacy state without cwd, and no-daemon startup

## Proof
```text
pnpm --filter happy typecheck
> happy@1.1.6 typecheck /Users/leo/project/happy/packages/happy-cli
> tsc --noEmit
```

```text
pnpm --filter happy exec vitest run src/daemon/ensureDaemonRunning.test.ts
✓ |unit| src/daemon/ensureDaemonRunning.test.ts (4 tests)
Test Files  1 passed (1)
Tests       4 passed (4)
```

## Manual verification to run
1. Start `happy` from one project root so the daemon is already running.
2. Launch `happy` from a different project root.
3. From web/mobile, open a child folder under that second root and send a message.
4. Confirm the session now responds normally instead of hanging with no reply.

## Notes
- This PR is focused on the daemon cwd reuse regression only.
- I have not yet captured end-to-end mobile/web proof from a live authenticated machine, so I am opening this as a draft first.

### DIFF
--- a/packages/happy-cli/src/daemon/ensureDaemonRunning.test.ts
+++ b/packages/happy-cli/src/daemon/ensureDaemonRunning.test.ts
@@ -1,61 +1,101 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-const mocks = vi.hoisted(() => ({
-  mockLoggerDebug: vi.fn(),
-  mockIsDaemonRunningCurrentlyInstalledHappyVersion: vi.fn(),
-  mockSpawnHappyCLI: vi.fn(),
-}))
-
-vi.mock('@/ui/logger', () => ({
-  logger: {
-    debug: mocks.mockLoggerDebug,
-  },
-}))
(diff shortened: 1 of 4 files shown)


iops-leo:
Manual verification update:

- Reproduced the original workflow after the fix.
- Launched `happy` from the intended project root.
- From web/mobile, opened a child folder under that root and sent a message.
- The spawned session responded normally instead of hanging with no reply.

This confirms the daemon cwd restart fix resolves the reported regression in my environment.

PR #1065, opened 2026-04-14
fix: UI improvements, markdown overflow fix, and MCP/skills passthrough

Scoteezy:
## Summary

- UI improvements and theme updates across multiple app components
- Fix markdown code blocks and tables overflowing beyond screen width — now properly constrained with horizontal scroll
- Add horizontal wheel-to-scroll for code blocks and tables on web
- Small tables retain their natural width instead of stretching to full width
- MCP and skills commands now pass through to Claude Code natively in local mode instead of being intercepted
- Skills metadata forwarded as-is from Claude Code init

## Changes

**happy-app:**
- Constrain agent message and tool containers with `maxWidth: 100%`
- Code blocks use `width: 100%` with native horizontal ScrollView scrollbar
- Tables use `maxWidth: 100%` + `alignSelf: flex-start` so small tables stay compact
- Add `useHorizontalWheelScroll` hook for web — converts vertical wheel to horizontal scroll on code blocks and tables
- UI updates to AgentInput, ChatList, ActiveSessionsGroupCompact, theme, and unistyles

**happy-cli:**
- `/mcp` and `/skills` commands pass through to Claude Code in local mode
- Only intercept these commands in remote mode
- Track current run mode (`local`/`remote`) for command routing
- Forward skills metadata as-is from system init

### DIFF
--- a/packages/happy-app/sources/unistyles.ts
+++ b/packages/happy-app/sources/unistyles.ts
@@ -1,7 +1,7 @@
 import { StyleSheet, UnistylesRuntime } from 'react-native-unistyles';
 import { darkTheme, lightTheme } from './theme';
 import { loadThemePreference } from './sync/persistence';
-import { Appearance } from 'react-native';
+import { Appearance, Platform } from 'react-native';
 import * as SystemUI from 'expo-system-ui';
 
 //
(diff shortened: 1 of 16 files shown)


bra1nDump:
Add useHorizontalWheelScroll hook for web — converts vertical wheel to horizontal scroll on code blocks and tables

- should only be worn shift holding - try other websites

packages/happy-app/sources/unistyles.ts
```diff
@@ -79,4 +79,19 @@ const setRootBackgroundColor = () => {
 };
 
 // Set initial background color
-setRootBackgroundColor();
\ No newline at end of file
+setRootBackgroundColor();
+
+// Re-sync theme when tab becomes visible (web only — Appearance API may miss changes while hidden)
+if (Platform.OS === 'web' && themePreference === 'adaptive') {
+    document.addEventListener('visibilitychange', () => {
```
bra1nDump, line 86: but most of the app still reloads the theme - how come? feels like only the chat background + text color doesn't. are we sure this is the proper fix?

packages/happy-cli/src/claude/runClaude.ts
```diff
@@ -354,6 +355,48 @@ export async function runClaude(credentials: Credentials, options: StartOptions
             return;
         }
 
+        if (specialCommand.type === 'mcp' || specialCommand.type === 'skills') {
+            // In local mode, let Claude Code handle these commands natively
+            if (currentRunMode === 'local') {
+                logger.debug(`[start] /${specialCommand.type} in local mode — passing through to Claude Code`);
```
bra1nDump, line 361: how would we even intercept the commands in local mode? not following - either way local we just don't want to touch at all

bra1nDump (approved):

Issue #45, opened 2025-08-24
Camera is still capturing after successfully capturing QR code


JamesEArnold:
I used the Mobile method to capture the QR code from my terminal when signing up.  The camera never properly turned off after capturing the QR code though, even when on different views.  When I opened the app again, the camera persisted until I revoked the permission in iOS settings.

ex3ndr:
Cant reproduce, what iOS version is it? and what device?

JamesEArnold:
iPhone 15 Pro - iOS 18.16.1.  I'll try to reproduce on my end and provide steps.

JamesEArnold:
I didn't do it like this initially but I was able to reproduce it like so and you can see the green active camera icon after reopening the app:

https://github.com/user-attachments/assets/f5429682-3478-4beb-bf5e-2fc06ec0277c

skyberk:
This is why I removed the app seems weird

prefrontalsys:
Seeing this on iOS as well. The green dot persists after QR scan until the app is force-quit.

Root cause is likely React Navigation screen caching — the scanner screen doesn't unmount when you navigate away, so the `<CameraView>` component keeps the `AVCaptureSession` alive. Conditionally rendering `<CameraView>` based on screen focus via `useFocusEffect` should fix it (same pattern recommended in https://github.com/expo/expo/issues/25181).

PR #1063, opened 2026-04-14
feat: add 'max' effort level and propagate effort setting to Claude SDK

chphch:
Closes #1036, #1005.

**What this fixes:** The Claude effort level menu (`low` / `medium` / `high`) is currently a dead UI control on this app — the selected value never reaches the Claude SDK, so changing it has no effect on Claude's behavior (#1005). This PR adds `max` to the option list per #1036 and wires the selection through the full pipeline (UI → message meta → CLI → `EnhancedMode` hash → SDK `effort` option) so the choice actually changes the API call. Effort changes invalidate the SDK session hash, so switching effort starts a fresh session — same pattern already used for model changes.

**Why one PR (not two):** #1036 and #1005 are coupled — adding `max` without the propagation fix is a no-op label, and fixing the propagation without surfacing `max` leaves a discoverable UX gap. The end-to-end story (UI ↔ meta ↔ CLI ↔ SDK) shares most touched files; splitting would create two diffs reviewers must read together anyway. Happy to split if maintainer prefers.

## Proof

Effort row cycles `low → medium → high → max`, confirming the new `max` option is exposed in the UI (SDK propagation is covered by the green CLI smoke test + typecheck):

effort level cycle including max

## End-to-end flow

```
User selects effort in UI
  → updateSessionEffortLevel() saves to local storage
  → resolveMessageModeMeta() reads effortLevel
  → sendMessage() includes it in message meta
  → runClaude.ts reads meta.effortLevel, puts in EnhancedMode (hash includes effort → new session on change)
  → claudeRemote.ts passes as SDK QueryOptions.effort
  → Claude Agent SDK applies effort to API call
```

## Changes

**UI (happy-app)**
- `modelModeOptions.ts`: Add `{ key: 'max', name: 'max' }` to `getClaudeEffortLevels()`
- `messageMeta.ts`: `resolveMessageModeMeta()` now reads/returns `effortLevel`
- `sync.ts`: `sendMessage()` includes `effortLevel` in message `meta`
- `new/index.tsx`: Saves selected effort to session storage on creation (same pattern as permission mode/model)
- `storageTypes.ts`: Comment update

**CLI (happy-cli)**
- `loop.ts`: Add `effortLevel?: string` to `EnhancedMode`
- `runClaude.ts`: Read `message.meta.effortLevel`, include in mode hash
- `claudeRemote.ts`: Pass `effortLevel` as `effort` in `QueryOptions`
[...]

PR #1077, opened 2026-04-15
feat(app): directory autocomplete for the new-session path pickers

### LINKED ISSUE #44: Add autocomplete for filepaths when creating new session
It would be really nice if the same filepicker autocomplete from the AgentInput was also available for the creating a new session with the daemon.

Image

See my experience report video: I have to switch back to my laptop terminal to find the path. I would be completely blocked here if I was away from my computer.

https://youtu.be/Pa1lQ1EMXQw?t=402

### LINKED ISSUE #970: [Feature Request] Search/filter folder names when typing path in new chat
## Description

When creating a new chat session, the path input field requires typing the full directory path manually. Add search/filter functionality so users can quickly find and select directories.

## Use Case

Users with deep or complex directory structures need to manually type or remember full paths. A search/filter would make it much faster to find the right project directory.

## Suggested Behavior

- As the user types in the path field, show filtered suggestions matching the input
- Filter should match against directory names (not just prefix, but substring/fuzzy)
- Selecting a suggestion fills in the path

---

chphch:
**Asked for twice, and the live request is still open.**

> "I have to switch back to my laptop terminal to find the path. I would be completely blocked here if I was away from my computer."
> — GrocerPublishAgent, #44, 2025-08-24 (with a screenshot and a recorded walkthrough)

> "the path input field requires typing the full directory path manually … As the user types in the path field, show filtered suggestions matching the input."
> — cruzanstx, #970, 2026-04-03

A note on #44, since it reads as resolved: it was closed on 2026-01-22 by merge commit 6e7c2a75, whose message is `Merge pull request #45 from CodingCanuck/vuln-restrict-file-access — Fix #44: Restrict file access in RPC handlers`. That commit hardens RPC file access (`pathSecurity.ts` plus handlers, 5 files under `cli/src`) and has nothing to do with path autocomplete, so the `Fix #44` reads as a number collision — the 45 is the PR number in that contributor's fork, not an issue here. The request itself was never addressed; #970 is the live one.

---

Closes #970. (Also requested first in #44 — see the note above.)

Live subdirectory autocomplete in the new-session **Project path** picker. As the user types a path, the picker lists matching subdirectories from the selected machine and shows them above the existing **Recent** list; tapping a suggestion appends `/` so you can keep drilling down.

## How it works

```
User types: ~/proj
        ↓
split → parentDir = ~/        prefix = proj
        ↓
resolve → /home/user/         (via resolveAbsolutePath + machine.metadata.homeDir)
        ↓
machineBash: ls -1ap "/home/user/" | grep '/$' | grep -v '^\.\.\?/$'
        ↓
filter by prefix → ["projects/", "project-alpha/"]
        ↓
suggestion tapped → value becomes ~/projects/   ← user's own form preserved
```

Debounced 250 ms, with a per-`(machineId, resolvedParentDir)` cache (10 s TTL). No network or disk hit when no machine is selected or the field is empty.

Two details that are load-bearing rather than incidental:

- **`~` is resolved before the shell sees it.** Bash does not expand `~` inside double quotes, so passing the user's typed parent straight into `ls -1ap "${dir}"` silently returns nothing for every `~/…` path — which is the default state of this field, since it renders home-relative. The parent is resolved through `resolveAbsolutePath(parentDir, homeDir)` (`homeDir` comes from `machine.metadata`) before it reaches `machineBash`, while the **user-typed** parent is preserved in the returned `fullPath` so drill-down stays home-relative. The cache is keyed on the *resolved* parent so `~/` and `/home/user/` can't compete for one slot.

- **The machine-level RPC boundary this depends on is already upstream.** `apiMachine.ts` used to validate incoming `bash` calls against `process.cwd()` — the directory the daemon happened to be started from, typically a project subdir — so `validatePath` rejected any cwd outside that subtree and autocomplete died the moment you typed `~/` or `/Users/…`. Upstream removed that boundary outright in `d1cdc796` (`registerCommonHandlers(this.rpcHandlerManager, null)`), for an unrelated symptom (worktree creation reporting *Not a Git repository*). This PR therefore changes no behaviour there; it only adds a two-line comment recording that new-session autocomplete now depends on that boundary staying unrestricted, so a future re-sandboxing does not silently break it.

The pure split/resolve logic lives in `useDirSuggestions.utils.ts` so it is unit-testable — the hook itself transitively imports react-native through `@/sync/ops`, which the test environment can't parse.

## Proof

Real `happy-cli` daemon, no mocks: standalone `happy-server` (PGlite) + Expo web + a paired daemon **started from a deep subdirectory of `$HOME`** — the cwd that broke autocomplete before upstream lifted the machine-level path boundary.

directory autocomplete

Frames: empty `~` → `~/ha` (narrowed to `happy*` worktrees) → `~/happy-` (full worktree list). Every suggestion comes from a real `ls -1ap` over the WebSocket RPC. The daemon log confirms `ls -1ap "/home/hyunwoojung/"` ran with `cwd: '/home/hyunwoojung'` and returned `success: true, stdoutLen: 1565` — i.e. path validation now allows targets outside the daemon's launch directory.

## Scope

App + CLI. No new dependencies, no server or schema change.

- `useDirSuggestions.ts` / `.utils.ts` — the hook and its pure helpers
- `new/index.tsx` — `machineId` threaded to all three `PathPickerContent` call sites (including the native mobile picker; `machineId` is optional, so omitting it type-checks and would have quietly disabled autocomplete on native)
- `apiMachine.ts` — a two-line comment only; the boundary itself is upstream's (`d1cdc796`), and no behaviour changes here

## Verification

`pnpm typecheck` clean; app suite 1125/1125.

- `useDirSuggestions.spec.ts` — 11 cases: tilde paths (`~/proj`, `~/projects/hap`, `~/`, bare `~`), absolute paths, relative/no-slash input, missing `homeDir` (machine offline), and a Windows-style `homeDir` separator.

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -40,6 +40,7 @@ import { isMachineOnline } from '@/utils/machineUtils';
 import { machineSpawnNewSession, sessionSetAgentModes } from '@/sync/ops';
 import { createWorktree, listWorktrees } from '@/utils/worktree';
 import { resolveAbsolutePath } from '@/utils/pathUtils';
+import { useDirSuggestions } from '@/hooks/useDirSuggestions';
 import { formatPathRelativeToHome, formatLastSeen } from '@/utils/sessionUtils';
 import { useNavigateToSession } from '@/hooks/useNavigateToSession';
 import { useNewSessionDraft } from '@/hooks/useNewSessionDraft';
@@ -494,6 +495,7 @@ function PathPickerContent({
     items,
     value,
     homeDir,
+    machineId,
     onChangeValue,
     onDone,
     embedded = false,
@@ -502,6 +504,7 @@ function PathPickerContent({
     items: PickerItem[];
     value: string | null;
     homeDir?: string;
+    machineId?: string | null;
     onChangeValue: (value: string) => void;
     onDone?: () => void;
     embedded?: boolean;
@@ -510,6 +513,8 @@ function PathPickerContent({
     const inputRef = React.useRef<TextInput>(null);
     const currentValue = value ?? '';
     const [selection, setSelection] = React.useState<{ start: number; end: number } | undefined>(undefined);
+
+    const dirSuggestions = useDirSuggestions(machineId, currentValue, homeDir);
 
     React.useEffect(() => {
         // Embedded mobile pickers are positioned next to their trigger. Opening
@@ -632,6 +637,37 @@ function PathPickerContent({
                 <Text style={[pickerStyles.pathMetaText, { color: theme.colors.textSecondary }]}>
                     using custom path above
                 </Text>
+            )}
+
+            {dirSuggestions.length > 0 && (
+                <>
+                    <Text style={[pickerStyles.sectionLabel, { color: theme.colors.textSecondary }]}>
+                        Suggestions
+                    </Text>
+                    <ScrollView style={pickerStyles.optionList} keyboardShouldPersistTaps="handled">
+                        {dirSuggestions.map((suggestion) => (
+                            <Pressable
+                                key={suggestion.fullPath}
+                                style={(p) => [pickerStyles.option, p.pressed && pickerStyles.optionPressed]}
+                                onPress={() => {
+                                    const nextValue = suggestion.fullPath + '/';
+                                    onChangeValue(nextValue);
+                                    setSelection({ start: nextValue.length, end: nextValue.length });
+                                    setTimeout(() => inputRef.current?.focus(), 0);
+                                }}
+                            >
+                                <Ionicons
+                                    name="folder-outline"
+                                    size={16}
+                                    color={theme.colors.textSecondary}
+                                />
+                                <Text style={[pickerStyles.optionText, { color: theme.colors.text }]}>
+                                    {suggestion.fullPath}
+                                </Text>
+                            </Pressable>
+                        ))}
+                    </ScrollView>
+                </>
             )}
 
             <Text style={[pickerStyles.sectionLabel, { color: theme.colors.textSecondary }]}>
@@ -1638,6 +1674,7 @@ function NewSessionScreen() {
                 items={pathItems}
                 value={selectedPath}
                 homeDir={selectedHomeDir}
+                machineId={selectedMachineId}
                 onChangeValue={setSelectedPath}
                 onDone={closePicker}
                 embedded={sidebarLayout.showSidebar}
@@ -1667,6 +1704,7 @@ function NewSessionScreen() {
         pathItems,
         pickerData,
         selectedHomeDir,
+        selectedMachineId,
         selectedPath,
         setSelectedPath,
         sidebarLayout.showSidebar,
@@ -1698,6 +1736,7 @@ function NewSessionScreen() {
             items={pathItems}
             value={selectedPath}
             homeDir={selectedHomeDir}
+            machineId={selectedMachineId}
             onChangeValue={setSelectedPath}
             onDone={closePicker}
             embedded
@@ -2339,6 +2378,7 @@ function NewSessionScreen() {
                             items={pathItems}
                             value={selectedPath}
                             homeDir={selectedHomeDir}
+                            machineId={selectedMachineId}
                             onChangeValue={setSelectedPath}
                             onDone={closePicker}
                         />
--- a/packages/happy-app/sources/hooks/useDirSuggestions.spec.ts
+++ b/packages/happy-app/sources/hooks/useDirSuggestions.spec.ts
@@ -1,0 +1,94 @@
+import { describe, it, expect } from 'vitest';
+import { splitPathForSuggestions } from './useDirSuggestions.utils';
+
+const HOME = '/home/steve';
+
+describe('splitPathForSuggestions', () => {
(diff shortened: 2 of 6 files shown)


chphch:
## Proof — directory autocomplete drills correctly

Captured locally via standalone `happy-server` (PGlite, port 3055) + Expo web (port 8088), authenticated through `getDevWebQueryCredentials()` dev URL bypass, driven headlessly with Playwright.

`useDirSuggestions` calls out to `machineBash(machineId, 'ls -1ap "<dir>"', dir)` — which requires a connected `happy-cli` daemon. To capture the UI flow without bringing up a CLI, the hook's `fetchDirs` was tap-injected at the `window.__HAPPY_DIR_MOCK__` boundary so a deterministic mock filesystem stands in for the real `ls` output. No production code path changed; the mock is removed after capture.

directory autocomplete

The GIF cycles three states:

1. **`/home/user/proj`** typed → Suggestions: `/home/user/projects`, `/home/user/project-alpha`, `/home/user/project-beta` (filter by `proj` prefix).
2. **Tap a suggestion** → path appended with trailing `/`, drilled into the selected directory.
3. **`/home/user/projects/happy`** typed → Suggestions: `/home/user/projects/happy`, `/home/user/projects/happy-cli`, `/home/user/projects/happy-server` (filter by `happy` prefix at the new depth).

Mock filesystem (substitutes for real `ls` output):

```js
{
  '/home/user/':            ['projects', 'project-alpha', 'project-beta', 'photos', 'docs'],
  '/home/user/projects/':   ['happy', 'happy-cli', 'happy-server', 'web-app', 'mobile-app'],
}
```

Production behaviour matches: with a real machine connected, `useDirSuggestions` calls `machineBash` with the same `ls -1ap | grep '/$' | grep -v '^\.\.\?/$'` pipeline, debounced 250 ms with a 10 s per-(machineId, parentDir) cache, and feeds the same prefix-filter into the same `<Suggestions>` UI rendered above the existing `Recent` list.

<details>
<summary>Capture details</summary>

- Workflow: `~/.claude/skills/happy-pr-proof-capture` v0.3.0.
- Mocking: 1-line tap added to `useDirSuggestions.fetchDirs` for the duration of capture, removed before commit. The hook's debounce/cache/prefix-filter logic runs unchanged — only the `ls` results are stubbed.
- Same `gh release upload pr-proofs` + `gh pr comment --body-file` chain used for #1063 / #1078 proofs.

</details>

chphch:
## Follow-up fix — `~`-rooted paths now actually return suggestions

Real-world regression I caught while dogfooding: typing into the path field on the default state (which renders home-relative, e.g. `~/proj`) produced an **empty** suggestions list, even though the `Recent` section worked. Root cause: `useDirSuggestions` shipped the user's typed parent directly into `ls -1ap "${dir}"`, but **bash does not expand `~` inside double quotes**, so the ls silently returned nothing for any `~/...` prefix.

The earlier proof GIF didn't catch this because the capture mocked `fetchDirs` against an absolute-path filesystem (`/home/user/...`), so the `~`-expansion path was never exercised.

### Fix (commit `4f35abbe`)

- `useDirSuggestions(machineId, pathText, homeDir?)` — accepts the selected machine's `homeDir` (already available on `machine.metadata`).
- Parent dir is resolved through `resolveAbsolutePath(parentDir, homeDir)` before being passed to `machineBash` / used as the `cwd`.
- The **user-typed parent** is preserved in the returned `fullPath`, so tapping `~/projects` fills `~/projects/` (not `/home/user/projects/`) and drill-down stays in home-relative form.
- Cache key switches from `(machineId, parentDir)` → `(machineId, resolvedParentDir)` to avoid `~/` and `/home/user/` competing for the same slot.

### Tests

The pure split/resolve helper is extracted into `useDirSuggestions.utils.ts` so it can be unit-tested with plain vitest (the hook itself transitively imports react-native via `@/sync/ops`, which the test env can't parse). 11 cases added in `useDirSuggestions.spec.ts`:

- Tilde paths: `~/proj`, `~/projects/hap`, `~/`, bare `~`
- Absolute paths: `/home/h`, `/usr/local/b`, bare `/`
- Relative / no-slash input
- Missing `homeDir` (machine offline)
- Windows-style `homeDir` separator

```
 ✓ sources/hooks/useDirSuggestions.spec.ts (11 tests) 5ms
```

Typecheck on the changed files is clean (`pnpm typecheck` against `~/happy` with this branch's files dropped in).

chphch:
### The dock composer now completes paths too — device capture

dock autocomplete

Rebased onto current `main` and extended to the second place a new session's directory gets typed: the session-list dock composer's PROJECT picker, whose `Enter custom path` was a bare `Modal.prompt` with no idea what exists on the machine.

Captured on a Pixel 5 AVD (arm64, 393 dp, API 30) paired to a real `happy` daemon, so the listings above come from an actual `ls` over the machine RPC, not a fixture. Same hook, same 250 ms debounce, same per-directory cache, same `expDirAutocomplete` gate — turning the setting off passes a null machineId, which disables the hook outright and leaves a plain field behind.

Correcting an earlier version of this comment: it said the new modal focuses its field on open "which the plain prompt did not". That was wrong on both halves — `WebPromptModal` already focuses on a 100 ms timer on every platform (the `autoFocus` line next to it is a web-only extra), the new modal does exactly the same, and on a real device the keyboard comes up in both. The emulator not raising it is an emulator artifact, not app behaviour. Focus handling is unchanged by this PR.

PR #1081, opened 2026-04-15
feat: file diffs sidebar, session resume via RPC, analytics toggle & cleanup

### LINKED ISSUE #1028: Mobile (codex): model/effort does not persist in sessions
Environment

Happy app version: (1.7.0)
Android version: 16
Device: Google Pixel 8a
Model Family: Codex

Description

When setting the model/effort within a session in the mobile app, leaving the app and then returning will reset the selection back to `default model / xhigh`.

Expected behavior

Similar to the permission mode, set model & effort should persist within a session.

### LINKED ISSUE #1082: Windows: spawn('node', ...) ENOENT in daemon start-sync — cross-spawn fix from #1022 missed two sites
## Summary

  `happy codex` and `happy claude` both fail immediately on Windows with `Error: spawn node ENOENT` when the CLI tries to
   launch the sync daemon. This is the same Windows-spawn root cause fixed in #1022 (closing #980 / #1016) — but the
  `cross-spawn` migration only covered the `codex` app-server spawn site. **Two `spawn("node", ...)` call sites remain in
   `dist/` for v1.1.6** and reproduce the bug.

  ## Environment

  - OS: Windows 11 Pro 10.0.26200 (x64)
  - Shell: PowerShell 7
  - Node: v24.2.0
  - happy: 1.1.6 (latest, installed via `npm i -g happy`)
  - node.exe location: `C:\Program Files\nodejs\node.exe` (on PATH; `where.exe node` resolves correctly)

  ## Reproduction

  ```powershell
  PS> happy --version
  happy version: 1.1.6

  PS> happy claude
  node:events:485
        throw er; // Unhandled 'error' event
        ^

  Error: spawn node ENOENT
      at ChildProcess._handle.onexit (node:internal/child_process:286:19)
      ...
  {
    errno: -4058,
    code: 'ENOENT',
    syscall: 'spawn node',
    path: 'node',
    spawnargs: [
      '--no-warnings',
      '--no-deprecation',
      'C:\\Users\\<user>\\AppData\\Roaming\\npm\\node_modules\\happy\\dist\\index.mjs',
      'daemon',
      'start-sync'
    ]
  }
  ```

  `happy codex` produces an identical stack — same `spawn node` failure with the same `daemon start-sync` args.

  ## Root cause

  Since Node's CVE-2024-27980 hardening, `child_process.spawn('node', ...)` on Windows no longer falls back to appending
  `.exe`. CreateProcess therefore can't locate `node` (only `node.exe` exists), even though PATH is fully correct
  (`where.exe node` works, `node -v` works, `execSync('node -v')` works).

  This is the same class of bug as #980 / #1016 / #1022, just at different spawn sites.

  ## Unfixed spawn sites in v1.1.6 dist

  Grepping the installed package:

  ```
  dist/index-DR7ZghBK.mjs:597
    spawnWithShell && spawnCommand ? spawnCommand : "node",

  dist/types-vPGliGYi.mjs:814
    const child = spawn("node", [RUNNER_PATH, JSON.stringify(args)], {
  ```

  (Mirrored in the `.cjs` builds at `index-DDeR3Cx8.cjs:618` and `types-D8mxjS52.cjs:817`.)

  The stack trace above matches `types-vPGliGYi.mjs:814` — the daemon spawn invoked via `spawnHappyCLI(["daemon",
  "start-sync"], ...)` from `index-DR7ZghBK.mjs:7182` and `:10174`.

  ## Suggested fix

  Apply the same `cross-spawn` swap from #1022 to these two sites (or pass `shell: true` / use `'node.exe'` literally).
  One consolidated `spawnCrossPlatform` helper used everywhere would prevent regression.

  ## Local workaround for other Windows users hitting this

  Until the patch lands, an extension-less hard link makes raw CreateProcess succeed:

  ```powershell
  # Admin PowerShell
  New-Item -ItemType HardLink -Path "C:\Program Files\nodejs\node" -Target "C:\Program Files\nodejs\node.exe"
  ```

  ## Impact

  `happy claude` and `happy codex` are both unusable on Windows + Node 24 out of the box on the current latest release.

### LINKED ISSUE #1084: bug(gemini): safe-yolo mode auto-approves all operations instead of read-only
## Summary

The `safe-yolo` permission mode in the Gemini handler auto-approves **all** operations, despite being documented as "auto-approve read-only operations, ask for write operations."

## Problem

In `packages/happy-cli/src/gemini/utils/permissionHandler.ts:76-79`:

```typescript
case 'safe-yolo':
    // Auto-approve read-only operations, ask for write operations
    // For now, we'll auto-approve everything (can be enhanced later)
    return true;
```

The comment on line 77 describes the intended behavior: read-only auto-approve with write prompts. However, line 79 simply returns `true` for every operation, making `safe-yolo` functionally identical to `yolo`. The comment "can be enhanced later" indicates this was a placeholder that was never completed.

## Impact

Users selecting `safe-yolo` expect write-operation protection. Instead, they get full auto-approve of all tool calls including file writes, deletes, and shell commands — the same behavior as `yolo`. The mode name actively misleads users into a false sense of safety.

## Suggested fix

Implement the documented behavior by checking whether the operation is read-only:

```typescript
case 'safe-yolo':
    // Auto-approve read-only, ask for write operations
    const readOnlyTools = ['read_file', 'list_dir', 'search', 'think', 'save_memory'];
    return readOnlyTools.includes(toolName) || false;
```

Or, if full implementation is deferred, rename the mode to indicate it auto-approves everything (e.g., `yolo-labeled` or remove it from the mode list).

### LINKED ISSUE #1085: bug(codex): auto-approve uses substring matching — potential bypass via crafted tool names
## Summary

The Codex permission handler uses `.includes()` substring matching for auto-approve decisions, which could allow unintended tool calls to bypass approval if tool names contain auto-approved substrings.

## Problem

In `packages/happy-cli/src/codex/utils/permissionHandler.ts:40-42`:

```typescript
private shouldAutoApprove(toolName: string, toolCallId: string): boolean {
    if (CodexPermissionHandler.ALWAYS_AUTO_APPROVE_NAMES.some(
        (name) => toolName.toLowerCase().includes(name.toLowerCase())
    )) {
        return true;
    }
```

The auto-approve list contains `change_title` (line 25). The `.includes()` check means any tool whose name contains the substring `change_title` would be auto-approved — for example, a hypothetical MCP tool named `change_title_and_run_command`.

The same substring pattern exists for `toolCallId` matching on line 45, and in the Gemini handler at `packages/happy-cli/src/gemini/utils/permissionHandler.ts:63`.

## Suggested fix

Switch to exact match:

```typescript
private static readonly ALWAYS_AUTO_APPROVE_NAMES = new Set(['change_title']);

private shouldAutoApprove(toolName: string, toolCallId: string): boolean {
    if (CodexPermissionHandler.ALWAYS_AUTO_APPROVE_NAMES.has(toolName)) {
        return true;
    }
    // ...
}
```

## Context

The current auto-approve list is small (`change_title` only for Codex), so the practical risk today is low. But as the list grows or if MCP tool names from external servers are encountered, substring matching becomes a wider attack surface. PR #935 introduced this pattern to fix the `change_title` deadlock — the fix was correct in intent, this is just about tightening the matching logic.

### LINKED ISSUE #1087: fix(server): token cache grows unbounded + terminal auth requests never expire
## Summary

Two related issues in the server auth layer:

1. The token verification cache grows without bound and never expires entries
2. Terminal auth requests remain retrievable indefinitely after approval

### 1. Unbounded token cache

In `packages/happy-server/sources/app/auth/auth.ts:102-104`:

```typescript
// Cache the result permanently
this.tokenCache.set(token, {
    userId, extras, cachedAt: Date.now()
});
```

The comment explicitly says "permanently." The `cleanup()` method (line 181-186) is a no-op — its comment says "Since tokens are cached 'forever' as requested, we don't do automatic cleanup." The `tokenCache` is a plain `Map` with no TTL, no max size, and no LRU eviction.

**Impact:**
- Memory grows linearly with unique tokens verified. In a multi-replica deployment (PR #1042), each replica accumulates its own cache independently.
- `invalidateUserTokens()` only clears the local process cache — other replicas continue accepting the cached token. Token revocation does not propagate across the cluster.

### 2. Terminal auth requests never consumed

In `packages/happy-server/sources/app/api/routes/authRoutes.ts:71-84`:

```typescript
const answer = await db.terminalAuthRequest.upsert({
    where: { publicKey: publicKeyHex },
    update: {},
    create: { publicKey: publicKeyHex, supportsV2: request.body.supportsV2 ?? false }
});

if (answer.response && answer.responseAccountId) {
    const token = await auth.createToken(answer.responseAccountId!);
    return reply.send({ state: 'authorized', token, response: answer.response });
}
```

Once approved, the token is returned on every subsequent poll to `/v1/auth/request` with the same public key. The auth request is never consumed (deleted after first use) and has no expiration timestamp.

### Token cache
```typescript
import { LRUCache } from 'lru-cache';

private tokenCache = new LRUCache<string, TokenCacheEntry>({
    max: 10_000,
    ttl: 15 * 60 * 1000, // 15 minutes
});
```

For cross-replica revocation, broadcast invalidation events via the Redis pub/sub channel already available through `REDIS_URL` (used by the streams adapter in PR #1042).

### Auth request expiry
```typescript
if (answer.response && answer.responseAccountId) {
    const token = await auth.createToken(answer.responseAccountId!);
    // Consume after first delivery
    await db.terminalAuthRequest.delete({ where: { id: answer.id } });
    return reply.send({ state: 'authorized', token, response: answer.response });
}
```

Add a periodic cleanup job for stale requests (e.g., older than 10 minutes).

### LINKED ISSUE #1092: security(cli): Codex path accepts unvalidated permissionMode from remote messages
## Summary

The Codex message handler accepts `permissionMode` from remote mobile app messages without runtime validation, allowing escalation to full system access modes.

## Problem

In `packages/happy-cli/src/codex/runCodex.ts:161-163`:

```typescript
if (message.meta?.permissionMode) {
    messagePermissionMode = message.meta.permissionMode as import('@/api/types').PermissionMode;
    currentPermissionMode = messagePermissionMode;
```

The value from `message.meta.permissionMode` is cast with `as PermissionMode` (a TypeScript-only assertion with zero runtime effect) and directly assigned as the current permission mode. A remote message with `permissionMode: 'yolo'` results in `sandbox: 'danger-full-access'` and `approvalPolicy: 'on-failure'` via `resolveCodexExecutionPolicy()`.

**For comparison**, the Gemini path (`runGemini.ts:215-218`) correctly validates against an allowlist:

```typescript
const validModes: PermissionMode[] = ['default', 'read-only', 'safe-yolo', 'yolo'];
if (validModes.includes(message.meta.permissionMode as PermissionMode)) {
```

(Though the Gemini allowlist still includes `yolo`, at least it prevents injection of arbitrary strings.)

## Mitigating factors

- All RPC messages are end-to-end encrypted with TweetNaCl — exploiting this requires compromising the encryption key
- The system's trust model treats the mobile client as the legitimate user
- Codex's `executionPolicy.ts` has a `default` fallback for unrecognized modes

## Suggested fix

Add runtime validation matching the Gemini pattern, but exclude dangerous modes from remote messages:

```typescript
const VALID_REMOTE_MODES: readonly PermissionMode[] = ['default', 'acceptEdits', 'plan'] as const;

if (message.meta?.permissionMode) {
    if (VALID_REMOTE_MODES.includes(message.meta.permissionMode as PermissionMode)) {
        messagePermissionMode = message.meta.permissionMode as PermissionMode;
        currentPermissionMode = messagePermissionMode;
    }
    // Silently ignore yolo/bypassPermissions from remote — these should only be set at CLI startup
}
```

## Related

- #728 (prevents yolo override from mobile in Claude path — same class of issue, Codex path)
- #702 (YOLO mode message handling fixes)

### LINKED ISSUE #1098: bug(app): modeHacks does not capitalize 'build'/'plan' mode names — unit tests fail
## Bug Description

`hackMode()` in `sources/sync/modeHacks.ts` normalizes \"build, build\" → `'build'` and \"plan/plan\" → `'plan'` (lowercase), but the test suite expects the output to be capitalized: `'Build'` and `'Plan'`.

Additionally, when the key and name are both plain lowercase (`'build'`/`'plan'`), the function applies no transformation at all, so capitalization is also missing for that case.

## Failing Tests

`packages/happy-app/sources/sync/modeHacks.test.ts` — 3 of 4 tests fail:

```
✗ capitalizes plan/build when key and name are lowercase
  Expected: 'Build'   Received: 'build'

✗ normalizes build and plan duplicated labels
  Expected: 'Build'   Received: 'build'

✗ applies hacks across mode arrays
  Expected: [{ key: 'build', name: 'Build' }, ...]
  Received: [{ key: 'build', name: 'build' }, ...]
```

Related failure also propagates to `modelModeOptions.test.ts` (`applies hacks to metadata-provided operating modes`).

## Root Cause

```ts
// modeHacks.ts (current behaviour)
if (normalizedName === 'build, build') {
    return { ...mode, name: 'build' };   // ← should be 'Build'
}
if (normalizedName === 'plan/plan') {
    return { ...mode, name: 'plan' };    // ← should be 'Plan'
}
// no arm for simple 'build' / 'plan' — should capitalize those too
```

## Expected Fix

```ts
if (normalizedKey === 'build') return { ...mode, name: 'Build' };
if (normalizedKey === 'plan')  return { ...mode, name: 'Plan' };
```

This would cover both the duplicated-label variants and the simple lowercase case.

### LINKED ISSUE #1107: bug(cli): Daemon keeps restarting due to version mismatch (0.13.0 vs 0.13.1)
The happy daemon keeps restarting every minute due to a version mismatch, even though the installed CLI version is 0.13.0.

Observed behavior:
- Installed happy-coder CLI version is 0.13.0
- Daemon starts successfully
- However, the daemon detects a version mismatch and tries to restart
- It appears to expect version 0.13.1
- This results in an infinite restart loop (approximately every minute)

Log excerpt:
- "Current CLI version: 0.13.1, Daemon started with version: 0.13.0"
- "Daemon is outdated, triggering self-restart with latest version"

State file (~/.happy/daemon.state.json):
- "startedWithCliVersion": "0.13.0"

Environment:
- OS: Linux
- Node.js: v24.14.0
- happy-coder: 0.13.0

Additional notes:
- Daemon successfully connects to the server and starts normally
- The issue occurs during periodic version checks
- It seems that the CLI version is incorrectly detected as 0.13.1, or there is a mismatch between local and expected versions

## log
```version
motoya@vm-d2daf1e5-17:~/obsidian_vault$ happy --version
happy version: 0.13.0
Using Claude Code v2.1.110 from npm
2.1.110 (Claude Code)
```

```.happy/logs/2026-04-16-18-34-06-pid-3741911-daemon.log
[18:34:06.668] [ClaudeSettings] Successfully read Claude settings from /home/motoya/.claude/settings.json 
[18:34:06.669] [ClaudeSettings] includeCoAuthoredBy: undefined 
[18:34:06.669] Starting happy CLI with args:  ["/home/motoya/.nvm/versions/node/v20.20.2/bin/node","/home/motoya/.nvm/versions/node/v20.20.2/lib/node_modules/happy-coder/dist/index.mjs","daemon","start-sync"]
[18:34:06.670] [DAEMON RUN] Starting daemon process... 
[18:34:06.671] In production, skipping message inspection 
[18:34:06.671] [DAEMON RUN] Environment 
 {
  "PWD": "/home/motoya/obsidian_vault",
  "workingDirectory": "/home/motoya/obsidian_vault",
  "processArgv": [
    "/home/motoya/.nvm/versions/node/v20.20.2/bin/node",
    "/home/motoya/.nvm/versions/node/v20.20.2/lib/node_modules/happy-coder/dist/index.mjs",
    "daemon",
    "start-sync"
  ],
  "happyDir": "/home/motoya/.happy",
  "serverUrl": "https://api.cluster-fluster.com",
  "logsDir": "/home/motoya/.happy/logs",
  "processPid": 3741911,
  "nodeVersion": "v20.20.2",
  "platform": "linux",
  "arch": "x64",
  "user": "motoya",
  "home": "/home/motoya",
  "shell": "/bin/bash",
  "terminal": "xterm-256color"
}
[18:34:06.672] [DAEMON CONTROL] Checking if daemon is running same version 
[18:34:06.684] [DAEMON CONTROL] Current CLI version: 0.13.1, Daemon started with version: 0.13.0 
[18:34:06.684] [DAEMON RUN] Daemon version mismatch detected, restarting daemon with current CLI version 
[18:34:06.685] Stopping daemon with PID 3741661 
[18:34:07.111] Daemon stopped gracefully via HTTP 
[18:34:07.113] [caffeinate] Not on macOS, skipping caffeinate 
[18:34:07.113] [AUTH] Starting auth and machine setup... 
[18:34:07.115] [AUTH] Using existing credentials 
[18:34:07.116] [AUTH] Machine ID: de7186cc-3cb4-4ede-81c4-bee6b9e97385 
[18:34:07.117] [DAEMON RUN] Auth and machine setup complete 
[18:34:07.146] [CONTROL SERVER] Started on port 41245 
[18:34:07.147] [DAEMON RUN] Daemon state written 
[18:34:07.670] [API] Machine de7186cc-3cb4-4ede-81c4-bee6b9e97385 registered/updated with server 
[18:34:07.671] [DAEMON RUN] Machine registered: de7186cc-3cb4-4ede-81c4-bee6b9e97385 
[18:34:07.672] [API MACHINE] Connecting to wss://api.cluster-fluster.com 
[18:34:07.678] [DAEMON RUN] Daemon started successfully, waiting for shutdown request 
[18:34:08.452] [API MACHINE] Connected to server 
[18:34:08.455] [API MACHINE] Keep-alive started (20s interval) 
[18:34:08.631] [API MACHINE] Daemon state updated successfully 
[18:34:08.632] [API MACHINE] Received external daemon state update 
```

### LINKED ISSUE #1110: Web worktree spawn fails with 'Not a Git repository' — request never reaches daemon
### Environment
- Happy CLI v1.1.6
- Linux x64, Node v25.8.2
- Web client (app.happy.engineering)

### Reproduction Steps

1. Open Happy Coder web client
2. Select a machine that is online and has a git repo (e.g. `/home/yhxiang/repos/vnext`)
3. Choose to create a new session with **"new worktree"** mode
4. **Error**: "Not a Git repository"

### Evidence

- The daemon log shows **no incoming RPC request** when this error occurs — the `spawn-happy-session` handler is never called
- All successful (non-worktree) spawns show `spawn-in-directory` type requests in daemon log with the correct git repo directory
- The `git worktree list --porcelain` bash RPC calls all succeed with `cwd: /home/yhxiang/repos/vnext`
- The daemon spawn code only generates args `claude --happy-starting-mode remote --started-by daemon` — no `-w`/`--worktree` flag is ever passed

### Analysis

The error originates from the **server/web side**, not the CLI daemon. The worktree creation request never reaches the daemon. The server appears to attempt worktree creation using incorrect repository information, resulting in the "Not a Git repository" error before any RPC call is made to the daemon.

### Expected Behavior
- Worktree session creation should work when the target directory is a valid git repository
- The server should pass the worktree creation request to the daemon, which can then execute `git worktree add` in the correct directory

### Impact
Worktree mode is completely non-functional for remote session creation from the web client.

Scoteezy:
## Summary

A batch of UX improvements, session management enhancements, and codebase cleanup across `happy-app`, `happy-cli`, and `happy-server`.

### UI (happy-app)

- **File diffs sidebar** — new `FilesSidebar` component showing changed files per session with collapsible animation. Clicking a file opens it inline in a centered diff pane.
- **File tree redesign** — sidebar renders changed files as a nested tree with path compression (`packages/frontend/src` collapses into one node when each intermediate dir has a single dir child), header count doubles as expand-all/collapse-all, text filter auto-expands matching branches, rotating chevron animation per folder.
- **Empty sidebar state** — large check-mark badge, centered vertically inside the scroll container, readable title/subtitle with proper contrast.
- **Pierre-based diff renderer (web)** — replaces the custom DiffView with [@pierre/diffs](https://www.npmjs.com/package/@pierre/diffs) for Shiki-backed syntax highlighting. Bundle loads once per app, prefetched on session mount so the first chat diff doesn't wait on lazy chunks. Native keeps a plain-text colorized renderer (no WASM in Hermes).
- **Diff style setting** — new Settings → Appearance entry lets users pick unified or split view. Applies only to the dedicated `InlineFileDiff` pane; chat tool rows are locked to inline unified.
- **Inline diff pane** — dedicated `InlineFileDiff` centered to match chat width (`layout.maxWidth`), single sticky header with file icon + path + `+N/-N` stats + Unified/Split toggle (web) + close button. Body scrolls, header stays put. Tracked files use `git diff HEAD -- path` so staged + unstaged render together; untracked files render as pure additions via `cat`.
- **Chat tool header stats** — `CodexDiffView` / `CodexPatchView` show `+N/-N` in their own file header; Pierre's built-in header is hidden inside chat diffs so there's only one row per file.
- **Tool section gap fix** — `fullWidthSection` compensates for the parent tool-view's `paddingTop`, so chat diffs sit flush under the tool header instead of leaving an empty strip.
- **Table row-first layout** — tables now use content-estimated column widths (clamped to min/max), row-based flex layout with proper cell alignment, wrapped in horizontal ScrollView.
- **Shared HorizontalScrollView** — extracted reusable `HorizontalScrollView` with smart wheel behavior (touchpad horizontal swipe consumed, Shift+wheel to horizontal scroll, plain vertical wheel passes through to page).
- **Scrollbar visibility fix** — scrollbar thumb used `var(--colors-divider)` which equals `surfaceHighest` in dark theme (`#292929`), making it invisible on code blocks. Switched to `rgba(150,150,150,0.4)` for consistent visibility across all surfaces.
- **Markdown overflow fix** — code blocks and tables properly constrained with horizontal scroll instead of overflowing beyond screen width.
- **Small tables stay compact** — tables use `maxWidth: 100%` + `alignSelf: flex-start` so narrow tables retain natural width instead of stretching.
- **UI updates** — improvements to AgentInput, ChatList, ActiveSessionsGroupCompact, theme, and unistyles.
- **Removed non-compact session list** — deleted `ActiveSessionsGroup` (500+ lines), compact view is now the only option.
- **Disable analytics toggle** — new toggle in Features settings to fully kill PostHog reporting.
- **Removed stale files** — old changelog, `3dparty.md`, `.codex/config.toml`, `.mcp.json`.
- **Model / effort persistence (closes #1028)** — `modelMode` and `effortLevel` now survive app restart.
- **Mode name capitalization fix (closes #1098)** — `hackMode` now capitalizes standalone lowercase `build`/`plan` labels.
- **Worktree error masking fix (closes #1110)** — `createWorktree` no longer masks RPC failures as "Not a Git repository".

### CLI (happy-cli)

- **MCP/skills commands passthrough** — `/mcp` and `/skills` commands pass through to Claude Code natively in local mode instead of being intercepted. Skills metadata forwarded as-is from Claude Code init.
- **Session resume via RPC** — resume sessions in-place through the daemon control server, passing `--mode` and `--model` from the original session.
- **Encryption data persistence** — daemon now stores session encryption keys so sessions can be resumed even without the key file on disk.
- **Atomic writes** — `sessions.json` is now written via temp file + rename to prevent corruption.
- **Path validation fix** — uses `path.sep` instead of hardcoded `/` for proper Windows support.
- **Windows spawn ENOENT fix (closes #1082)** — migrated remaining `child_process.spawn` call sites to `cross-spawn`.
- **Daemon restart loop fix (closes #1107)** — version-mismatch detection no longer reads `package.json` from disk.
- **Codex auto-approve bypass fix (closes #1085)** — switched from `.includes()` substring matching to exact `Set.has()` matching.
- **Codex permissionMode validation (closes #1092)** — now validated against the Codex-native mode allowlist.
- **Plan mode permission fix** — auto-approves non-dangerous tools in plan mode.
- **Gemini safe-yolo security fix (closes #1084)** — safe-yolo now shares the same dangerous-tool filter as read-only mode.

### Server (happy-server)

- Cleaned up machine routes and socket handling related to session resume flow.
- **Token cache TTL + bounded size (closes #1087)** — 24h TTL, 10k cap with LRU eviction, periodic cleanup.

## Notes

- The resume flow currently uses the existing bash RPC — a dedicated spawn RPC is planned.
- All 10 locale files updated with new translation keys for the sidebar, analytics toggle, and diff style setting.
- Removed `fix-livekit-room-reuse.cjs` patch and its postinstall hook.
- Windows daemon fixes verified manually on Windows 11.
- Codex permission fixes verified manually across all 4 modes.
- modeHacks unit tests pass (4/4) after capitalize fix.


bra1nDump:
> **Automated review** — posted by Claude via [Happy](https://happy.engineering). This is advisory; findings need human verification before acting.

## Summary
Batched PR: **85 files, +3246/-1233, 37 commits**, spanning `happy-app` (Pierre diff rewrite + file tree sidebar + inline diff pane), `happy-cli` (session resume via RPC + encryption persistence + daemon restart-loop fix + Windows spawn + auto-approve bypass fix), `happy-server` (token cache TTL/LRU). Spot-checks of all 9 linked issues match the claimed fixes.

## Blockers

- **Shell command injection — `packages/happy-app/sources/components/InlineFileDiff.tsx:57-73`.** ``cat -- "${gitDiffPath}"`` and ``git diff HEAD --no-ext-diff -- "${gitDiffPath}"`` double-quote-interpolate a path into a shell command. A file named ``foo";rm -rf ~/"`` in a user's working tree executes arbitrary shell when they click it in the sidebar. ``--`` blocks flag injection but not quote-escape. Fix: reject paths containing ``"`` / ``$`` / `` ` `` / ``\`` / newlines, or pass via stdin / use an RPC that takes a structured filename arg.
- **`packages/happy-app/sources/sync/settings.spec.ts` fails on this branch.** The main merge pulled in `voiceBypassToken` / `voiceCustomAgentId` schema fields but the expected-defaults object in the test wasn't updated.
- **Codex resume silently drops `--model` / `--permission-mode`.** `daemon/run.ts:657-662` appends them to `launch.args` for both Claude and Codex. Claude's `index.ts` forwards them into `claudeArgs`; Codex's `commands/codexCommand.ts` + `codex/cliArgs.ts` only parse `--resume`, so for Codex resume the passthrough (a headline feature) is a no-op.
- **Concurrent webhook writes clobber `sessions.json`.** `packages/happy-cli/src/persistence.ts:434-444` does read → mutate → write-tmp → rename with no lock. Two near-simultaneous session starts will race: the second read sees a stale snapshot and overwrites the first addition. Reuse the `.lock` pattern from `updateSettings` in the same file.
- **`sessions.json` writes encryption keys with default umask (~0644).** `persistence.ts:434-444` passes no `mode` to `writeFileSync`. On multi-user hosts, any local user can read session encryption keys. Fix: ``{ mode: 0o600 }`` on write + ``mkdir { mode: 0o700 }`` + defensive ``chmodSync`` (Node keeps existing perms on overwrite).

## Needs-fix

- **Plan mode auto-approves `WebFetch` / `WebSearch` / `Task` / `mcp__*`.** `claude/utils/getToolDescriptor.ts:1-12` only marks `Edit|MultiEdit|Write|NotebookEdit|Bash` as dangerous; `claude/utils/permissionHandler.ts:182-184` auto-approves anything else in plan mode. Invert the default — allowlist the read-only tools instead.
- **Resume RPC's `model` / `permissionMode` args not allow-listed.** `apiMachine.ts:172-206` validates only that `sessionId` is a string. Not argv-level RCE (spawn uses argv array), but Claude CLI's argparse could be nudged in unexpected ways.
- **Concurrent resume can spawn duplicate live processes.** `daemon/run.ts:625-687` doesn't check whether a tracked session's PID is still alive before spawning. Two resumes, or a resume during a running session, produces two processes driving the same ``sessionId``.
- **`dist/index.mjs` mtime-touch can leave no daemon running.** `daemon/run.ts:764-771,838-875`: any ``touch`` flips mtime → heartbeat decides "bundle replaced" → spawns new daemon → old daemon hangs 10s then exits → new daemon sees matching baked version and exits immediately at `run.ts:125-128`. Add content-hash or mtime-AND-size check.

## Minor

- **Token cache (#1087) docs mismatch code.** `happy-server/sources/app/auth/auth.ts:118-125`: PR description says "O(1) LRU"; actual impl is ``[...entries].sort().slice()`` on eviction — O(N log N), and not LRU (timestamp only set on insert, not refreshed on hit). Functionally fine, but the claim is wrong.
- **14-day session pruning only filters on read.** `persistence.ts:413,424` — stale keys stay on disk until the next `persistSession` call happens to exclude them. If a user runs Happy once and stops, their key lives forever. Rewrite file eagerly from daemon startup after filter.
- **`HorizontalScrollView.tsx:40` missing useEffect deps** — re-subscribes on every render.
- **`getPatchDiffStats` (`calculateDiff.ts:302-317`) undercounts rare `---` deletion lines** — `startsWith('---')` inside a hunk matches both file-header and a deleted `---` literal. Low severity.
- **`cat` on untracked files has no size guard (`InlineFileDiff.tsx:57-75`).** Large binary → multi-MB send over socket → Pierre panic. Add a size precheck.
- **Hardcoded user-visible strings (i18n violations):** `features.tsx:27-38,82-93`, `InlineFileDiff.tsx:212,215` ("Unified" / "Split"), `SessionView.tsx:664`. All 10 locales are complete for keys that *were* added, but these new strings skipped i18n.
- **`patches/expose-pierre-diffs-style.cjs`** — hand-written ``postinstall`` mutation of ``node_modules/@pierre/diffs/package.json``. Brittle. Use ``patch-package`` (repo already has it) or submit upstream.
- **`happy-cli` CLAUDE.md says "No mocking"** — new `ensureDaemonRunning.test.ts` and `permissionHandler.test.ts` use ``vi.mock`` / ``vi.fn``. Either the convention is stale (looks systemic) or the tests need rewriting.

## OK / positive

- Codex/Gemini auto-approve bypass fix (#1085) — `Set.has` exact-match is correct; regression tests explicitly reproduce the bypass vectors.
- Windows spawn (#1082) — all bare `spawn('node', ...)` sites migrated to `cross-spawn` (grep-confirmed).
- Path validation — `path.sep` fix correct; tests cover traversal + prefix-sibling.
- Daemon restart loop (#1107) — baked-in `currentCliVersion` fixes the npm stub-publish case.
- Typecheck — 0 new real errors in either `happy-cli` or `happy-app` (the 6 happy-app deltas are `@pierre/diffs` module-resolution pre-install).
- New permission-handler tests pass 11/11.
- i18n — every newly added key is present in all 10 locale files.

## Process

This is too large to review as one unit. Natural split: (a) Pierre diff + file tree + inline pane, (b) session resume via RPC + encryption persistence, (c) bundled bugfixes, (d) analytics toggle + misc cleanup. The resume-via-RPC flow and daemon persistence are the biggest risk surfaces and ship with **zero new tests** — highest-leverage place to slow down.

``@pierre/diffs@1.1.16`` is a young (Dec 2025) package from reputable maintainers (mdo, fat, Sexton) but no community track record. A bespoke dep with a runtime patch is fine as long as the team is OK owning that upstream coordination.

PR #991, opened 2026-04-05
feat: add image paste and display support for web UI

chaehyun2:
## Summary

- Add image paste/attach support in the web UI chat input, allowing users to send images alongside text messages
- Display images inline within chat message bubbles in the web UI
- Implement image message TTL (3 days) with automatic server-side cleanup
- Fix Dockerfile build args, Web Crypto API encryption module, and schema types for full image pipeline support

## Changes

- **Web UI**: Image paste handler in chat input, inline image rendering in message bubbles
- **CLI pipeline**: `buildContent()` helper converts text+images into Claude SDK content blocks (base64 image sources)
- **Server integration**: Image upload/download endpoints with TTL-based auto-deletion (3 days)
- **Encryption**: Restore `aes.web.ts` for Web Crypto API compatibility in the web app
- **Docker**: Pass `HAPPY_SERVER_URL` build arg to `Dockerfile.webapp`
- **Schema**: Add `images` field to `RawRecord` Zod schema for type-safe image message handling
- **Reducer**: Fix message ID matching so images display on the correct chat bubble

## Test plan

- [ ] Paste an image into the web UI chat input and verify it appears as a preview
- [ ] Send a message with an image and confirm it displays inline in the chat
- [ ] Verify text-only messages still work correctly (no regressions)
- [ ] Confirm images are uploaded to the server and accessible via the returned URL
- [ ] Wait for TTL expiry and verify server-side cleanup removes the image
- [ ] Test the web app build with the Dockerfile to ensure `HAPPY_SERVER_URL` is passed correctly

🤖 Generated with [Claude Code](https://claude.ai/code)
via [Happy](https://happy.engineering)

PR #1101, opened 2026-04-16
fix: avoid stack overflow in encodeBase64 for large buffers

chaehyun2:
## Summary

`encodeBase64` in `packages/happy-app/sources/encryption/base64.ts` (the web path — the native path via `base64.native.ts` uses `react-native-quick-base64` and is unaffected) calls `String.fromCharCode.apply(null, Array.from(buffer))`, which hits V8's argument limit once the buffer exceeds roughly half a megabyte.

### Reproducer (Node 20 / V8)

```js
const buf = new Uint8Array(500_000);
String.fromCharCode.apply(null, Array.from(buf));
// RangeError: Maximum call stack size exceeded
```

A typical JPEG screenshot is 1–3 MB, so any code path that base64-encodes an image on web currently throws. The fix processes the buffer in 8 KB chunks and joins the pieces.

## Changes

- `encodeBase64` now iterates the buffer in 8 KB windows and joins the resulting strings — output is byte-identical to the previous implementation for inputs below the old limit.
- New `base64.test.ts` with regression tests:
  - 1 MB and 5 MB round-trip
  - Boundary sizes 8191 / 8192 / 8193 / 16383 / 16384 / 16385 to cover the internal chunk edge

I verified the tests fail on `main` with `RangeError` on the 1 MB and 5 MB cases and pass on this branch (the boundary cases pass on both — they're just sanity).

## Test plan

- [x] `pnpm --filter happy-app exec vitest run sources/encryption/base64.test.ts` — 3/3 passing
- [x] Reverted `base64.ts` to `origin/main`; confirmed the 1 MB and 5 MB tests fail with `RangeError: Maximum call stack size exceeded`
- [x] 10 MB round-trip (ad-hoc) completes in ~110 ms with correct byte-for-byte result

### DIFF
--- a/packages/happy-app/sources/encryption/base64.test.ts
+++ b/packages/happy-app/sources/encryption/base64.test.ts
@@ -1,0 +1,38 @@
+import { describe, it, expect } from 'vitest';
+import { encodeBase64, decodeBase64 } from './base64';
+
+describe('encodeBase64 large buffers', () => {
+    it('encodes a 1MB buffer without stack overflow', () => {
+        const size = 1024 * 1024;
+        const input = new Uint8Array(size);
+        for (let i = 0; i < size; i++) input[i] = (i * 37) % 256;
+
+        const encoded = encodeBase64(input);
+        const decoded = decodeBase64(encoded);
+
+        expect(decoded.length).toBe(size);
+        expect(decoded[0]).toBe(input[0]);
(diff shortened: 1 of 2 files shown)


bra1nDump:
closing - this is already on main: @Scoteezy's 26f465bdc chunks encodeBase64 the same way and his commit notes it restored the chunking lost from your #991. thanks @chaehyun2!

PR #1108, opened 2026-04-16
perf: lazy-load chat messages on scroll-to-top

chaehyun2:
## Summary
- Initial session load fetches only the latest 50 messages instead of the entire history
- Server adds a `before_seq` query parameter to fetch older messages
- ChatList triggers `loadOlderMessages` via `onEndReached` (inverted FlatList) when the user scrolls to the oldest visible message
- Loading spinner is shown while fetching the next batch
- The `latest=true` parameter returns messages in descending order, then they're reversed to ASC for the client

## Why
For long sessions (thousands of messages), the existing flow fetched the entire history on every page refresh, causing multi-second loading delays. With lazy loading, the first paint is fast and older messages load only when the user actually scrolls back.

## Changes
- **Server** (`v3SessionRoutes.ts`): Add `before_seq` and `latest` query params; reuse the `latest=true` reverse-order path when `before_seq` is provided
- **App** (`sync.ts`): Track `sessionOldestSeq` and `sessionHasMoreOlder` per session; add `loadOlderMessages(sessionId)` that fetches a 50-message page older than the current oldest
- **App** (`ChatList.tsx`): Wire `onEndReached` to `sync.loadOlderMessages`; render an `ActivityIndicator` while loading

## Test plan
- [ ] Open a session with > 50 messages — only the latest 50 render initially and the chat is responsive immediately
- [ ] Scroll up to the oldest visible message — older messages load with a spinner, position is preserved
- [ ] Scroll all the way back — loading stops cleanly when there are no more messages
- [ ] Open a short session (< 50 messages) — `loadOlderMessages` no-ops without an extra request

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
         "overrides": {
             "whatwg-url": "14.2.0",
             "parse-path": "7.0.3",
-            "@types/parse-path": "7.0.3"
+            "@types/parse-path": "7.0.3",
+            "cli-cursor@4>restore-cursor": "^5.1.0"
         },
         "onlyBuiltDependencies": [
             "@more-tech/react-native-libsodium",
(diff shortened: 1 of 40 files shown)


bra1nDump:
closing - lazy-load history + backward pagination shipped on main via #1242 (merged), which covers the initial-50 + scroll-to-load-older approach here. thanks @chaehyun2!

PR #1075, opened 2026-04-15
fix(server): increase cross-replica fetchSockets timeout to prevent false "RPC method not available"

ratio-dd:
## Summary

- Split `fetchSockets()` timeout into two: **2s for daemon lookup** (initial + grace window) and **500ms for in-flight presence polling** (unchanged)
- Increase reconnect grace window from **10s → 15s** to maintain ~6-7 poll iterations per window
- Fix misleading comment about cluster-adapter default timeout (10s, not 5s)

## Root cause

PR #1042 rewrote RPC routing to use Socket.IO rooms + cross-replica `fetchSockets()`. PR #1046 scaled to 3 replicas. The combination caused intermittent "RPC method not available" errors when creating sessions from web/iOS.

The 500ms `fetchSockets()` timeout was too aggressive for cross-replica queries via the Redis streams adapter. The full round-trip (XADD → peer XREAD → process → XADD response → local XREAD) can exceed 500ms under normal load. When it does, `fetchRoomSockets` catches the timeout and returns `[]`. The 10s grace window polled ~14 times with the same 500ms timeout — all failing — resulting in "RPC method not available" even though the daemon was connected on another replica.

## What changed

**File**: `packages/happy-server/sources/app/api/socket/rpcHandler.ts` (1 file, no client changes)

| Constant | Before | After | Why |
|----------|--------|-------|-----|
| `RPC_LOOKUP_FETCH_TIMEOUT_MS` | (didn't exist, used 500ms) | `2_000` | Reliable cross-replica lookup |
| `RPC_PRESENCE_FETCH_TIMEOUT_MS` | `500` | `500` (unchanged) | Fast daemon-death detection |
| `RPC_RECONNECT_GRACE_MS` | `10_000` | `15_000` | Compensate for slower per-poll (~2.2s vs ~0.7s) |

`fetchRoomSockets()` now takes a `timeoutMs` parameter so callers specify the appropriate timeout:
- Initial lookup + grace window polling → `RPC_LOOKUP_FETCH_TIMEOUT_MS` (2s)
- In-flight presence poll → `RPC_PRESENCE_FETCH_TIMEOUT_MS` (500ms)

## What this does NOT change

- In-flight presence poll: stays at 500ms for fast dead-daemon detection (~1s)
- Client-side code: zero changes to daemon or web app
- RPC protocol: no wire format changes, no new events
- Single-replica behavior: local `fetchSockets` resolves instantly, fix is transparent

## Test plan

[...]

PR #1133, opened 2026-04-19
fix(server): add cluster-adapter round-trip barrier to rpc-register

### LINKED ISSUE #1118: fix: RPC registration fire-and-forget causes intermittent 'method not available'
### Environment
- Happy CLI v1.1.6
- Linux x64, Node v25.8.2
- Server with Redis streams adapter (`@socket.io/redis-streams-adapter@0.2.3`)
- Multi-replica deployment

### Reproduction Steps

1. Start daemon → connects to server, registers 10 RPC methods
2. Server confirms all 10 `rpc-registered` acks
3. From web/app client, `bash` RPC calls succeed (directory listing, worktree listing)
4. From web/app client, `spawn-happy-session` RPC returns "RPC method not available"
5. Retrying multiple times eventually succeeds

### Evidence

**Daemon-side (all confirmed working):**
- 10/10 `rpc-register` emitted, 10/10 `rpc-registered` acks received
- `bash` RPC requests arrive and are handled (336+ `[RPC]` log entries)
- `spawn-happy-session` RPC requests **never arrive** (0 log entries)
- Keep-alive running normally, no disconnect/error in daemon log

**Server-side routing:**
- Room format `rpc:${userId}:${machineId}:${method}` — identical for all methods
- Same daemon socket in all rooms — `fetchSockets` should find it consistently

### Root Cause

The daemon's `RpcHandlerManager.onSocketConnect()` emits `rpc-register` as **fire-and-forget** — it never waits for the `rpc-registered` ack:

```ts
// packages/happy-cli/src/api/rpc/RpcHandlerManager.ts:98-103
onSocketConnect(socket: Socket): void {
    this.socket = socket;
    for (const [prefixedMethod] of this.handlers) {
        socket.emit('rpc-register', { method: prefixedMethod }); // fire-and-forget
    }
}
```

The existing stress test (`deploy/integration-tests/stress-rpc-registration.mjs`) already documents this exact problem (lines 5-7):

> the daemon emits rpc-register once on connect without waiting for ack, exactly like RpcHandlerManager.onSocketConnect() does. The server's async verifyToken() may not have set up the rpc-register listener yet, causing silent event drops and "RPC method not available" for callers.

While the server-side auth middleware fix (commit `c744ae20`) moved `verifyToken` to middleware to reduce the race window, the fire-and-forget registration pattern remains fragile in these scenarios:

1. **Cross-replica room sync delay**: After daemon reconnects, `fetchSockets` across replicas may not reflect the newly-joined rooms immediately. The `fetchRoomSockets` timeout is only 500ms, and the `waitForRoomMember` grace window (10s) may not be enough during replica restarts.

2. **`bash` appears to always work** because it's called frequently (directory browsing, worktree listing) — occasional failures are masked by the next successful call. `spawn-happy-session` is a one-shot call, so any transient registration gap is visible to the user.

3. **Reconnection re-registration**: Daemon uses custom reconnect logic (`startSmartReconnect` with `reconnection: false`). Each reconnect fires another round of fire-and-forget `rpc-register` emits. If the server's `connection` handler hasn't fully attached the `rpc-register` listener when the emits arrive, they are silently dropped.

### Proposed Fix

Change `onSocketConnect` to await `rpc-registered` acks for each method:

```ts
async onSocketConnect(socket: Socket): Promise<void> {
    this.socket = socket;
    for (const [prefixedMethod] of this.handlers) {
        await new Promise<void>((resolve) => {
            socket.timeout(5000).emit('rpc-register', { method: prefixedMethod }, (ack) => {
                if (ack?.method) {
                    resolve();
                } else {
                    this.logger('[RPC] Registration failed or timed out', { method: prefixedMethod });
                    resolve(); // Don't block — best effort
                }
            });
        });
    }
}
```

This ensures room membership is committed on the server before the method is considered available, eliminating the registration race across all scenarios.

### Related
- #1110 (worktree creation masks RPC failure as "Not a Git repository")
- #1111 (worktree session message sync — unrelated root cause but same symptom of intermittent unavailability)
- Stress test: `deploy/integration-tests/stress-rpc-registration.mjs` scenario `fire-and-forget`

ratio-dd:
## Summary

- Add a cluster-adapter round-trip barrier inside the `rpc-register` handler: after `socket.join(room)`, await `io.in(room).fetchSockets()` before emitting `rpc-registered`
- 1-second timeout on the barrier; on timeout the handler still acks (best-effort, no worse than today)
- 1 file changed, +24 / -2 — `packages/happy-server/sources/app/api/socket/rpcHandler.ts`

## Root cause

After PR #1042 moved RPC routing to Socket.IO rooms and PR #1046 scaled to 3 replicas, the `rpc-register` handler acked immediately after a local `socket.join()`:

```ts
socket.on('rpc-register', (data) => {
    socket.join(rpcRoom(userId, method));
    socket.emit('rpc-registered', { method });  // acked before propagation
});
```

`socket.join()` mutates local adapter state only. When a caller lands on a peer replica and fires `rpc-call` immediately after receiving `rpc-registered`, the caller's `fetchSockets()` has to round-trip through the Redis streams adapter. Under streams backlog (noise from concurrent daemon reconnects, churn during rolling deploys), the peer replica's XREAD cursor may not yet have processed the registration-adjacent events, producing two observed symptoms — both documented in #1118:

1. **`RPC method not available`** — `fetchSockets()` returns `[]` before the daemon becomes visible cross-replica
2. **`operation has timed out`** — `fetchSockets()` finds the daemon but the cross-replica `emitWithAck` response comes back after the 6s caller timeout

#1075 (cross-replica fetch timeout 500ms→2s) addressed some of the lookup-side lag but left the ack-path cursor lag unmitigated. This PR closes that last gap for the common registration path.

## Fix

Force one cluster-adapter round-trip before acking:

```ts
socket.on('rpc-register', async (data) => {
    const room = rpcRoom(userId, method);
    socket.join(room);
    await fetchRoomSockets(io, room, RPC_REGISTER_BARRIER_TIMEOUT_MS);
    socket.emit('rpc-registered', { method });
});
```

`ClusterAdapter.fetchSockets()` publishes a `FETCH_SOCKETS` message on the stream and waits for every known peer to respond. Because streams consumers advance their XREAD cursor strictly in-order, any peer that responds to this request has necessarily read every earlier entry on the stream. When the barrier resolves, each peer's cursor is caught up past the point of `socket.join()`, so the caller's subsequent `fetchSockets()` from any replica is guaranteed to see the daemon.

Single-replica deployments short-circuit inside the adapter (`expectedResponseCount === 0`, `socket.io-adapter/cluster-adapter.js:374-376`) — zero additional latency. Under load the barrier is adaptive to actual streams RTT instead of a wall-clock heuristic. On barrier timeout `fetchRoomSockets` returns `[]` + logs and the handler still acks (best-effort degradation — no regression vs. current behavior).

## What changed

**File**: `packages/happy-server/sources/app/api/socket/rpcHandler.ts` (1 file, no client changes)

| Constant | Value | Why |
|----------|-------|-----|
| `RPC_REGISTER_BARRIER_TIMEOUT_MS` | `1_000` (new) | Short enough that a dead peer doesn't stall every registration for the adapter's 10s `heartbeatTimeout`; long enough to absorb typical streams backlog |

Handler signature changes from sync to `async`. Semantics of the ack are unchanged — it still emits `rpc-registered { method }` on success and `rpc-error` on validation failure. The `try/catch` continues to cover the whole body; `fetchRoomSockets` already swallows its own timeout (returns `[]` + logs), so the catch arm is unreachable from the new code path.

## What this does NOT change

- **Client (daemon) code**: zero changes. Daemon's registration loop behaves identically — it still fires `rpc-register` per method and listens for `rpc-registered`
- **Wire protocol**: no new events, no payload changes
- **`rpc-call` path**: unchanged (`waitForRoomMember` + 15s grace window remains as the second line of defense)
- **Single-replica / standalone**: barrier is a no-op via adapter short-circuit
- **`rpc-unregister`**: untouched; symmetrical concern is de minimis in practice (stale room membership resolves on next reconnect)

## Alternative considered

A pure daemon-side fix (await `rpc-registered` ack on the client) — proposed in #1119 — does not address the scenario here: even if the daemon awaits the ack, the current server code still emits that ack before peers have consumed the join-adjacent events, so the caller on another replica can still race. The two approaches are complementary; this PR fixes the server side only.

## Test plan

- [x] Typecheck: `pnpm --filter happy-server build`
- [x] Local standalone (`pnpm standalone:dev`): single-process, no Redis — no regression
- [x] Multi-replica validation: minikube 3-replica cluster with Redis streams adapter — results below

## Test results

Validated with a local minikube cluster: 3 `handy-server` replicas + Redis streams adapter, using the existing `deploy/integration-tests/` harness pattern from PR #1075.

Harness: 10 noise daemons churning connect → register-10-methods → disconnect every 200–600ms across 3 pods, plus 1 victim daemon per round doing `register 10 methods on POD1` → `caller on POD2 fires 10 concurrent rpc-calls`. Two independent runs of 120 rounds × 10 methods each:

```
=== Load test: 10 noise daemons + victim over 120 rounds ===

Per-method victim results:
idx method                     ok   fail  errors
 0  spawn-happy-session     120     0
 1  stop-session            120     0
 2  stop-daemon             120     0
 3  bash                    120     0
 4  readFile                120     0
 5  writeFile               120     0
 6  listDirectory           120     0
 7  getDirectoryTree        120     0
 8  ripgrep                 120     0
 9  difftastic              120     0

Total failures: 0/1200 (0.0%)  ── run 1
Total failures: 0/1200 (0.0%)  ── run 2
```

Comparable main-without-this-fix baseline runs (same harness, pre-fix image) previously hit `1/600` on METHODS[0] with `operation has timed out`.

Fixes #1118

### DIFF
--- a/packages/happy-server/sources/app/api/socket/rpcHandler.ts
+++ b/packages/happy-server/sources/app/api/socket/rpcHandler.ts
@@ -27,6 +27,10 @@ const RPC_LOOKUP_FETCH_TIMEOUT_MS = 2_000;
 // so a dead replica doesn't stall each poll for the full adapter heartbeatTimeout
 // (10s). 500ms keeps daemon-death detection responsive (~1s).
 const RPC_PRESENCE_FETCH_TIMEOUT_MS = 500;
+// Timeout for the post-register barrier fetchSockets. Short enough that a dead
+// peer replica doesn't stall every registration for the full heartbeatTimeout
+// (10s), long enough to absorb typical streams backlog under load.
+const RPC_REGISTER_BARRIER_TIMEOUT_MS = 1_000;
 // How long an rpc-call waits for the daemon socket to appear in the room when
 // the room is empty at call time (e.g. brief daemon reconnect window). With
 // RPC_LOOKUP_FETCH_TIMEOUT_MS at 2s + RPC_RECONNECT_POLL_MS at 200ms, each
@@ -127,14 +131,32 @@ async function waitForRoomMember(io: Server, room: string, maxMs: number, metric
 
 export function rpcHandler(userId: string, socket: Socket, io: Server) {
 
-    socket.on('rpc-register', (data: any) => {
+    socket.on('rpc-register', async (data: any) => {
         try {
             const { method } = data ?? {};
             if (!method || typeof method !== 'string') {
                 socket.emit('rpc-error', { type: 'register', error: 'Invalid method name' });
                 return;
             }
-            socket.join(rpcRoom(userId, method));
+            const room = rpcRoom(userId, method);
+            socket.join(room);
+            // Force a cluster-adapter round-trip before acking, so peer
+            // replicas drain their Redis-streams XREAD cursor past this point.
+            // Without it, a caller that fires rpc-call on another replica
+            // immediately after rpc-registered can hit either an empty
+            // fetchSockets (peer backlog) or a stalled cross-replica
+            // emitWithAck (ack-path cursor lag). fetchSockets publishes a
+            // FETCH_SOCKETS message and waits for every peer to respond —
+            // when it resolves, every known peer has XREAD at least up to
+            // our request, which implicitly drains any earlier backlog.
+            //
+            // Single-replica deployments short-circuit inside the adapter
+            // (expectedResponseCount === 0), so this is effectively free
+            // there. Under load, it adapts to actual streams RTT instead of
+            // a fixed sleep. On timeout fetchRoomSockets returns [] + logs;
+            // we still ack (best effort) — worse than a confirmed barrier
+            // but no worse than a fixed-sleep fallback.
+            await fetchRoomSockets(io, room, RPC_REGISTER_BARRIER_TIMEOUT_MS);
             socket.emit('rpc-registered', { method });
         } catch (error) {
             log({ module: 'websocket', level: 'error' }, `Error in rpc-register: ${error}`);


ratio-dd:
Hey @bra1nDump — moving this to draft after evaluating against `main`.

**Context:** After rebasing onto current `main` (includes your `readCount 100→2000`, exponential backoff on `fetchSockets`, hardened presence poll, and the new `run-all.sh` test runner), I reran the original reproducer (`test-1120-load.mjs`, 10 noise daemons + victim, 100 rounds × 10 methods).

**Results on 3-replica minikube + Redis streams:**
| Config | Victim failures |
|---|---|
| Old main (pre your fixes) — historical | 1/600 (0.17%) |
| New main + this PR's barrier | **0/1000** |
| New main alone (no barrier) | **0/1000** |
| `run-all.sh --safe-only` on new main + barrier | 7/7 RPC tests pass |

**Takeaway:** your `readCount 100→2000` alone seems to have suppressed the observable failure to below my test's detection threshold. This PR's barrier (awaiting a cluster-adapter `fetchSockets` round-trip after `socket.join()` to guarantee peer backlog is drained before acking `rpc-registered`) becomes belt-and-suspenders in this regime.

**Why I'm leaving the PR open as draft rather than closing:**
- The barrier gives a *causal* guarantee (ack only happens after every known peer has XREAD past the FETCH_SOCKETS offset), whereas `readCount=2000` gives a *probabilistic* one (wider window, still a window). At sufficient churn + replica count, the same race could theoretically re-emerge.
- Single-replica deployments short-circuit inside the adapter (`expectedResponseCount === 0`), so there's no cost there.
- If my test scale (10 noise daemons) is nowhere near prod, the barrier might still earn its keep under heavier load — I don't have visibility into your prod traffic shape.

**Questions for you:**
1. Do you think the `readCount=2000` change is sufficient for your prod load, or do you want an application-level guarantee as well?
2. If interested, I can push harder on the test (500–1000 rounds, more noise) to see if a regime exists where the barrier still measurably helps. Otherwise happy to close this as subsumed by your fix.

No rush — ship what you've got; this is a "nice to have" at best given your data.

bra1nDump:
Thanks for this fix and for helping with debugging these annoying rpc issues! I am inclined to not merge this for now to avoid complicating the client with more events.

I have also added some metrics on the backend to track failed rpcs and redis health metrics, should signal us if there are more issues. Ping me if you run into new related issues. Thanks again for the help!

PR #1154, opened 2026-04-21
perf(chat): fix UI freeze entering large chats on tablets

Evyatar108:
## Summary

Entering a large chat on a tablet freezes the entire app UI — even the Back button becomes unresponsive. Both panes (permanent sidebar + chat) are mounted simultaneously on tablets, which amplifies a classic React Native trap: `maintainVisibleContentPosition` on an inverted FlatList with variable-height cells and no `getItemLayout` forces synchronous measurement of candidate-anchor items on mount. Cost scales with `window size × message count`, and the wider tablet viewport pulls more items into that initial window than a phone does, tipping the measurement pass into a multi-second JS-thread block. Most painful on constrained Android devices (e.g. e-ink tablets), where it turns into an outright hang.

## Root cause (more detail)

In `packages/happy-app/sources/components/ChatList.tsx`, the `FlatList` was configured with:
```tsx
inverted={true}
maintainVisibleContentPosition={{
    minIndexForVisible: 0,
    autoscrollToTopThreshold: 10,
}}
// no initialNumToRender / windowSize / maxToRenderPerBatch
// no getItemLayout
```

On `inverted={true}`, items appear at the bottom of the list when data is prepended — this is already the natural behavior of the inverted scroller, so `maintainVisibleContentPosition: { minIndexForVisible: 0 }` is mostly redundant here. But it still pays the full cost: RN walks and measures items looking for the anchor on mount, synchronously, on the JS thread. With hundreds of messages and a wide tablet viewport fitting 20+ cells on screen, this blocks the thread long enough that touch input never gets processed.

Entering split-pane layout on tablets (`SidebarNavigator.tsx` — permanent drawer via `drawerType: 'permanent'`, `lazy: false`) compounds the issue: both the sessions list AND the chat mount at the same time, each competing for the JS thread.

## Changes

- **Drop `maintainVisibleContentPosition`.** On `inverted={true}` the new-message-at-bottom behaviour falls out of the list naturally; dropping the prop removes the synchronous measurement pass.
- **Add conservative virtualization props**: `initialNumToRender=8`, `maxToRenderPerBatch=4`, `windowSize=5`, `removeClippedSubviews=true`. Caps the mount-time work on devices with wide viewports.
- **Bump `scrollEventThrottle` 16 → 32.** Halves the JS work during scroll on constrained devices.
- **Wrap `MessageView` in `React.memo`.** Stable message references stop re-rendering when the parent's `showScrollButton` state toggles during scroll.

No visible change on phones. Significant mount-time improvement on tablets.

## Testing

**Proof-of-fix on device is still pending** — this PR is marked draft until I can record a before/after on an Android e-ink tablet where the freeze reproduces reliably. Sharing the diagnosis + patch now because the root cause is clear and the diff is small.

- [x] `pnpm --filter happy-app typecheck` — clean
- [ ] Tested on Android e-ink tablet (large chat entry) — **pending**
- [ ] Tested on phone to confirm no regression (small + large chats, new-message auto-scroll) — **pending**
- [ ] Video/screenshots of before/after — **pending**

I'll un-draft once the device tests are recorded.

## Notes / follow-ups (not in this PR)

- `MessageView.tsx:22` sets `renderToHardwareTextureAndroid={true}` on every message wrapper. On devices without a real GPU (e-ink), this is actively harmful (texture upload stalls). Worth a separate discussion — leaving untouched here to keep this PR focused on the list-perf root cause.
- `FlashList` would be a better long-term fit for inverted chat lists with variable heights. Also out of scope for this fix.

### DIFF
--- a/packages/happy-app/sources/components/ChatList.tsx
+++ b/packages/happy-app/sources/components/ChatList.tsx
@@ -69,6 +69,15 @@ const ChatListInternal = React.memo((props: {
                 data={props.messages}
                 inverted={true}
                 keyExtractor={keyExtractor}
+                // Restored after the initial perf fix: the tight virtualization props
+                // below cap the mount-time measurement cost that made this expensive
+                // on large chats, so it no longer blocks the JS thread. Without
+                // maintainVisibleContentPosition, users scrolled up to read history
+                // see the viewport jump when new messages prepend at data[0].
+                maintainVisibleContentPosition={{
+                    minIndexForVisible: 0,
+                    autoscrollToTopThreshold: 10,
+                }}
                 initialNumToRender={8}
                 maxToRenderPerBatch={4}
                 windowSize={5}
--- a/packages/happy-app/sources/components/ChatList.tsx
+++ b/packages/happy-app/sources/components/ChatList.tsx
@@ -69,15 +69,15 @@ const ChatListInternal = React.memo((props: {
                 data={props.messages}
                 inverted={true}
                 keyExtractor={keyExtractor}
-                maintainVisibleContentPosition={{
-                    minIndexForVisible: 0,
-                    autoscrollToTopThreshold: 10,
-                }}
+                initialNumToRender={8}
+                maxToRenderPerBatch={4}
+                windowSize={5}
+                removeClippedSubviews={true}
                 keyboardShouldPersistTaps="handled"
                 keyboardDismissMode={Platform.OS === 'ios' ? 'interactive' : 'none'}
                 renderItem={renderItem}
                 onScroll={handleScroll}
-                scrollEventThrottle={16}
+                scrollEventThrottle={32}
                 ListHeaderComponent={<ListFooter sessionId={props.sessionId} />}
                 ListFooterComponent={<ListHeader />}
             />
--- a/packages/happy-app/sources/components/MessageView.tsx
+++ b/packages/happy-app/sources/components/MessageView.tsx
@@ -12,7 +12,7 @@ import { sync } from '@/sync/sync';
 import { Option } from './markdown/MarkdownView';
 
 
-export const MessageView = (props: {
+export const MessageView = React.memo((props: {
   message: Message;
   metadata: Metadata | null;
   sessionId: string;
@@ -30,7 +30,7 @@ export const MessageView = (props: {
       </View>
     </View>
   );
-};
+});
 
 // RenderBlock function that dispatches to the correct component based on message kind
 function RenderBlock(props: {


Evyatar108:
## On-device verification

Verified the fix on the device class where the bug reliably reproduces.

**Environment**
- Android e-ink tablet (Boox-class, weak CPU + software compositor)
- Local dev build (`com.slopus.happy.dev`) produced from this PR branch via `expo run:android`, installed over adb
- Server: self-hosted happy-server on localhost exposed via Cloudflare Tunnel
- Chat paired from existing phone account; includes one large session ("many messages" — exact count not checked, but scroll-to-top takes several seconds of content)

**Before (Play Store `com.ex3ndr.happy`, same account, same large chat)**
- Tapping into the large chat from the permanent sidebar → UI freezes completely within ~1 second.
- Back button, scroll gestures, sidebar taps — all unresponsive. Not a crash (the UI remains on screen); the JS thread is blocked.
- Small chat in the same session list opens and scrolls fine.
- Same large chat opens and scrolls fine on the paired phone.

**After (this PR branch, dev build)**
- Large chat opens immediately.
- Scroll / back / tap all responsive.
- No visible behavioural regressions on the sidebar chat list, message rendering, or auto-scroll-on-new-message on either tablet or phone.

Happy to record a before/after screen recording if a reviewer wants one — I skipped it here because a static screenshot of the frozen state is indistinguishable from the working one, and didn't want to delay posting for evidence that would only land if explicitly asked for.

Un-drafting now. Follow-up items from my earlier notes (e-ink-specific \`renderToHardwareTextureAndroid={true}\` on every message, and a \`FlashList\` migration) remain out of scope for this PR.

Evyatar108:
## Follow-up commit: restore scroll anchoring

Added \`0f629a7b\` restoring \`maintainVisibleContentPosition\` after a second review round flagged that my original removal was too aggressive.

**Why it was removed in the first place:** the prop forced synchronous measurement of candidate-anchor items on mount, and on an inverted \`FlatList\` with variable-height cells and no \`getItemLayout\`, that measurement pass blocked the JS thread for seconds on large chats — the actual freeze.

**Why it needs to come back:** without it, new messages prepending at \`data[0]\` (sorted newest-first, inverted → visual bottom) shift the content while the user's pixel scroll offset stays fixed. A user scrolled up reading history would see the viewport jump when messages arrive during an active stream.

**Why both can coexist now:** the virtualization props added in the same PR (\`initialNumToRender=8\`, \`maxToRenderPerBatch=4\`, \`windowSize=5\`, \`removeClippedSubviews=true\`) cap the mount-time measurement pass to ~8-13 visible cells. That's fast enough not to block the JS thread, so the freeze fix holds.

Net: the PR now preserves both the perf win and scroll stability. Verified freeze fix holds on Android e-ink tablet with my local dev build; will verify scroll-anchor behaviour on the same device before updating the PR body checklist.

PR #1186, opened 2026-04-26
fix(openclaw): isolate per-Happy-session gateway sessionKey

### LINKED ISSUE #1185: OpenClaw multi-session collapse: every Happy tab pins to `agent:<id>:main`
# OpenClaw multi-session collapse: every Happy tab pins to `agent:<id>:main`

**Summary:** When using the OpenClaw agent backend, every Happy session — across all tabs, devices, and "new chat" tabs on the same daemon — is pinned to the same gateway session `agent:<agentId>:main`. Opening a second Happy session while the first is mid-turn either blocks the new session behind the first turn, merges both transcripts into one, or both. This regressed the per-chat isolation that the original Clawdbot prototype (#481) explicitly implemented as *"New chat creates isolated sessions (not using main session)"* — that design was dropped during the OpenClaw rebrand (#816).

## Repro

1. Run `openclaw` gateway locally.
2. `happy daemon start`, then on the Happy app open one OpenClaw session and send a slow prompt (e.g. ask it to summarize a long file with a non-Haiku model).
3. Within a few seconds, open a second Happy chat (new tab, or new device pointing at the same daemon) — also OpenClaw.
4. Send a different prompt in tab 2.

**Observed:**
- Tab 2's prompt does not start until tab 1's turn finishes (gateway queues turns per-sessionKey).
- Tab 2 sees tab 1's reply text streaming in alongside its own.
- `~/.openclaw/agents/main/sessions/sessions.json` contains a single `agent:main:main` entry regardless of how many Happy chats are open. The single transcript file (`agent/main/sessions/<uuid>.jsonl`) grows unbounded as both tabs' turns are appended.
- Each subsequent turn re-sends the combined history to the model → prompt-cost inflation.

**Expected:** Each Happy session gets an isolated gateway session (parallel turns, separate transcripts), matching how Telegram (`agent:<id>:telegram:group:<chatId>`) and Discord channels already scope per-chat.

## Root cause

`OpenClawBackend.startSession()` ([packages/happy-cli/src/openclaw/OpenClawBackend.ts](https://github.com/slopus/happy/blob/main/packages/happy-cli/src/openclaw/OpenClawBackend.ts#L104)) hardcodes:

```ts
this.sessionKey = this.socket.getMainSessionKey();
```

`getMainSessionKey()` returns the gateway-provided default `agent:<agentId>:main`, which is shared by design — the gateway warns `session.mainKey is ignored; main session is always "main".` All Happy tabs/devices use the same key, so they collide on a single gateway session. Per-channel routing in OpenClaw relies on each channel constructing its own scoped key (e.g. `agent:<id>:telegram:group:<chatId>`); Happy's bridge never did that, so it inherits the catch-all main session.

The gateway accepts arbitrary suffixes — `toAgentStoreSessionKey` in `dist/session-key-Cn2QoOyX.js` parses any `agent:<id>:<rest>` shape and routes to a per-key store. So the fix lives entirely on the Happy side.

## Diagnostics

- `journalctl --user -u openclaw.service | grep 'stuck session'` shows `sessionKey=agent:main:main state=processing queueDepth>=N` while a second Happy tab waits.
- `ls -laS ~/.openclaw/agents/main/sessions/` shows one transcript file growing unbounded (we observed 12 MB in a regular dev session).
- Switch to a fast model (Haiku) and the symptom hides because turns finish before the queue effect is visible.

## Proposed fix

Have `OpenClawBackend` construct a per-Happy-session gateway sessionKey of shape `agent:<agentId>:happy:<sessionTag>`, where `<sessionTag>` is the unique tag already generated per `runOpenClaw` invocation. Mirrors how Telegram/Discord scope per-chat. Adds a `HAPPY_OPENCLAW_SESSION_KEY` env override for users who want the legacy shared-session behavior or a custom key.

Code change is small and contained to happy-cli (no server, RPC, or encryption changes). PR forthcoming.

## Environment

- happy-cli (upstream main as of 2026-04-26)
- openclaw gateway local v3
- Reproduced on web Happy app + multiple browser tabs against one daemon.

chphch:
Fixes #1185.

`OpenClawBackend.startSession()` was hardcoding `socket.getMainSessionKey()`, so every Happy session — all tabs, all devices on one daemon — pinned to the gateway's shared `agent:<id>:main` session. A slow turn in one tab queued every other Happy session, transcripts merged across tabs, and each subsequent turn re-sent the combined history (prompt-cost inflation). This patch constructs an isolated per-Happy-session gateway sessionKey of shape `agent:<agentId>:happy:<sessionTag>`, mirroring how Telegram/Discord channels already scope per-chat. `<sessionTag>` is the UUID already generated per `runOpenClaw` spawn, so each Happy session naturally lands on its own gateway session.

The fix lives entirely in happy-cli — the OpenClaw gateway already routes any `agent:<id>:<rest>` shape through `toAgentStoreSessionKey` to a per-key store (see `dist/session-key-Cn2QoOyX.js:106-117`). No server, RPC, or encryption changes.

`HAPPY_OPENCLAW_SESSION_KEY` env var added as an opt-out for users who want the legacy shared-session behavior or a custom key.

## Files

- `packages/happy-cli/src/openclaw/OpenClawBackend.ts` — `sessionTag` option + `buildHappySessionKey` / `resolveOpenClawSessionKey` helpers.
- `packages/happy-cli/src/openclaw/runOpenClaw.ts` — pass `sessionTag` to backend.
- `packages/happy-cli/src/openclaw/openclawSessionKey.test.ts` — 8 unit tests covering env override, sessionTag isolation, legacy fallback, and edge cases.

## Background

Restores the per-chat isolation that the original Clawdbot prototype (#481) explicitly implemented as *"New chat creates isolated sessions (not using main session)"* — that design was dropped during the OpenClaw rebrand (#816).

Proof of multi-session isolation (before/after) attached as a comment.

### DIFF
--- a/packages/happy-cli/src/openclaw/OpenClawBackend.ts
+++ b/packages/happy-cli/src/openclaw/OpenClawBackend.ts
@@ -39,6 +39,44 @@ export interface OpenClawBackendOptions {
   clientMode?: OpenClawSocketOptions['clientMode'];
   displayName?: string;
   log?: (msg: string) => void;
+  /**
+   * Per-Happy-session unique tag, used to construct an isolated gateway
+   * sessionKey (`agent:<agentId>:happy:<sessionTag>`). Without this, every
+   * Happy session collapses onto the gateway's shared `agent:<id>:main`
+   * session, which queues turns and merges transcripts across tabs/devices.
+   * Mirrors the per-chat scoping used by Telegram/Discord channels.
+   */
+  sessionTag?: string;
+}
+
+/**
+ * Build a per-Happy-session gateway sessionKey from the gateway-provided
+ * main key (e.g. `agent:main:main`) and a unique tag. The gateway accepts
+ * arbitrary suffixes — see Telegram's `agent:<id>:telegram:group:<chatId>`
+ * pattern. Falls back to mainKey if parsing fails (defensive).
+ */
+export function buildHappySessionKey(mainKey: string, sessionTag: string): string {
+  // mainKey is `agent:<agentId>:<mainKey>` (typically `agent:main:main`).
+  const parts = mainKey.split(':');
+  if (parts.length < 3 || parts[0] !== 'agent' || !parts[1]) return mainKey;
+  return `agent:${parts[1]}:happy:${sessionTag}`;
+}
+
+/**
+ * Resolve the gateway sessionKey for a Happy session. Priority:
+ *   1. `HAPPY_OPENCLAW_SESSION_KEY` env override (any string)
+ *   2. Per-tag isolation: `agent:<agentId>:happy:<sessionTag>` derived from mainKey
+ *   3. Legacy: gateway-provided mainKey (shared across all Happy sessions)
+ */
+export function resolveOpenClawSessionKey(params: {
+  mainKey: string;
+  sessionTag?: string;
+  envOverride?: string;
+}): string {
+  const trimmedOverride = params.envOverride?.trim();
+  if (trimmedOverride) return trimmedOverride;
+  if (params.sessionTag) return buildHappySessionKey(params.mainKey, params.sessionTag);
+  return params.mainKey;
 }
 
 function extractTextFromMessage(message?: OpenClawChatMessage): string {
@@ -59,6 +97,7 @@ export class OpenClawBackend implements AgentBackend {
   private gatewayConfig: OpenClawGatewayConfig;
   private handlers = new Set<AgentMessageHandler>();
   private sessionKey: string | null = null;
+  private sessionTag: string | undefined;
   private lastDeltaText: string | null = null;
   private log: (msg: string) => void;
 
@@ -75,6 +114,7 @@ export class OpenClawBackend implements AgentBackend {
   constructor(opts: OpenClawBackendOptions) {
     this.log = opts.log ?? (() => {});
     this.gatewayConfig = opts.gatewayConfig;
+    this.sessionTag = opts.sessionTag;
     this.socket = new OpenClawSocket({
       homeDir: opts.homeDir,
       clientId: opts.clientId,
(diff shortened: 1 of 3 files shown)


chphch:
## Proof — alternating multi-turn across two tabs

Two browser tabs side-by-side, recorded with Playwright. Each tab takes **two turns**, alternating: `A1 → B1 → A2 → B2`. Identical action sequence in both phases — only the `happy-cli` build differs.

### Before (`upstream/main`)

BEFORE — alternating multi-turn

A and B both pin to the gateway's shared `agent:main:main`. Final state of both tabs:

```
Tab A                     Tab B
─────────────             ─────────────
You: A1 prompt            You: B1 prompt
You: A2 prompt            You: B2 prompt
AI:  A1-BEFORE-89H4       AI:  A1-BEFORE-89H4   ← same reply, leaked from Tab A
                          (no other replies — runOpenClaw turn-end fires early on broadcast)
```

Both tabs end up rendering **the same reply** (`A1-BEFORE-89H4`) — Tab A's first turn — because the gateway broadcasts every chat event to every connected socket, and B's runOpenClaw forwards A's `text` envelope to B's Happy session. Neither tab gets its own 2-turn conversation; B never sees `B1`/`B2` replies.

### After (this PR)

AFTER — alternating multi-turn

A lands on `agent:main:happy:<sessionTag-A>`, B on `agent:main:happy:<sessionTag-B>`, and each `OpenClawBackend` filters incoming chat events by sessionKey. Final state:

```
Tab A                     Tab B
─────────────             ─────────────
You: A1 prompt            You: B1 prompt
AI:  A1-AFTER-ON0V        AI:  B1-AFTER-ON0V
You: A2 prompt            You: B2 prompt
AI:  A2-AFTER-ON0V        AI:  B2-AFTER-ON0V
```

Each tab keeps a clean independent multi-turn conversation, both running in parallel against the gateway. This matches the issue's "you should be able to open multiple Happy sessions and have each conversation work" expectation.

### Setup

Sandbox stack — standalone `happy-server` (PGlite, port 3055), Expo web (port 8088), a fresh `happy-cli` daemon (`HAPPY_HOME_DIR=/tmp/happy-proof-1186-home`, registered as a separate machine on the test server), against the live OpenClaw gateway at `ws://127.0.0.1:18789`. Two Playwright contexts, each authed via `getDevWebQueryCredentials` URL bypass and recording its own webm; ffmpeg `hstack`-ed into the side-by-side GIFs above. Per-phase nonce in the prompts (`A1-<PHASE>-<NONCE>` / `B1-<PHASE>-<NONCE>` / …) prevents OpenClaw's `memory_search` RAG from short-circuiting either reply through indexes of past sessions.

Full webm recordings (no quality loss): [`1186-before-alt.webm`](https://github.com/chphch/happy/releases/download/pr-proofs/1186-before-alt.webm), [`1186-after-alt.webm`](https://github.com/chphch/happy/releases/download/pr-proofs/1186-after-alt.webm).

chphch:
Closing this one out — we've since stopped using the OpenClaw backend on our side, so I'm no longer running or maintaining this fix. If per-session gateway isolation becomes relevant upstream, the diff is here to salvage.

PR #1211, opened 2026-05-02
fix(happy-cli): respect attribution.commit empty string setting

Puppo:
When users set attribution.commit to "" in their Claude settings.json, Happy now skips injecting Co-Authored-By trailers, letting Claude Code's own attribution handling work instead of overriding it.

Issue: #1180


ilteoood:
+1 on this, would be amazing to see the landed

PR #1212, opened 2026-05-03
fix(cli): surface hook `systemMessage` in chat (was silently dropped)

chphch:
## Bug

Claude Code hooks can emit `{"systemMessage": "..."}` JSON to signal a user-visible note. For example a `UserPromptSubmit` hook that exits non-zero with:

```bash
echo '{"hookSpecificOutput":{"decision":"BLOCK"},"systemMessage":"⏰ Workout time!"}' >&2
exit 2
```

In `claude` (terminal), the user sees `⏺ Workout time!` and the prompt is blocked. In Happy (mobile/web), **the prompt is blocked but no message appears** — the user has no idea why their input vanished. Same for `SessionStart` summary hooks, `PreToolUse` warnings, etc.

## Root cause

For current `happy-cli`, every outgoing log message goes:

```
SDK message → sdkToLogConverter.convert() → OutgoingMessageQueue.enqueue()
            → sendFunction → apiSession.sendClaudeSessionMessage()
            → mapClaudeLogMessageToSessionEnvelopes() → service envelope → server → app
```

Two cuts on this path silently dropped hook output before it could ever reach the chat:

1. **`OutgoingMessageQueue.ts`** had `if (item.logMessage.type !== 'system') sendFunction(...)` — an unconditional drop, so *every* `type === 'system'` message (`init`, `hook_started`, `hook_progress`, **and** `hook_response`) was discarded before the mapper ran.
2. **`sessionProtocolMapper.ts`**'s `if (message.type === 'system')` branch returned no envelopes — even if the queue had let the message through, the mapper had no `hook_response` handler.

Two ancillary paths that aren't on the hot path also lacked handling:

3. **`typesRaw.ts:normalizeRawMessage`** (app-side, legacy raw-output format used by older clients) had no `type === 'system'` branch.
4. **`messageFormatterInk.ts`** (CLI Ink terminal renderer in remote mode) only handled `subtype: 'init'` for system messages.

## Fix

| File(s) | What |
|---|---|
| `OutgoingMessageQueue.ts` + `sessionProtocolMapper.ts` | **The user-visible fix.** Stop dropping `type === 'system'` at the queue and let the mapper decide which subtypes produce envelopes; emit a `service` envelope from `hook_response` carrying the `systemMessage` payload, extracted from `output`/`stdout`/`stderr`. The app's existing `normalizeSessionEnvelope` already renders service envelopes as agent text, so nothing new is needed app-side for the live path. |
| `typesRaw.ts` | App normalizer for the legacy raw-output format still used by older clients: extract `systemMessage` from the hook payload and render it as agent text. |
| `messageFormatterInk.ts` | CLI Ink terminal renderer: show `systemMessage` with a `⏺` prefix in remote-mode display. |

`sdkToLogConverter` already passes system messages through unchanged — no change needed there.

## Proof — live e2e on a real CLI ↔ server ↔ app stack

BEFORE/AFTER side-by-side screenshot

Same dev-token account on both halves: standalone `happy-server` (PGlite) + Expo web + a `HAPPY_HOME_DIR`-isolated `happy daemon`. A `SessionStart` hook in the user's global `~/.claude/settings.json` emits a startup banner (`{"continue":true,"systemMessage":"🚀 cwd Session started\n▸ Time:…\n▸ Directory:…"}`). Playwright drives the "Start New Session" modal, sends `hello there`, and screenshots the chat after the turn.

- **BEFORE (upstream/main)** — only the user prompt is visible. The hook fired (confirmed in the CLI session log) and the SDK emitted `hook_response`, but it never reached the server, so the chat shows nothing.
- **AFTER (this branch)** — the same hook's `systemMessage` renders as a chat bubble.

## Verification

- **17 vitest cases** across 3 spec files, all green: 5 in `sessionProtocolMapper.test.ts` (the live-path fix), 7 in `typesRaw.spec.ts` (legacy normalizer), 5 in `messageFormatterInk.test.ts` (CLI Ink). Cases cover blocking-hook stderr (exit 2), success-hook stdout (exit 0), hooks without `systemMessage`, non-JSON output, malformed JSON, `hook_started`/`hook_progress`, and other system subtypes.
- `pnpm typecheck` (CLI + app) clean — no new errors.

### DIFF
--- a/packages/happy-app/sources/sync/typesRaw.spec.ts
+++ b/packages/happy-app/sources/sync/typesRaw.spec.ts
@@ -1995,4 +1995,191 @@ describe('Zod Transform - WOLOG Content Normalization', () => {
             expect(normalized).toBeNull();
         });
     });
+
+    describe('Hook system messages (subtype: hook_response)', () => {
+
+        const baseHook = {
+            role: 'agent' as const,
+            content: {
+                type: 'output' as const,
+                data: {
+                    type: 'system',
+                    subtype: 'hook_response',
+                    hook_id: 'h1',
+                    hook_name: 'UserPromptSubmit:workout',
+                    hook_event: 'UserPromptSubmit',
+                    uuid: 'hook-uuid-1',
+                    parentUuid: null,
+                    session_id: 's1',
+                }
+            }
+        };
+
+        it('surfaces systemMessage from a blocking hook (stderr + exit_code 2)', () => {
+            const stderrJson = JSON.stringify({
+                hookSpecificOutput: { decision: 'BLOCK' },
+                systemMessage: '⏰ Workout time! All sessions blocked until 06:40.'
+            });
+            const normalized = normalizeRawMessage('msg-hook-1', null, 1, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        output: stderrJson,
+                        stdout: '',
+                        stderr: stderrJson,
+                        exit_code: 2,
+                        outcome: 'error',
+                    }
+                }
+            } as any);
+
+            expect(normalized).toBeTruthy();
+            expect(normalized?.role).toBe('agent');
+            if (normalized && normalized.role === 'agent') {
+                expect(normalized.content).toHaveLength(1);
+                expect(normalized.content[0]).toMatchObject({
+                    type: 'text',
+                    text: '⏰ Workout time! All sessions blocked until 06:40.',
+                    uuid: 'hook-uuid-1',
+                    parentUUID: null,
+                });
+            }
+        });
+
+        it('surfaces systemMessage from a passing hook (stdout + exit_code 0)', () => {
+            const stdoutJson = JSON.stringify({
+                continue: true,
+                systemMessage: '🚀 Session started'
+            });
+            const normalized = normalizeRawMessage('msg-hook-2', null, 2, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        hook_event: 'SessionStart',
+                        hook_name: 'SessionStart:summary',
+                        output: stdoutJson,
+                        stdout: stdoutJson,
+                        stderr: '',
+                        exit_code: 0,
+                        outcome: 'success',
+                    }
+                }
+            } as any);
+
+            expect(normalized).toBeTruthy();
+            if (normalized && normalized.role === 'agent') {
+                expect(normalized.content[0]).toMatchObject({
+                    type: 'text',
+                    text: '🚀 Session started',
+                });
+            }
+        });
+
+        it('drops hook_response with no systemMessage (e.g. plain logging hooks)', () => {
+            const normalized = normalizeRawMessage('msg-hook-3', null, 3, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        output: 'Sync started in background\n',
+                        stdout: 'Sync started in background\n',
+                        stderr: '',
+                        exit_code: 0,
+                        outcome: 'success',
+                    }
+                }
+            } as any);
+
+            expect(normalized).toBeNull();
+        });
+
+        it('drops hook_response with non-JSON stderr', () => {
+            const normalized = normalizeRawMessage('msg-hook-4', null, 4, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        output: 'Error: command not found\n',
+                        stdout: '',
+                        stderr: 'Error: command not found\n',
+                        exit_code: 127,
+                        outcome: 'error',
+                    }
+                }
+            } as any);
+
+            expect(normalized).toBeNull();
+        });
+
+        it('drops hook_started / hook_progress (no systemMessage to show)', () => {
+            const started = normalizeRawMessage('msg-hook-5', null, 5, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        subtype: 'hook_started',
+                    }
+                }
+            } as any);
+            expect(started).toBeNull();
+
+            const progress = normalizeRawMessage('msg-hook-6', null, 6, {
+                ...baseHook,
+                content: {
+                    ...baseHook.content,
+                    data: {
+                        ...baseHook.content.data,
+                        subtype: 'hook_progress',
+                        stdout: 'progress',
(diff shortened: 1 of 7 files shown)


chphch:
## Proof

This fix is a pure normalizer change in `typesRaw.ts` — its observable behavior is whether a `system / hook_response` raw message becomes a renderable agent text or returns `null`. That collapses neatly into unit tests, so the proof here is **7 vitest cases that pin the contract** plus a one-shot manual reproduction.

### Unit tests (7/7 pass — covers all `subtype` × output-channel × payload shape combinations)

```
$ pnpm vitest run sources/sync/typesRaw.spec.ts -t "Hook system messages"

 ✓ Hook system messages (subtype: hook_response) > surfaces systemMessage from a blocking hook (stderr + exit_code 2) 5ms
 ✓ Hook system messages (subtype: hook_response) > surfaces systemMessage from a passing hook (stdout + exit_code 0) 1ms
 ✓ Hook system messages (subtype: hook_response) > drops hook_response with no systemMessage (e.g. plain logging hooks) 0ms
 ✓ Hook system messages (subtype: hook_response) > drops hook_response with non-JSON stderr 0ms
 ✓ Hook system messages (subtype: hook_response) > drops hook_started / hook_progress (no systemMessage to show) 0ms
 ✓ Hook system messages (subtype: hook_response) > drops other system subtypes (e.g. init) — they are not user-facing 0ms
 ✓ Hook system messages (subtype: hook_response) > handles malformed JSON in stdout/stderr without throwing 0ms

 Test Files  1 passed (1)
      Tests  66 passed (66)   ← full file: existing 59 still green
```

The full `typesRaw.spec.ts` suite (66 tests, ~830 ms) stays green; `pnpm typecheck` introduces zero new errors against this diff.

### Manual reproduction recipe

If a reviewer wants to see it in a real session:

```bash
# 1. Add a UserPromptSubmit hook to ~/.claude/settings.json
{
  "hooks": {
    "UserPromptSubmit": [{
      "matcher": "*",
      "hooks": [{"type":"command","command":"bash /tmp/test-hook.sh","timeout":5}]
    }]
  }
}

# 2. /tmp/test-hook.sh — block + show a message
#!/bin/bash
echo '{"hookSpecificOutput":{"decision":"BLOCK"},"systemMessage":"⏰ TEST: blocked by hook"}' >&2
exit 2

# 3. Open Happy on web/mobile, send any prompt.
#    BEFORE this PR: prompt is silently dropped — no message, no event.
#    AFTER this PR:  "⏰ TEST: blocked by hook" renders as a system note.
```

Same hook in stock `claude` (terminal) renders the message as a gray system line — this PR brings Happy to parity.

### Why no end-to-end GIF

The proof-capture skill in this repo can drive Expo web headlessly with a synthetic auth, but to fire a hook end-to-end I'd need to also boot a real `happy-cli` daemon paired to that synthetic account and inject a hook into its `~/.claude/settings.json`. The skill's docs explicitly call out CLI-coupled features as out of scope, and the unit suite already pins every observable branch. Happy to add an e2e test later if the project picks up Playwright infra.

chphch:
## Proof — live e2e on real CLI ↔ server ↔ app stack

BEFORE/AFTER side-by-side screenshot

Same dev-token account on both halves: standalone `happy-server` (PGlite, port 3055) + Expo web (port 8088) + a `HAPPY_HOME_DIR`-isolated `happy daemon` paired via hand-crafted `access.key`. A SessionStart hook in the user's global `~/.claude/settings.json` emits a startup banner JSON (`{"continue":true,"systemMessage":"🚀 cwd Session started\n▸ Time:…\n▸ Directory:…"}`). I drove the "Start New Session" modal in Playwright, sent `hello there`, and screenshotted the chat after the agent processed the turn.

- **BEFORE (upstream/main)** — only the user prompt is visible. The hook fired (verified in CLI session log), the SDK emitted `hook_response`, but it never reached the server, so the chat shows nothing about the hook.
- **AFTER (this branch)** — the same hook's `systemMessage` renders as a chat bubble.

## Where the bug actually lived

Initial PR (`f7bd0d0a` + `fc5eed4c`) targeted the legacy raw-output normalizer path (`typesRaw.ts`) and the CLI Ink terminal renderer (`messageFormatterInk.ts`). That covered the terminal-only flow. Driving the live setup uncovered that **current `happy-cli` doesn't use the legacy raw-output format anymore** — every outgoing log message goes through `apiSession.sendClaudeSessionMessage` → `mapClaudeLogMessageToSessionEnvelopes`. So the typesRaw branch I added was dead code from the user-visible app's perspective.

Two extra issues had to be fixed for the real bug:

1. **`OutgoingMessageQueue.ts:124`** dropped *every* `type === 'system'` log message before it ever reached `sendClaudeSessionMessage`:
   ```ts
   if (item.logMessage.type !== 'system') {  // ← unconditional drop
       this.sendFunction(item.logMessage);
   }
   ```
   This silently swallowed `hook_response` events alongside the noisy ones (`hook_started`, `hook_progress`, `init`). New commit `19cb0e61` removes the filter and lets the downstream mapper decide which subtypes produce envelopes.

2. **`sessionProtocolMapper.ts:468`** had a `type === 'system'` branch that returned no envelopes. `19cb0e61` adds a `hook_response` handler that extracts `systemMessage` from `output`/`stdout`/`stderr` and emits a `service` envelope (the app's existing `normalizeSessionEnvelope` already renders service envelopes as agent text).

## Tests

```
$ pnpm vitest run packages/happy-cli/src/claude/utils/sessionProtocolMapper.test.ts -t "hook_response"

 ✓ hook_response system messages > emits a service envelope when stderr carries {systemMessage} (block hook, exit 2)
 ✓ hook_response system messages > emits a service envelope when stdout carries {systemMessage} (passing hook, exit 0)
 ✓ hook_response system messages > emits no envelopes for hook_response without a systemMessage payload
 ✓ hook_response system messages > emits no envelopes for non-JSON stderr
 ✓ hook_response system messages > emits no envelopes for system subtypes other than hook_response (init, hook_started, etc.)

 Tests  5 passed, 11 skipped (16)   ← full mapper file: 16/16 green
```

Plus the prior 7 typesRaw + 5 messageFormatterInk vitest cases from earlier commits all stay green; CLI typecheck clean.

## Diff summary

| Commit | File | Path |
|---|---|---|
| `f7bd0d0a` | `typesRaw.ts` + `.spec.ts` | App normalizer (legacy raw-output path) |
| `fc5eed4c` | `messageFormatterInk.ts` + `.test.ts` | CLI Ink terminal renderer |
| `19cb0e61` | `OutgoingMessageQueue.ts` + `sessionProtocolMapper.ts` + `.test.ts` | **The live-bug fix** — session-protocol path |

Total: 17 vitest cases across 3 spec files; 0 new typecheck errors. Reproduction recipe and live-e2e capture script live in `/tmp/happy-proof-1212/` (also distilled into reusable helpers under `~/.claude/skills/happy-pr-proof-capture/scripts/lib/`).

Issue #636, opened 2026-02-16
Sort connections by last activity, show timestamp


nikhilsitaram:
## Problem

The connections list in the mobile/web app is ordered by working directory path. When you have several active sessions across different projects, finding the most recent one means scanning every entry manually. Path-based ordering doesn't reflect which sessions are actively being used.

## Proposed Solution

1. **Default sort by last activity** - Order connections by `metadata.summary.updatedAt` descending so the most recently active session appears first.
2. **Display a timestamp** - Show relative time (e.g., "2m ago", "3h ago") or an absolute timestamp next to each connection entry.

Either one of these would help. Both together would be ideal.

## Context

The CLI already sends `metadata.summary.updatedAt` (epoch ms) with each session's metadata, so the data is available without any backend changes. This is purely a frontend sorting/display change.

`lifecycleStateSince` could also be useful as a secondary sort key for separating running sessions from archived ones.

## Prior Art

Most tools with session/tab lists sort by recency by default: terminal multiplexers, browser tab managers, chat apps, IDE recent projects. Path-based sort is useful too, so a sort toggle ("Recent / Path / Name") would cover all use cases if changing the default is a concern.

PR #1214, opened 2026-05-03
feat(app): sort archived sessions by last seen activity (toggle)

chphch:
## Summary

Adds an inline sort picker for the archived (inactive) sessions list, surfaced where the data is — directly under the "Hide archived" toggle.

- **Started** (default): sort by `session.createdAt` (current behavior — newest started first)
- **Last seen**: sort by `session.activeAt` (newest activity first)

Date group headers ("Today" / "Yesterday" / "N days ago") are recomputed from the chosen field so the headers match the visible order in both modes.

Gated behind an experimental **`expArchiveSortByLastSeen`** setting, **off by default**. With it off the picker row is not emitted at all and archived sessions sort by `createdAt` exactly as upstream. Users opt in from Settings → Features.

## UX

A small right-aligned row appears under "Hide archived" with the two options inline as text pills: `↑↓ Started · Last seen`. The current selection is shown in bold; the other is dimmed. Tapping either pill switches directly to that ordering — no popup, no hidden state, both alternatives are visible at all times so the user always knows what they're switching between.

The row is rendered as a new `archive-sort` `SessionListViewItem` emitted by `useVisibleSessionListViewData` only when archive is visible (between the "Hide archived" toggle and the first date header). It collapses with the rest of the archive section.

## Why a picker, not a default change

#636 proposed flipping the default sort for the entire connections list and was closed `not_planned`. This is narrower:

- Archive section only — active sessions are unchanged. (Complements #1440's `sortSessionsByActivity`, which orders the *active* list; the two are independent settings.)
- Opt-in twice over — behind an experiment flag, and the default within it stays `Started`.
- Two equally valid mental models (when I started it vs when I last touched it), surfaced inline so each user picks.

## Scope

App-only. No server / sync protocol / encryption changes. `activeAt` is already on `Session` and exposed via `SessionRowData` for inactive sessions, so no new data plumbing is needed.

- Two settings: `archivedSessionsSortBy: 'createdAt' | 'lastSeenAt'` (the value, default `'createdAt'`) and `expArchiveSortByLastSeen: boolean` (the gate, default `false`)
- New `SessionListViewItem` variant: `{ type: 'archive-sort'; current: ... }`
- `useVisibleSessionListViewData` resorts + regroups + emits the inline picker row when the flag is on
- `SessionsList.tsx` renders the row as two `Pressable` text pills with active/inactive styling
- 10 language translations + spec update

Forward/backward compatible: older clients ignore the unknown setting fields via the existing partial-parse path in `settingsParse`.

## Proof

End-to-end run on standalone `happy-server` (PGlite) + Expo web, driven headlessly by Playwright. Cursor sprite injected as a DOM overlay so the click target is visible (Playwright headless does not capture the OS cursor).

The 5-frame GIF cycles a full there-and-back interaction:

1. Default sort. Indicator: **Started** · Last seen. Order reflects `createdAt` (newest started first): `bravo-feature` → `delta-spike` → `charlie-bug` → `alpha-project`.
2. Cursor moves to "Last seen" pill (showing what's about to be tapped).
3. After tap → list reorders inline. Indicator: Started · **Last seen**. Order reflects `activeAt` (newest activity first): `alpha-project` (1h) → `delta-spike` (7h) → `bravo-feature` (1d) → `charlie-bug` (5d). Date group headers (`Today` / `Yesterday` / `5 days ago`) recompute from the chosen field.
4. Cursor moves back to "Started" pill.
5. After tap → reverts to original order. Indicator back to **Started** · Last seen.

archive sort reorder

The 4 archived sessions in the GIF are synthetic fixtures injected via a tiny capture-only shim on `window.__happyStorage` (production code unchanged) — picked so the two orderings differ in 3 of 4 positions, making the reorder visually unambiguous.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `settings.spec.ts` — defaults include `archivedSessionsSortBy: 'createdAt'` and `expArchiveSortByLastSeen: false`
- [x] Web smoke (Playwright + standalone happy-server): inline picker appears below archive toggle, both pills always visible, list reorders end-to-end on tap, indicator reflects state in real time, fully reversible


chphch:
Closing this one.

Two reasons, and the second is the one that decides it.

The code has moved: upstream's session-list rework for Rig sessions (11b5f115) means this now conflicts in 11 places across `SessionsList.tsx`, `useVisibleSessionListViewData.ts` and `storage.ts`. That alone would just be work.

The part that actually settles it is that I have since dropped this approach from my own fork — gating the archived-session sort behind an `expArchiveSortByLastSeen` flag. Landing those eleven resolutions would mean proposing an implementation I no longer run myself, which is not a reasonable thing to ask a reviewer to take on.

If ordering archived sessions by last activity is worth having, it deserves rebuilding against the current list structure rather than replaying this. Thanks all the same.

Issue #615, opened 2026-02-14
Feature Request: Full terminal output mode (show all details like desktop)


mkmkkkkk:
## Problem

When using Happy on mobile, the app hides/simplifies some output that is visible in the desktop terminal. This makes it harder to debug or follow along with what Claude is doing.

## Request

Add a setting to show full, unfiltered terminal output — matching exactly what you'd see in the desktop terminal. Something like a "verbose mode" or "raw output" toggle in the app settings.

## Use Case

- Power users who want to see every tool call result, file diff, and intermediate output
- Debugging sessions where hidden details matter
- Users running `--yolo` mode who want full visibility without needing to check their desktop

## Additional note

The CLI already passes `--verbose` and `--output-format stream-json`, so the data is already being sent to the app. This is a renderer/UI-side change to display everything instead of collapsing/hiding certain content.

PR #1223, opened 2026-05-05
fix(session-protocol): preserve Claude tool_result content on tool-call-end

WEIFENG2333:
## Summary

In remote Claude sessions, the mobile/web client shows tool calls completing with no output — `Bash`, `Read`, `Grep`, etc. all render empty bodies after the spinner stops.

The Claude session-protocol mapper drops `tool_result.content` when it converts a SDK `user`/`tool_result` block into an envelope. Only `{ t: 'tool-call-end', call }` is emitted, so the app's reducer marks the tool completed and writes `result = null`. The legacy `role: 'agent', content: { type: 'output', data }` fallback that used to carry the raw message was removed in the same change, so the content is gone end-to-end.

Reproduces on every remote Claude session ≥ 1.1.7. The codex mapper does not have this bug because it emits the content as a separate text envelope before `tool-call-end`.

## Root cause

Introduced in [`36bcc1c`](https://github.com/slopus/happy/commit/36bcc1cb0cd46e0c6f9640a37713f5c0b15263a7) (`feat(session-protocol): finalize subagent lifecycle protocol`):

- `packages/happy-cli/src/claude/utils/sessionProtocolMapper.ts` was added with the lossy `tool_result → tool-call-end` mapping.
- `packages/happy-cli/src/api/apiSession.ts` lost the `role: 'agent', content: { type: 'output', data: body }` raw-passthrough branch in the same commit, so there is no other channel carrying the result.

The codex mapper added at the same time (`b0c781c`) intentionally pre-emits the content as a `text` envelope. The Claude side appears to have just been missed — this is consistent with the docs/example in `docs/session-protocol-claude.md` (Example 1) where the input `tool_result.content` literally has no corresponding output envelope.

Tracking issue: #615.

## Fix

Carry the result on the `tool-call-end` envelope itself, so the wire format has a single source of truth for the tool's lifecycle (rather than a sibling text block that the reducer would have to associate by ordering).

- `happy-wire`: extend `sessionToolCallEndEventSchema` with optional `output: string` and `isError: boolean`.
- `happy-cli`: in the Claude mapper, populate `output` from the SDK `tool_result.content` (`string | Array<{ type: 'text', text }>`) and set `isError` when `is_error: true`.
- `happy-app`: in `normalizeSessionEnvelope`, forward `output` / `isError` into the normalized `tool-result` content. Reducer logic for `result` and the running → completed/error transition is unchanged.

Both fields are optional and additive, so older clients that only see `{ t: 'tool-call-end', call }` keep working.

## Test plan

- [x] `pnpm --filter @slopus/happy-wire test` — 19/19 (existing + new envelope-shape cases for `output` / `isError`).
- [x] `pnpm vitest run src/claude/utils/sessionProtocolMapper.test.ts` in `happy-cli` — 14/14, including new cases for string content, array-of-text content, error results, and empty content.
- [x] `pnpm vitest run sources/sync/typesRaw.spec.ts` in `happy-app` — 60/60, including the new `tool-call-end → tool-result` round-trip with `output`/`isError`.
- [x] `pnpm typecheck` (happy-app) clean.
- [x] Verified the four pre-existing test-file failures on `main` (`apiGithub.spec.ts`, `modelModeOptions.test.ts`, `parseMarkdownBlock.test.ts`, `settings.spec.ts`) are unrelated and unchanged by this PR.

## Notes for reviewers

- I left the codex mapper alone — it still emits content as a text envelope. A follow-up could converge it on the same `output` field if desired, but I didn't want to bundle a behavior change for codex into a Claude bug fix.
- The "frozen" comment at the top of `packages/happy-wire/src/sessionProtocol.ts` looks stale relative to actual usage (`sendClaudeSessionMessage` is wholly on the envelope path); happy to drop or update it in a separate PR if you'd like.

### DIFF
--- a/packages/happy-app/sources/sync/typesRaw.spec.ts
+++ b/packages/happy-app/sources/sync/typesRaw.spec.ts
@@ -1706,6 +1706,36 @@ describe('Zod Transform - WOLOG Content Normalization', () => {
             }
         });
 
+        it('preserves tool-call-end output and isError on the normalized tool-result', () => {
+            const normalized = normalizeRawMessage('db-output', null, 1, {
+                ...base,
+                content: {
+                    type: 'session',
+                    data: {
+                        id: 'env-output',
+                        time: 1,
+                        role: 'agent',
+                        turn: 'turn-1',
+                        ev: {
+                            t: 'tool-call-end',
+                            call: 'call-output',
+                            output: 'src/auth/index.ts',
+                            isError: true
+                        }
+                    }
+                }
+            });
+            expect(normalized).toBeTruthy();
+            if (normalized && normalized.role === 'agent') {
+                expect(normalized.content[0]).toMatchObject({
+                    type: 'tool-result',
+                    tool_use_id: 'call-output',
+                    content: 'src/auth/index.ts',
+                    is_error: true
+                });
+            }
+        });
+
         it('maps turn-end to ready event and drops turn-start', () => {
             const turnStart = normalizeRawMessage('db-5', null, 1, {
                 ...base,
--- a/packages/happy-app/sources/sync/typesRaw.ts
+++ b/packages/happy-app/sources/sync/typesRaw.ts
@@ -54,6 +54,8 @@ const sessionToolCallStartEventSchema = z.object({
 const sessionToolCallEndEventSchema = z.object({
     t: z.literal('tool-call-end'),
     call: z.string(),
+    output: z.string().optional(),
+    isError: z.boolean().optional(),
 });
 
 const sessionFileEventSchema = z.object({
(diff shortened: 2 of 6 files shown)


Jecvay:
+1 — confirming this reproduces on CLI 1.1.10 (latest, `cli-1.1.10`) with remote/mobile sessions. `Bash`, `Read`, `Grep`, `Edit` all render empty bodies ("tool produced no output") in the app, while the transcript clearly contains the `tool_result` content — so the data is intact end-to-end, only the app rendering is broken.

This makes the mobile app basically unusable for monitoring real Claude Code work, since every single tool call looks empty. Would really appreciate a review/merge here — the PR has been open since May 5 with zero feedback. Thanks @WEIFENG2333 for the thorough root-cause writeup and the codex-mapper comparison.

PR #1242, opened 2026-05-08
perf(messages): lazy-load history + parallel AES decrypt + backward pagination

buallen:
## Problem

Opening a long session (thousands of messages) feels frozen. The chat list stays blank for many seconds — sometimes minutes on a slow link — while the app fetches the entire history from `seq=0` forward. The user cannot start typing until every page has been fetched, decrypted, normalised, and reduced.

## Root cause

Three reinforcing issues that together turn message sync into the gating step for opening a session:

### 1. Forward-only pagination on the server

`GET /v3/sessions/:id/messages` only supports `after_seq=N` (ASC) — no way to ask the server for the latest page first. So the client paginates forward from `seq=0` in 100-message slices and only renders once the very last page arrives.

### 2. Sequential AES-GCM decryption on the client

`AES256Encryption.decrypt` awaits each item in a `for await` loop:

```ts
async decrypt(data: Uint8Array[]) {
    const results = [];
    for (const item of data) {
        results.push(await decryptAESGCMString(...)); // sequential!
    }
    return results;
}
```

For a 1 000-message session that's ~1 000 strictly serial AES-GCM calls before the UI can render anything. Web Crypto / the JSI crypto bridge can interleave fine — we were just not letting them.

### 3. No way to fetch older history on demand

Even if the client wanted to load only the latest page first, there was no backward-paging API to fall back on when the user scrolled up.

## Impact

- Multi-second to multi-minute blank screens when opening a long session over a slow network.
- Same blank window again on reconnect, because the forward sweep restarts from the persisted `lastSeq`.
- The CPU spike from sequential decryption is itself enough to keep the message list blank on lower-end phones even on fast networks.

### Server (`packages/happy-server`)

- `v3SessionRoutes.ts`: add an opt-in `before_seq` query parameter that returns messages with `seq < before_seq`, ordered DESC. Mutually exclusive with `after_seq`. **Backward compatible** — with neither parameter, behaviour is unchanged (forward from `seq=0` ASC), so existing clients keep working unchanged.
- `v3SessionRoutes.test.ts`: extend the in-memory mock to honour `seq.lt` and DESC ordering. New coverage:
  - backward pagination returns newest-first and respects `hasMore`,
  - the route rejects requests that combine `after_seq` and `before_seq`.

### Client (`packages/happy-app`)

- `sync/encryption/encryptor.ts`: `AES256Encryption.decrypt` now uses `Promise.all` so AES-GCM calls interleave on the microtask queue.
- `sync/sync.ts`: split `fetchMessages` into two paths:
  - **Initial load** — fetch a single latest page via `before_seq=2_147_483_647` (the Postgres `int4` upper bound, so Prisma will not coerce or overflow). Anchor both `sessionLastSeq` (forward cursor) and a new `sessionOldestSeq` (backward cursor) and unblock the UI immediately.
  - **Incremental forward sync** — used on every subsequent visit / reconnect; only pulls messages newer than what is already in memory.
  - Add public `loadOlderMessages(sessionId)` for the chat list to call when the user scrolls past the visual top.
  - Kick off a fire-and-forget background prefetch loop after the initial latest page so older history streams in **even when the user does not scroll**. The loop yields between pages to keep the UI thread responsive and to spread out server load.
- `sync/storage.ts`: extend `SessionMessages` with `hasMoreOlder` and `isLoadingOlder`. Add `applyOlderMessagesPagination` / `applyOlderMessagesLoading` mutators. Expose the new fields through `useSessionMessages`.
- `components/ChatList.tsx`: wire FlatList `onEndReached` (which fires at the visual top in `inverted` mode) to `sync.loadOlderMessages`, and render an `ActivityIndicator` footer above the oldest message while a backward page is in flight.

## Risk

- **Reducer ordering**: the reducer is idempotent and Map-keyed by message id, so feeding pages out of strict chronological order is already supported (live socket pushes already arrive interleaved with fetch responses). No behaviour change for the existing message-application path.
- **Server route compatibility**: old clients that send `after_seq` — or send neither — keep their existing semantics. Only new clients opt into `before_seq`. The Zod schema now refuses requests that combine both params; this is a stricter check, but no existing client sends both.
- **Sentinel value**: `2_147_483_647` is the `int4` upper bound, so Prisma neither coerces nor overflows when using it as `before_seq`.
- **Lock semantics**: `loadOlderMessages` and the background prefetch share the same per-session `AsyncLock` as `fetchMessages` and `applyMessages`, so they cannot race with live socket updates or duplicate themselves.

## Test plan

- [ ] Open a session with a few thousand messages on a slow network — chat input should be usable within ~1 RTT, not after the full history downloads.
- [ ] Scroll up — older pages stream in with a spinner; no flicker, no duplicates.
- [ ] Leave the session foregrounded without scrolling — older pages still fill in via the background prefetch loop.
- [ ] Reconnect / re-invalidate — only forward sync runs; no full re-download.
- [ ] Send a new message while the background prefetch is mid-flight — no race; the new message lands at the top.
- [ ] Server tests: `pnpm --filter happy-server test sources/app/api/routes/v3SessionRoutes.test.ts` (10/10 pass, including the two new cases).
- [ ] Client typecheck: `pnpm --filter happy-app typecheck` (clean).

### DIFF
--- a/packages/happy-app/sources/components/ChatList.tsx
+++ b/packages/happy-app/sources/components/ChatList.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { useSession, useSessionMessages } from "@/sync/storage";
-import { FlatList, NativeScrollEvent, NativeSyntheticEvent, Platform, Pressable, View } from 'react-native';
+import { sync } from '@/sync/sync';
+import { ActivityIndicator, FlatList, NativeScrollEvent, NativeSyntheticEvent, Platform, Pressable, View } from 'react-native';
 import { useCallback } from 'react';
 import { useHeaderHeight } from '@/utils/responsive';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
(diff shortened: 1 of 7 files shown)


dreamer2q:
In urgent need of this, the memory happy web occupied is very high due to this.

bra1nDump:
Thank you! Super important bug fix!

buallen:
@bra1nDump You are welcome!!!  You guys have done a great project which makes me WLB.  And it's my honor to have a chance to make it better!

Issue #652, opened 2026-02-16
Context limit reached is a dead-end in the app - no way to compact/continue


omachala:
## Issue
When a session reaches the context limit in the mobile/web app, there is no way to recover or continue. The user must manually switch to the terminal to handle context compaction, which defeats the purpose of mobile access.

## Current Behavior
1. Session runs in the app and reaches context limit
2. Session becomes stuck/unusable
3. No option in the app to compact context or continue
4. User must:
   - Switch to their computer
   - Open terminal
   - Find and connect to the session
   - Manually handle the context limit situation

## Why This Is Problematic
- **Defeats mobile use case**: If I have to go to my terminal anyway, why use the mobile app?
- **Common occurrence**: Context limits are hit frequently during longer sessions
- **No self-service recovery**: Users are stuck and can't resolve the issue from where they are
- **Poor UX**: Dead-end error state with no actionable next steps

## Expected Behavior
When context limit is reached, the app should provide options to:

1. **Automatic context compaction**
   - "Compact and continue" button
   - Let Claude summarize earlier parts of the conversation
   - Resume with compacted context

2. **Manual context management**
   - View context usage/remaining tokens
   - Choose which parts of conversation to keep/remove
   - "Start fresh" option while keeping session context

3. **Graceful continuation options**
   - "Continue in new session" (with summary of current work)
   - "Save conversation and restart"
   - Clear error message explaining the situation and available actions

### Minimal Fix
Add a "Compact Context" button that:
- Triggers automatic context compaction via API
- Shows progress/status
- Resumes session when ready

### Better Fix
- Show context usage indicator before hitting limit (e.g., "85% of context used")
- Warning when approaching limit with proactive compaction offer
- Smart compaction that preserves important context
- Visual feedback during compaction process

### Ideal Fix
- Automatic context management (no user intervention needed)
- Seamless continuation without interrupting flow
- Context usage visible in session info
- User-configurable auto-compact threshold

## Impact
**High** - This is a critical mobile UX failure. Long-running sessions from mobile become unusable when they hit context limits, forcing users to abandon the mobile workflow entirely.

## Related Issues
- Similar to #504 (sync issues) - requires round-tripping to terminal to fix
- Part of larger pattern where app lacks parity with CLI functionality

## Technical Notes
Claude API supports context compaction via summarization. The app should be able to trigger this remotely without requiring terminal access.

## User Story
[...]

PR #1254, opened 2026-05-11
fix: relay usage data to app via session protocol

### LINKED ISSUE #1126: Show plan session usage + reset time in mobile status bar (Claude + Codex)
**Summary**                                                                       

  Add a status-bar indicator on the mobile app (and web) showing the active     
  vendor's plan session usage — % remaining and reset time — so it's visible at
  a glance without opening the CLI.                                             
                                                                  
  Should support both vendors Happy already wraps:                              
  - Claude (Pro / Max) — 5-hour session window + weekly cap
  - Codex (ChatGPT Plus / Pro / Business) — equivalent quota / reset window     
                                                                  
  **Why**                                                                           
                                                                  
  Happy's strength is driving long-running coding sessions from the phone, often
   across the full 5-hour window. Without visibility into how much of the
  session budget is left, it's easy to hit the wall mid-task or mis-schedule    
  work. Desktop wrappers like Mission Control / Project Studio already surface
  this in a status bar, and having it on mobile would reach parity with the
  desktop experience Happy is competing with.

  **Data source**

  - Claude: claude /cost and /context already return usage + reset timestamps;  
  the Claude Agent SDK exposes the same via response metadata. Happy wraps the
  CLI so it has access to both.                                                 
  - Codex: equivalent data available from the Codex CLI / ChatGPT backend. Less
  familiar to me — maintainers would know the right surface.                    
   
  **Suggested UX**                                                                  
                                                                  
  - Compact pill in the session header: ◐ 42% · resets 19:30                    
  - Tap to expand: session window % + weekly cap % + explicit reset timestamps.
  - Icon/color shifts (green → amber → red) as the session window runs low, and 
  a warning when < 10% or < 15 min remains.                                     
  - Hide automatically when the session is tied to API-key auth rather than a   
  plan (no quota to show).                                                      
                                                                  
  **Scope**                                                                         
                                                                  
  Mobile (Android + iOS) and web clients. Not asking for any changes to the CLI 
  side.

EricSeastrand:
The UI for displaying context window usage already exists — `AgentInput`'s `getContextWarning()`, `processUsageData()` in the reducer, the `alwaysShowContextSize` setting, and all the translations were added in 3095a2b9, 9908e992, and 80035e24 (Aug–Sep 2025). But the data never reaches them: usage token counts are sent to the server via `usage-report` socket events for billing, but are not relayed back to the app through the session protocol. The result is that `latestUsage` is always empty and the context percentage display never appears.

This commit adds the missing plumbing:
- **Wire schema**: new `usage` event type in the session protocol discriminated union (`happy-wire`)
- **CLI**: emit a usage envelope after each assistant message, alongside the existing `usage-report` (`happy-cli`)
- **App**: handle the usage envelope in the message normalizer so the reducer's `processUsageData()` actually receives data (`happy-app`)

No new UI, no database changes, no server changes.

Closes #1126
Refs #652

### DIFF
--- a/packages/happy-app/sources/sync/typesRaw.ts
+++ b/packages/happy-app/sources/sync/typesRaw.ts
@@ -89,6 +89,14 @@ const sessionStopEventSchema = z.object({
     t: z.literal('stop'),
 });
 
+const sessionUsageEventSchema = z.object({
+    t: z.literal('usage'),
+    input_tokens: z.number(),
+    output_tokens: z.number(),
+    cache_creation_input_tokens: z.number().optional(),
+    cache_read_input_tokens: z.number().optional(),
+});
+
 const sessionEventSchema = z.discriminatedUnion('t', [
     sessionTextEventSchema,
     sessionServiceMessageEventSchema,
@@ -99,6 +107,7 @@ const sessionEventSchema = z.discriminatedUnion('t', [
     sessionStartEventSchema,
     sessionTurnEndEventSchema,
     sessionStopEventSchema,
+    sessionUsageEventSchema,
 ]);
 
 const sessionEnvelopeSchema = z.object({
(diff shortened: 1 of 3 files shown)


EricSeastrand:
Here is a screenshot of it working on iOS, web-based (not native; I only use web).
See the “59%” at bottom of screen.
IMG_2537

PR #1271, opened 2026-05-14
fix(app): scale MAX_CONTEXT_SIZE for 1M models + render alwaysShow before usage streams in

### LINKED ISSUE #910: Context remaining shows 0% on 1M context models (hardcoded MAX_CONTEXT_SIZE = 190000)
## Problem

The context remaining percentage in the session input bar always shows **0%** when using Claude Opus 4.6 with 1M context window. The actual context usage is only ~20%, but the UI clamps it to 0%.

## Root Cause

`MAX_CONTEXT_SIZE` is hardcoded to `190000` in `AgentInput.tsx`, which is appropriate for 200K context models (Sonnet) but not for 1M context models (Opus 4.6 `[1m]`).

### Code pointers

1. **`packages/happy-app/sources/sync/reducer/reducer.ts:1115`** — `contextSize` is calculated correctly:
   ```ts
   contextSize: (usage.cache_creation_input_tokens || 0)
              + (usage.cache_read_input_tokens || 0)
              + usage.input_tokens
   ```

2. **`packages/happy-app/sources/components/AgentInput.tsx:78, 286-287`** — percentage uses hardcoded limit:
   ```ts
   const MAX_CONTEXT_SIZE = 190000;
   const percentageUsed = (contextSize / MAX_CONTEXT_SIZE) * 100;
   const percentageRemaining = Math.max(0, Math.min(100, 100 - percentageUsed));
   ```

### Example

With Opus 4.6 (1M context), a session at ~205K context:
- `percentageUsed = 205000 / 190000 * 100 = 107.8%`
- `percentageRemaining = Math.max(0, 100 - 107.8) = 0%`
- Actual remaining: ~79.5% (205K / 1M)

## Suggested Fix

Make `MAX_CONTEXT_SIZE` dynamic based on the model in use. The CLI's `usage-report` doesn't currently include model info, so the model name may need to be passed through or inferred.

Known context window sizes:
- Claude Sonnet / Haiku: 200K → ~190K effective
- Claude Opus 4.6 `[1m]`: 1M → ~950K effective

## Impact

- UI-only: does not affect functionality or billing
- Users on 1M context models see 0% for entire session, losing visibility into actual context usage

### LINKED ISSUE #1274: alwaysShowContextSize toggle has no visible effect on Android
## Summary

Enabling the **Always Show Context Size** toggle in Settings → Appearance on
Android has no visible effect — no context-size indicator appears in the
agent input, even after multiple agent replies have landed in the session.

## Repro

1. Android app, latest from Play Store
2. Settings → Appearance → enable "Always Show Context Size"
3. Open or start any session (Claude backend, Opus 4.7 `[1m]`)
4. Send a prompt, wait for the agent to reply
5. Observed: no "% remaining" text anywhere in the input area
6. Expected: yellow "% remaining" label, per `getContextWarning` in
   `packages/happy-app/sources/components/AgentInput.tsx`

## Environment

- happy CLI: `happy-coder@1.1.9`
- Model: `claude-opus-4-7[1m]` (1M context)
- Session usage observed in CLI logs at time of test: ~30,715 tokens
  (~16% of the hardcoded 190K limit → would expect "84% remaining" in yellow
  when `alwaysShow` is true)
- CLI is emitting `[SOCKET] Sending usage data` every turn, so usage data
  is flowing to the server

## What I ruled out

- ✅ Setting key exists and is wired through SessionView →
  AgentInput (`SessionView.tsx:482, 697`)
- ✅ `getContextWarning` returns a non-null result whenever
  `alwaysShow` is true, regardless of percentage
  (`AgentInput.tsx:303`)
- ✅ CLI is forwarding usage end-to-end — `sendUsageData` runs every
  turn (the log-stripping at `types-BtoZF_Bo.mjs:285` was a red herring
  that initially threw me off; the actual data carries `usage`)
- ✅ Old "context warning nested inside connection status" bug
  was fixed in `80035e2` (2025-09-02) and is in `main`

## Possible causes (not yet narrowed)

1. **Stale APK** — latest release tag is `v1.1.6` (2026-04-13); the
   wiring has been touched multiple times since. Possible the Play
   Store build doesn't include the current code path. Can someone
   confirm which release ships on the current Android channel?
2. **Setting save vs reducer-state mismatch** — toggle persists
   locally but doesn't end up in the slice that AgentInput reads.
3. **`latestUsage` not populated for some session types** — would
   want to confirm `usageData?.contextSize` is actually non-zero on
   device when the indicator should appear.

## Related

#910 — hardcoded `MAX_CONTEXT_SIZE = 190000` for 1M-context models.
Related area, but a different symptom (#910 shows 0%, this shows
nothing at all).

g0rdonL:
## Summary

Two related fixes to the **Always Show Context Size** indicator in `AgentInput`. Originally only #910; on rebase I traced #1274 to the same render path and added a one-line gate fix that pairs naturally with the model plumbing.

### Commit 1 — Closes #910 (scale MAX_CONTEXT_SIZE for 1M-context models)
- Context-remaining indicator was hardcoded to **190K**, so any session on a 1M-context Anthropic model (e.g. `claude-opus-4-7[1m]`) clamped to 0% within minutes and the toggle appeared broken.
- Plumb the assistant `model` field from the raw Claude API event through `NormalizedMessage` → reducer `latestUsage` → `Session.latestUsage` → `AgentInput.usageData`.
- `AgentInput` now picks the limit dynamically via `getMaxContextSize(model)`: **950K** for model ids matching `[1m]`, **190K** otherwise.

### Commit 2 — Closes #1274 (render alwaysShow before usage streams in)
- The outer gate at the call site short-circuited to `null` whenever `usageData.contextSize` was `0` or `undefined`:
  ```ts
  const contextWarning = props.usageData?.contextSize
      ? getContextWarning(...)
      : null;
  ```
  …which swallowed the alwaysShow signal that `getContextWarning` itself honors (line 311). Two failure modes this addresses on Android (and silently on iOS/web — same render path):
  1. Fresh session with no agent reply yet — `latestUsage` is undefined, indicator never showed.
  2. Sessions where the reducer resets `contextSize` to 0 on session events (`reducer.ts:324, 338`) with no subsequent assistant message carrying usage.
- Gate is now `alwaysShow || contextSize`, and `getContextWarning` receives `contextSize ?? 0`. On fresh always-show sessions the indicator renders **"100% remaining"** until real usage arrives.

## Files changed
- `sources/sync/typesRaw.ts` — capture `message.model` when normalizing assistant messages; add `model?: string` to `NormalizedMessage`.
- `sources/sync/reducer/reducer.ts` — thread `model` through `processUsageData` / `ReducerState.latestUsage` / `ReducerResult.usage`.
- `sources/sync/storageTypes.ts` — add `model?: string` to persisted `Session.latestUsage`.
- `sources/-session/SessionView.tsx` — thread `model` into the memoized `usageData` source mapping.
- `sources/components/AgentInput.tsx` — `getMaxContextSize(model)` helper + rewrite the `contextWarning` gate to honor `alwaysShow` when usage is missing.

## Test plan
- [x] `pnpm typecheck` in `packages/happy-app` passes.
- [x] `pnpm test --run` in `packages/happy-app` — **identical** failure baseline to `main` (11 failed / 466 passed / 57 skipped on both branches; failures are pre-existing markdown, GitHub API, settings-defaults, and model-mode tests untouched by this PR).
- [ ] Manual: open a session on `claude-opus-4-7[1m]` with **Always Show Context Size** enabled; the percentage should reflect 1M-token usage rather than sticking at 0%.
- [ ] Manual: open a 200K-context session (Sonnet/Haiku); behavior unchanged.
- [ ] Manual: enable **Always Show Context Size** on a fresh session (no agent reply yet) — the indicator should render "100% remaining" immediately rather than waiting for usage.

## Notes
- `[1m]` detection is `/\[1m\]/i` against the model id returned by the Claude API stream. Happy to switch to an explicit lookup table if you'd prefer.
- `model` is optional throughout — pre-existing sessions and non-Claude flavors keep working with the 190K default.
- Rebased onto `main` (270+ commits since the original PR); `SessionView.tsx` was reconciled with the new memoized `usageData` introduced in main.

Closes #910
Closes #1274

### DIFF
--- a/packages/happy-app/sources/-session/SessionView.tsx
+++ b/packages/happy-app/sources/-session/SessionView.tsx
@@ -571,6 +571,7 @@ function SessionViewLoaded({ sessionId, session }: { sessionId: string, session:
             cacheCreation: source.cacheCreation,
             cacheRead: source.cacheRead,
             contextSize: source.contextSize,
+            model: source.model,
         };
     }, [sessionUsage, session.latestUsage]);
 
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -72,6 +72,7 @@ interface AgentInputProps {
         cacheCreation: number;
         cacheRead: number;
         contextSize: number;
+        model?: string;
     };
     alwaysShowContextSize?: boolean;
     onFileViewerPress?: () => void;
@@ -93,7 +94,18 @@ interface AgentInputProps {
     onAddImages?: (images: AttachmentPreview[]) => void;
 }
 
-const MAX_CONTEXT_SIZE = 190000;
+// Effective usable context per model. Anthropic 1M-context variants
+// expose ~950K usable tokens; everything else is treated as 200K (~190K usable).
+// See https://github.com/slopus/happy/issues/910
+const DEFAULT_MAX_CONTEXT_SIZE = 190000;
+const ONE_MILLION_MAX_CONTEXT_SIZE = 950000;
+
+function getMaxContextSize(model?: string): number {
+    if (model && /\[1m\]/i.test(model)) {
+        return ONE_MILLION_MAX_CONTEXT_SIZE;
+    }
+    return DEFAULT_MAX_CONTEXT_SIZE;
+}
 
 const stylesheet = StyleSheet.create((theme, runtime) => ({
     container: {
@@ -300,8 +312,9 @@ const stylesheet = StyleSheet.create((theme, runtime) => ({
     },
 }));
 
-const getContextWarning = (contextSize: number, alwaysShow: boolean = false, theme: Theme) => {
-    const percentageUsed = (contextSize / MAX_CONTEXT_SIZE) * 100;
+const getContextWarning = (contextSize: number, alwaysShow: boolean = false, theme: Theme, model?: string) => {
+    const maxContextSize = getMaxContextSize(model);
+    const percentageUsed = (contextSize / maxContextSize) * 100;
     const percentageRemaining = Math.max(0, Math.min(100, 100 - percentageUsed));
 
     if (percentageRemaining <= 5) {
@@ -598,7 +611,7 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
 
     // Calculate context warning
     const contextWarning = props.usageData?.contextSize
-        ? getContextWarning(props.usageData.contextSize, props.alwaysShowContextSize ?? false, theme)
+        ? getContextWarning(props.usageData.contextSize, props.alwaysShowContextSize ?? false, theme, props.usageData.model)
         : null;
 
     const agentInputEnterToSend = useSetting('agentInputEnterToSend');
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -609,9 +609,10 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
         return label;
     }, [isSandboxEnabled]);
 
-    // Calculate context warning
-    const contextWarning = props.usageData?.contextSize
-        ? getContextWarning(props.usageData.contextSize, props.alwaysShowContextSize ?? false, theme, props.usageData.model)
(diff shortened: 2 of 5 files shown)


g0rdonL:
Closing — no longer needed.

PR #1284, opened 2026-05-18
feat(acp): decode tool_call.content subtypes and read args from rawInput

YOMXXX:
## Summary
Two related ACP tool-call improvements inside \`packages/happy-cli/src/agent/acp/sessionUpdateHandlers.ts\`, both grounded in the ACP spec (SDK v0.14.1).

### 1. Read tool args from \`rawInput\` (spec-correct), fall back to \`content\` (legacy)
Per ACP spec (\`ToolCall\` and \`ToolCallUpdate\`):
- \`rawInput?: unknown\` — Raw input parameters sent to the tool.
- \`content?: Array<ToolCallContent>\` — Content **produced by** the tool call.

The previous code in \`startToolCall\` read args from \`update.content\`, which mistakes the output collection for the input parameters. \`startToolCall\` now goes through a new \`parseToolArgs\` helper:

\`\`\`ts
export function parseToolArgs(update: SessionUpdate): Record<string, unknown> {
  const rawInput = (update as { rawInput?: unknown }).rawInput;
  if (rawInput !== undefined && rawInput !== null) {
    if (typeof rawInput === 'object' && !Array.isArray(rawInput)) {
      return rawInput as Record<string, unknown>;
    }
    return { value: rawInput };
  }
  return parseArgsFromContent(update.content);
}
\`\`\`

\`rawInput\` wins when present; otherwise we fall back to the existing \`parseArgsFromContent\` shape so agents that pre-date \`rawInput\` keep working — fully backwards-compatible.

### 2. Decode \`ToolCallContent\` subtypes into structured AgentMessages
Per ACP spec, each entry in \`ToolCall.content\` is tagged:

\`\`\`ts
type ToolCallContent =
  | (Content  & { type: 'content' })   // text / image / audio / resource_link / resource
  | (Diff     & { type: 'diff' })       // { path, oldText?, newText }
  | (Terminal & { type: 'terminal' });  // { terminalId }
\`\`\`

The previous code passed the entire array through verbatim as \`tool-result.result\`, so diffs and terminal references reached downstream as opaque objects. \`completeToolCall\` now keeps emitting the original \`tool-result\` (consumers reading \`result\` keep working) and **additionally** walks the array via a new \`emitToolContentItems\`:

| ACP subtype | Extra AgentMessage emitted |
|---|---|
| \`diff\` | \`{ type: 'fs-edit', description, diff, path }\` — picked up by \`runAcp.ts\`'s fs-edit switch case and \`GeminiDiffProcessor\` |
| \`terminal\` | \`{ type: 'event', name: 'tool_terminal_ref', payload: { toolCallId, toolName, terminalId } }\` |
| \`content\` | (no extra emit; the \`ContentBlock\` stays inside the \`tool-result.result\` passthrough) |

The new \`formatToolCallDiff\` helper renders a minimal unified-diff-style string for the \`fs-edit.diff\` field. \`GeminiDiffProcessor\` treats this as an opaque blob and only compares it across edits, so a real \`diff(1)\` patch is not needed.

## Scope
- Touches **only** \`packages/happy-cli/src/agent/acp/sessionUpdateHandlers.ts\` and adds a colocated test file.
- Does NOT change \`AcpBackend.ts\`, \`runAcp.ts\`, \`AcpSessionManager.ts\`, \`apiSession.ts\`, the mobile app, or any socket schema. The \`fs-edit\` AgentMessage already has a downstream consumer (\`runAcp.ts:162\` + \`GeminiDiffProcessor\`); the \`tool_terminal_ref\` event is new and is a signal-only addition — downstream UI rendering can follow up.

## Test plan
- [x] \`pnpm --filter happy typecheck\` — passes, no new TypeScript errors
- [x] \`pnpm exec vitest run src/agent/acp/\` — 63 tests pass (incl. 15 new tests in \`sessionUpdateHandlers.toolContent.test.ts\`)

The 15 new tests cover:
- \`parseToolArgs\`: \`rawInput\` priority (object form), value-wrapping for non-object \`rawInput\`, content fallback for legacy agents, array-wrapped \`{ items }\` fallback, empty-update case.
- \`formatToolCallDiff\`: header presence, line prefixes, empty-text handling, no-path branch.
- \`emitToolContentItems\`: one \`fs-edit\` per \`diff\` entry, \`tool_terminal_ref\` per terminal, no-op for plain \`content\`, mixed scenarios, no-throw for non-array input, defensive parsing for missing string fields.

## Spec references
- \`ToolCall\` and \`ToolCallUpdate\` field semantics: \`@agentclientprotocol/sdk@0.14.1\` \`dist/schema/types.gen.d.ts\` lines 2532–2577 and 2641–2700.
- \`ToolCallContent\` union: same file, lines 2578–2592.
- \`Diff\` / \`Terminal\` shapes: lines 525–548 and 2403–2415.

### DIFF
--- a/packages/happy-cli/src/agent/acp/sessionUpdateHandlers.toolContent.test.ts
+++ b/packages/happy-cli/src/agent/acp/sessionUpdateHandlers.toolContent.test.ts
@@ -1,0 +1,226 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import type { AgentMessage } from '../core';
+import {
+  emitToolContentItems,
+  formatToolCallDiff,
+  parseToolArgs,
+  type HandlerContext,
+  type SessionUpdate,
+} from './sessionUpdateHandlers';
+
+function makeCtx(): { ctx: HandlerContext; messages: AgentMessage[] } {
+  const messages: AgentMessage[] = [];
+  const ctx: HandlerContext = {
+    transport: {} as HandlerContext['transport'],
+    activeToolCalls: new Set<string>(),
+    toolCallStartTimes: new Map<string, number>(),
+    toolCallTimeouts: new Map<string, NodeJS.Timeout>(),
+    toolCallIdToNameMap: new Map<string, string>(),
+    idleTimeout: null,
+    toolCallCountSincePrompt: 0,
+    emit: (msg) => {
+      messages.push(msg);
+    },
+    emitIdleStatus: vi.fn(),
+    clearIdleTimeout: vi.fn(),
+    setIdleTimeout: vi.fn(),
+  };
+  return { ctx, messages };
+}
+
+describe('parseToolArgs', () => {
+  it('uses rawInput when present (object form)', () => {
+    const update: SessionUpdate = {
+      sessionUpdate: 'tool_call',
+      toolCallId: 't1',
+      rawInput: { path: '/tmp/foo', limit: 5 },
+    } as SessionUpdate;
+
+    expect(parseToolArgs(update)).toEqual({ path: '/tmp/foo', limit: 5 });
+  });
+
+  it('wraps a non-object rawInput in { value }', () => {
+    const update: SessionUpdate = {
+      sessionUpdate: 'tool_call',
+      toolCallId: 't1',
+      rawInput: 'just a string',
+    } as SessionUpdate;
+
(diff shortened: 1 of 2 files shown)


YOMXXX:
@codex review

PR #1286, opened 2026-05-18
fix(acp): defer idle after tool completion (UI flicker / stuck spinner)

YOMXXX:
## Summary
When the last tool in a turn completes (or fails), \`completeToolCall\` and \`failToolCall\` previously emitted \`{ type: 'status', status: 'idle' }\` **synchronously**. But agents almost always stream a short follow-up text chunk right after a tool result (e.g. \"tool used: result is 42\"), and that chunk arrives in the *next* tick.

The synchronous \`idle\` emit causes downstream UIs to see \`idle → busy → idle\` flicker — or, in some third-party clients (e.g. atomcode), get stuck on a stale idle view while later chunks arrive on what looks like a closed turn.

\`handleAgentMessageChunk\` already deferred \`idle\` through a \`setIdleTimeout(DEFAULT_IDLE_TIMEOUT_MS)\` (500 ms) window so a streaming chunk wouldn't trip the same flicker. Tool completion / failure now goes through the same deferral via a new \`scheduleDeferredIdle\` helper, so both event types converge on **one** idle-settling rule.

## Diagnosis
| Path | Before | After |
|---|---|---|
| \`handleAgentMessageChunk\` | \`setIdleTimeout(callback, 500ms)\` | unchanged |
| \`completeToolCall\` (last active) | \`emitIdleStatus()\` **immediately** | \`scheduleDeferredIdle()\` (defers 500 ms) |
| \`failToolCall\` (last active) | \`emitIdleStatus()\` **immediately** | \`scheduleDeferredIdle()\` (defers 500 ms) |
| \`startToolCall\`'s 120 s safety net | \`emitIdleStatus()\` **immediately** | unchanged — already waited the full timeout |

The discrepancy between chunks (deferred) and tool completions (synchronous) was the race that downstream UIs were tripping over.

## Behavior after this PR
- Tool completes (last active) → schedule deferred idle in \`DEFAULT_IDLE_TIMEOUT_MS\`.
- Follow-up text chunk arrives within the window → chunk handler clears and re-arms the timer; no extra \`idle\` is emitted.
- New tool starts within the window → \`scheduleDeferredIdle\` skips the emit because \`activeToolCalls.size > 0\` at fire time.
- Multiple tools finish back-to-back → only the last completion fires the deferred timer; one final \`idle\`.

## Scope
- Touches **only** \`packages/happy-cli/src/agent/acp/sessionUpdateHandlers.ts\` and adds a colocated test file.
- Does NOT change \`AcpBackend.ts\`, \`runAcp.ts\`, \`AcpSessionManager.ts\`, the mobile app, or any socket schema.
- The tool-call-timeout safety net path is left as immediate-emit because it has already waited the full 120 s timeout.

## Test plan
- [x] \`pnpm exec vitest run src/agent/acp/\` — all 55 ACP tests pass (incl. 7 new fake-timer tests)

The new tests use \`vi.useFakeTimers()\` to deterministically reproduce the original race and pin the new behavior:
1. \`completeToolCall\` does NOT emit \`idle\` synchronously.
2. \`idle\` fires exactly when \`DEFAULT_IDLE_TIMEOUT_MS\` elapses.
3. **Race fix:** a follow-up text chunk inside the deferral window re-arms the timer; only the chunk-handler's later deadline emits, never the original tool-deadline.
4. A new tool starting inside the window cancels the deferred idle (still busy).
5. \`failToolCall\` defers idle the same way (mirrors \`completeToolCall\`).
6. \`scheduleDeferredIdle\` skips emit if a tool became active during the wait.
7. Multiple back-to-back completions collapse into a single \`idle\`.

## Downstream impact
Third-party ACP clients (e.g. atomcode) previously worked around this by sending an empty text chunk after every tool result, just to re-arm happy-cli's idle timer. That workaround can be removed once this lands.

### DIFF
--- a/packages/happy-cli/src/agent/acp/sessionUpdateHandlers.idleDeferral.test.ts
+++ b/packages/happy-cli/src/agent/acp/sessionUpdateHandlers.idleDeferral.test.ts
@@ -1,0 +1,182 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { AgentMessage } from '../core';
+import {
+  DEFAULT_IDLE_TIMEOUT_MS,
+  completeToolCall,
+  failToolCall,
+  handleAgentMessageChunk,
+  scheduleDeferredIdle,
+  startToolCall,
+  type HandlerContext,
+  type SessionUpdate,
+} from './sessionUpdateHandlers';
+
+function makeCtx() {
+  const messages: AgentMessage[] = [];
+  let idleTimeoutHandle: NodeJS.Timeout | null = null;
+
+  const ctx: HandlerContext = {
+    transport: {
+      getIdleTimeout: () => DEFAULT_IDLE_TIMEOUT_MS,
+    } as HandlerContext['transport'],
+    activeToolCalls: new Set<string>(),
+    toolCallStartTimes: new Map<string, number>(),
+    toolCallTimeouts: new Map<string, NodeJS.Timeout>(),
+    toolCallIdToNameMap: new Map<string, string>(),
+    idleTimeout: null,
+    toolCallCountSincePrompt: 0,
+    emit: (msg) => {
+      messages.push(msg);
+    },
+    emitIdleStatus: () => {
+      messages.push({ type: 'status', status: 'idle' });
+    },
+    clearIdleTimeout: () => {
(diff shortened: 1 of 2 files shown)


YOMXXX:
@codex review

PR #1287, opened 2026-05-18
fix(acp): two-signal turn-complete state machine (response + updates)

YOMXXX:
## Summary

\`AcpBackend.sendPrompt\` previously relied on a **single** completion signal — the update-side \`idleResolver\` — to detect when an ACP turn was done. The RPC response from \`connection.prompt(...)\` was simply awaited for error propagation but did **not** participate in turn-complete detection.

Per ACP spec, agents are free to send the \`prompt\` response **before, during, or after** their final \`session/update\`, so the single-signal design races. This PR introduces a two-signal state machine so the response and the update stream are tracked independently and the turn finishes only when **both** have settled.

> **Stacking note:** this PR is built on top of #1286 (defer idle after tool completion). If #1286 has not yet merged, the diff here will look larger than just the state-machine changes; rebasing on \`main\` post-#1286 leaves only this PR's changes. The two PRs touch the same idle code path, so the recommended merge order is **#1286 first**, then this one.

## Diagnosed races

| Scenario | Old behavior | After this PR |
|---|---|---|
| Updates first, response last (spec-typical Gemini) | Works by luck — \`idleResolver\` fires from the chunk idle timer | Same observable outcome, but now also gated on \`responseReceived\` |
| Response first, then late chunks (Codex-ish) | Synchronous \`idle\` while chunks are still streaming — UI sees a closed turn that keeps writing | Grace window holds back \`idle\` until updates settle |
| Empty turn (response with no updates ever) | Deadlocks on \`waitForResponseComplete\` (120s timeout) | \`armPostResponseGrace()\` settles the turn after \`DEFAULT_IDLE_TIMEOUT_MS\` (500 ms) |
| \`tool_call_update\` completes before response | Deferred idle fires before response is back, turn flips to \`idle\` pending response | Update side settles, but \`idle\` waits for response |
| Two prompts overlap (stale resolution) | First prompt's late \`.then\` mutates current turn state | \`currentTurnId\` discriminator drops stale resolutions |

### \`packages/happy-cli/src/agent/acp/AcpBackend.ts\`
Adds four private fields to track the new state machine:

\`\`\`ts
private responseReceived = false;
private updatesSettled = false;
private currentTurnId = 0;
private postResponseGraceTimer: NodeJS.Timeout | null = null;
\`\`\`

\`sendPrompt\` now bumps \`currentTurnId\`, resets the two booleans, then attaches a \`.then(...)\` to the RPC promise instead of bare-\`await\`ing it. That \`.then\` ignores stale resolutions (where \`turnId !== this.currentTurnId\`).

\`emitIdleStatus()\` is the single internal sink that actually emits \`{ type: 'status', status: 'idle' }\` and resolves \`idleResolver\`. It also clears all turn-complete state so a stale Promise or grace fire can't double-emit. Three new helpers (\`markUpdatesSettled\`, \`maybeCompleteTurn\`, \`armPostResponseGrace\`, \`clearPostResponseGrace\`) sit beside it; \`dispose()\` clears the grace timer.

### \`packages/happy-cli/src/agent/acp/sessionUpdateHandlers.ts\`
\`HandlerContext.emitIdleStatus\` is renamed to \`markUpdatesSettled\`. Its job is now reporting the update-stream's state, not deciding whether to emit \`idle\` — that decision moved to \`AcpBackend.maybeCompleteTurn\`. The three call sites (chunk idle timer, deferred-idle helper, tool-call timeout cleanup) are renamed accordingly.

### Tests
- **New** \`AcpBackend.race.test.ts\` (6 \`vi.useFakeTimers\` scenarios) covering: spec ordering, response-first race, empty turn, tool-complete-before-response, error rejection path, stale-turn ignoring.
- **Adjusted** \`sessionUpdateHandlers.idleDeferral.test.ts\` mock to track \`markUpdatesSettled\` (semantically the right hook for that file's deferral assertions).

## Test plan
- [x] \`pnpm exec vitest run src/agent/acp/\` — 61 tests pass (incl. 6 new race tests, no regressions)

## Risk

- \`status: idle\` event **schema** is unchanged. \`waitForResponseComplete\` keeps its existing contract (resolves on idle, rejects on its own timeout, configurable).
- Event **timing** changes when the agent sends response-first or empty-turn: idle is delayed up to \`DEFAULT_IDLE_TIMEOUT_MS\` (500 ms) — strictly never longer than the existing \`waitForResponseComplete\` 120s ceiling.
- No socket-schema, no mobile-app-side changes.
- Easy rollback: \`maybeCompleteTurn()\` can be reduced to an unconditional \`emitIdleStatus()\` call to restore the old behavior.

## Spec references
- \`session/prompt\` request/response: ACP spec, [Prompt Turn](https://agentclientprotocol.com/protocol/prompt-turn).
- \`session/update\` ordering with the \`prompt\` response is not constrained — both directions are valid per the JSON-RPC interleave rules used by \`@agentclientprotocol/sdk\`.

### DIFF
--- a/packages/happy-cli/src/agent/acp/AcpBackend.race.test.ts
+++ b/packages/happy-cli/src/agent/acp/AcpBackend.race.test.ts
@@ -134,6 +134,28 @@ describe('AcpBackend turn-completion race conditions', () => {
     expect(idleCount()).toBe(1);
   });
 
+  it('content-only tool_call_update does not cancel the post-response grace window', async () => {
+    const backend = createBackend();
+    const response = defer<unknown>();
+    const { idleCount } = stubBackend(backend, response);
+
+    const sendPromise = backend.sendPrompt('test-session', 'hello');
+    await Promise.resolve();
+
+    response.resolve({ stopReason: 'end_turn' });
+    await sendPromise;
+    expect(idleCount()).toBe(0);
+
+    dispatchUpdate(backend, {
+      sessionUpdate: 'tool_call_update',
+      toolCallId: 't1',
+      content: [{ type: 'text', text: 'still streaming' }],
+    });
+
+    await vi.advanceTimersByTimeAsync(DEFAULT_IDLE_TIMEOUT_MS);
+    expect(idleCount()).toBe(1);
+  });
+
   it('race: empty turn — response arrives, NO updates ever — grace timer still settles the turn', async () => {
     const backend = createBackend();
     const response = defer<unknown>();
--- a/packages/happy-cli/src/agent/acp/AcpBackend.race.test.ts
+++ b/packages/happy-cli/src/agent/acp/AcpBackend.race.test.ts
@@ -1,0 +1,227 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { AgentMessage } from '@/agent/core';
+import { DEFAULT_IDLE_TIMEOUT_MS } from './sessionUpdateHandlers';
+import { AcpBackend } from './AcpBackend';
+
+interface DeferredPromise<T> {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+  reject: (reason: unknown) => void;
+}
+
+function defer<T>(): DeferredPromise<T> {
+  let resolve!: (value: T) => void;
+  let reject!: (reason: unknown) => void;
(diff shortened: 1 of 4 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

PR #1291, opened 2026-05-18
fix(acp): harden AcpBackend.dispose lifecycle and drop dead pendingPermissions

YOMXXX:
## Summary
Two related dispose-path issues in \`packages/happy-cli/src/agent/acp/AcpBackend.ts\`:

### 1. Dead \`pendingPermissions\` Map masquerading as a cleanup hook
\`AcpBackend.pendingPermissions\` was declared at line 329 but **never** \`.set()\` anywhere in the codebase — only declared, then \`.clear()\`'d in \`dispose()\`. The actual pending-permission state lives in \`BasePermissionHandler.pendingRequests\` (\`packages/happy-cli/src/utils/BasePermissionHandler.ts:47\`), owned by \`GeminiPermissionHandler\` / \`CodexPermissionHandler\` and used by the synchronous \`requestPermission\` RPC handler at \`AcpBackend.ts:567-720\`.

Because \`dispose()\` never reached into the real pending map, disposing the backend while a permission dialog is open on the mobile app left the \`requestPermission\` SDK handler \`await\`ing a Promise nobody would ever resolve — blocking the ACP SDK's RPC reader thread for the full SIGTERM + SIGKILL grace period (~3 s + 1 s).

**Fix:**
- Delete the unused \`pendingPermissions\` Map and its \`.clear()\` call.
- Call \`this.options.permissionHandler?.reset?.('AcpBackend disposed')\` from \`dispose()\`. \`BasePermissionHandler.reset()\` (lines 189-235) already rejects every pending Promise with \`Error('Session reset')\` — same semantics that \`runAcp.ts:518, 864, 944\` already rely on.
- Surface \`reset?\` on the public \`AcpPermissionHandler\` interface as an **optional** method so both Gemini and Codex handlers wire up automatically; injected custom handlers don't need to change.
- Wrap the \`reset()\` call in try/catch so a misbehaving custom handler can't block the rest of \`dispose()\`.

### 2. \`HandlerContext\` callbacks fire on disposed backends
\`createHandlerContext()\` produced \`emitIdleStatus\`, \`setIdleTimeout\`, and \`clearIdleTimeout\` closures that captured \`this\` (the AcpBackend) but didn't gate on \`this.disposed\`. Tool-call timeouts in \`sessionUpdateHandlers.ts\` use bare \`setTimeout\` (not tracked by \`idleTimeout\`), so a timeout firing within milliseconds of \`dispose()\` — exactly what happens in the Gemini mode-switch flow that disposes the old backend while a tool is in-flight — would call \`emitIdleStatus\` on a dead instance and emit a ghost \`idle\` status that could pollute a freshly-recreated backend.

**Fix:** add \`if (this.disposed) return\` guards to:
- \`emitIdleStatus\` (before delegating to the instance method)
- \`setIdleTimeout\` (both at scheduling time and inside the inner \`setTimeout\` callback before invoking the caller's callback)

\`emit\` already had the disposed guard at line 364, so no change needed there.

## Scope
- Touches **only** \`packages/happy-cli/src/agent/acp/AcpBackend.ts\` plus a new colocated test file.
- No change to \`runAcp.ts\`, \`sessionUpdateHandlers.ts\`, or the permission handlers themselves.
- The \`AcpPermissionHandler\` interface gets a new optional method — fully backwards-compatible with any existing custom implementations.

## Test plan
- [x] \`pnpm exec vitest run src/agent/acp/\` — 53 tests pass (incl. 5 new dispose-safety tests, no regressions)

### New tests cover:
1. \`setIdleTimeout\` scheduled **before** dispose: callback does not fire after \`vi.advanceTimersByTime(500)\`.
2. \`setIdleTimeout\` scheduled **after** dispose: no timer is even armed.
3. \`emitIdleStatus\` after dispose: no \`idle\` AgentMessage is pushed.
4. \`dispose()\` invokes \`permissionHandler.reset()\` exactly once with a disposed-context reason string.
5. \`dispose()\` tolerates \`permissionHandler.reset()\` throwing — \`disposed\` is still set to \`true\`.

## Notes
The Gemini mode-switch flow (\`runGemini.ts:947-979\`) is the most reliable way to hit this race in real usage: it tears down a backend with a tool-call in-flight and immediately spins up a new one. Before this fix, a ghost \`idle\` from the dead backend could land on the new one's listeners and trip turn-state assertions downstream.

### DIFF
--- a/packages/happy-cli/src/agent/acp/AcpBackend.disposeSafety.test.ts
+++ b/packages/happy-cli/src/agent/acp/AcpBackend.disposeSafety.test.ts
@@ -1,0 +1,108 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { AgentMessage } from '@/agent/core';
+import { AcpBackend } from './AcpBackend';
+
+function createBackend(extraOptions: Partial<ConstructorParameters<typeof AcpBackend>[0]> = {}): AcpBackend {
+  return new AcpBackend({
+    agentName: 'test',
+    cwd: '/tmp',
+    command: '/bin/true',
+    ...extraOptions,
+  });
+}
+
+function captureEvents(backend: AcpBackend): AgentMessage[] {
+  const events: AgentMessage[] = [];
+  backend.onMessage((msg) => events.push(msg));
+  return events;
+}
+
+describe('AcpBackend dispose safety', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('createHandlerContext.setIdleTimeout — callback does NOT fire after dispose', async () => {
+    const backend = createBackend();
+    const events = captureEvents(backend);
+    const ctx = (backend as unknown as { createHandlerContext(): { setIdleTimeout(cb: () => void, ms: number): void; emitIdleStatus(): void } }).createHandlerContext();
+
+    const callback = vi.fn();
+    ctx.setIdleTimeout(callback, 500);
+
+    // Dispose before the timer fires (don't await — dispose's process.kill
+    // paths take real time; we just need the synchronous `disposed = true`).
+    void backend.dispose();
+    expect((backend as unknown as { disposed: boolean }).disposed).toBe(true);
+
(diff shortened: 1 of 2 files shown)


YOMXXX:
@codex review

PR #1292, opened 2026-05-18
fix(codex): plug abort+resend race in pendingInterrupt timing

YOMXXX:
## Summary
Smashing the Stop button and immediately submitting a new prompt could leave the new turn matched-and-aborted by the *previous* turn's \`turn_aborted\` notification — no chunks delivered, mobile UI just sat there. The bug is a sub-microtask race inside \`packages/happy-cli/src/codex/codexAppServerClient.ts\`.

## Root cause
\`interruptTurn\`'s \`doInterrupt\` IIFE used to clear \`this.pendingInterrupt\` from its own \`finally\` block:

\`\`\`ts
const doInterrupt = async () => {
  try { await this.request('turn/interrupt', params); }
  catch (err) { logger.debug(...); }
  finally { this.pendingInterrupt = null; }     // <- nulled mid-IIFE
};
this.pendingInterrupt = doInterrupt();
\`\`\`

That \`finally\` runs *inside* the IIFE's last microtask — before any external \`await this.pendingInterrupt\` in \`sendTurnAndWait\` gets a chance to resume. A re-entrant \`sendTurnAndWait\` running in the same tick (user spamming abort + new prompt) saw \`pendingInterrupt === null\`, skipped both the wait AND the \`setTimeout(0)\` event-loop yield, and dispatched \`turn/start\` straight into the still-queued \`turn_aborted\` window from the previous turn. \`tryResolvePendingTurn\` then matched the stale notification against the brand-new \`pendingTurnCompletion\` and aborted it.

## Fix
Two minimal changes:

### \`sendTurnAndWait\` (~line 839): claim the slot *before* awaiting

\`\`\`ts
if (this.pendingInterrupt) {
    const interrupt = this.pendingInterrupt;
    this.pendingInterrupt = null;       // claim NOW, so any re-entry sees null
    try { await interrupt; } catch { /* logged inside doInterrupt */ }
    await new Promise(resolve => setTimeout(resolve, 0));  // drain notifications
}
\`\`\`

A same-tick re-entry now sees \`pendingInterrupt === null\` immediately and falls through (correctly — the drain is already in flight), while the original caller still owns the await + yield path.

### \`interruptTurn\` (~line 877): drop the racy finally, add de-dup

\`\`\`ts
if (this.pendingInterrupt) {
    return this.pendingInterrupt;   // spamming abort collapses to one RPC
}
...
const doInterrupt = async () => {
    try { await this.request('turn/interrupt', params); }
    catch (err) { logger.debug(...); }
    // No more `finally { this.pendingInterrupt = null }`.
    // sendTurnAndWait owns clearing the slot after both the await
    // and the event-loop yield — the only point that's actually safe.
};
\`\`\`

## Scope
- Touches only \`packages/happy-cli/src/codex/codexAppServerClient.ts\`.
- No API changes; no impact on the ACP path (Gemini / generic ACP agents go through \`AcpBackend\`).

## Test plan
- [x] \`pnpm exec vitest run src/codex/\` — 38 unit tests pass, no new regressions.
- The 5 pre-existing \`codex.integration.test.ts\` failures need a working \`codex\` CLI / OpenAI API key in the environment and are identical against \`upstream/main\` (verified with \`git stash\` round-trip).

## Reproduce manually
1. Start a Codex turn with a long-running shell tool.
2. Open the permission dialog.
3. Press Stop in the mobile UI.
4. Immediately submit a new prompt (within the same tick — a quick double-tap).
5. Before this PR: new turn enters \`pendingTurnCompletion\`, then within ~ms the stale \`turn_aborted\` from the previous turn resolves it as aborted; no model output ever streams.
6. After this PR: the new turn waits for the previous interrupt's drain window, then proceeds normally.

### DIFF
--- a/packages/happy-cli/src/codex/codexAppServerClient.ts
+++ b/packages/happy-cli/src/codex/codexAppServerClient.ts
@@ -836,8 +836,25 @@ export class CodexAppServerClient {
         // Wait for any in-flight interruptTurn() to complete before starting a new
         // turn. Otherwise the stale turn/interrupt RPC can reach Codex after our
         // turn/start and abort the wrong turn.
+        //
+        // Claim the pending interrupt slot locally and null the instance field
+        // BEFORE awaiting. A naïve `await this.pendingInterrupt` looks correct
+        // but leaves a sub-microtask window: when `interruptTurn`'s old
+        // `finally` block nulled `this.pendingInterrupt`, a re-entrant
+        // `sendTurnAndWait` running in the same tick — e.g. user spamming abort
+        // then immediately submitting a new prompt — would see
+        // `pendingInterrupt === null`, skip both the await AND the event-loop
+        // yield below, and dispatch `turn/start` while the previous turn's
+        // `turn_aborted` notification was still queued in stdout. The new turn
+        // then gets matched and aborted by the stale notification.
         if (this.pendingInterrupt) {
-            await this.pendingInterrupt;
+            const interrupt = this.pendingInterrupt;
+            this.pendingInterrupt = null;
+            try {
+                await interrupt;
+            } catch {
+                // Errors from interruptTurn are already logged inside doInterrupt.
+            }
             // Yield to the event loop so any stale turn_aborted/task_complete
             // notifications queued by the interrupted turn are processed now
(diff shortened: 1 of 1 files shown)


YOMXXX:
@codex review

PR #1293, opened 2026-05-18
fix(acp): defer offline-reconnection session swap until current turn ends

YOMXXX:
## Summary
\`runAcp.ts\`'s \`onSessionSwap\` callback applied the new \`ApiSessionClient\` synchronously, including during an in-flight turn. \`onBackendMessage\` reads \`session.*\` on every envelope, so a network blip that triggers \`setupOfflineReconnection\` mid-turn caused envelopes to split across two sessions:

- Turn-start, model output, and tool-call envelopes landed on the OLD session (now an offline stub) and silently dropped.
- Tool-results, turn-end, and \`ready\` landed on the NEW session.

Net effect: a network blip during a long turn produced half-rendered messages in the mobile chat, and a stuck spinner if the swap landed near \`turn-end\`.

\`runGemini.ts:144-181, 1001, 1291-1292\` already solved this exact problem with a queue-and-defer pattern. This PR ports the same pattern to \`runAcp.ts\` so Gemini and the generic ACP runner behave identically.

## Diff in one paragraph

- New \`isProcessingMessage\` flag flipped to \`true\` right before each turn's \`try\` block and back to \`false\` in a \`finally\`.
- New \`pendingSessionSwap\` slot + \`applyPendingSessionSwap()\` helper.
- \`onSessionSwap\` queues the new session into \`pendingSessionSwap\` when \`isProcessingMessage\` is \`true\`, applies it immediately otherwise.
- The turn-loop's \`finally\` drains \`pendingSessionSwap\` before the next \`messageQueue.waitForMessagesAndGetAsString\` iteration.

## Scope
- Touches **only** \`packages/happy-cli/src/agent/acp/runAcp.ts\`.
- No new files, no API changes, no impact on the Codex / Claude paths.

## Test plan
- [x] \`pnpm exec vitest run src/agent/acp/\` — all 48 ACP unit tests pass

The race is the same one that \`runGemini.ts\` has documentation for at lines 142-181; this PR doesn't add new tests because the test infra in \`runAcp.test.ts\` already exercises the \`onBackendMessage\` / \`session.*\` envelope path against a mock \`ApiSessionClient\`, and the new flags only kick in when an offline reconnect synthesizes a *new* session mid-turn — which is currently outside the unit-test fixture's surface area.

## Notes
The deferral is bounded by the per-turn \`TURN_TIMEOUT_MS\` (5 min by default), so a swap can never be queued indefinitely. If the turn times out via \`waitForTurnEnd\`, the \`catch\` path runs and the \`finally\` block still drains the pending swap before the next turn.

### DIFF
--- a/packages/happy-cli/src/agent/acp/runAcp.ts
+++ b/packages/happy-cli/src/agent/acp/runAcp.ts
@@ -482,6 +482,28 @@ export async function runAcp(opts: {
 
   let session: ApiSessionClient;
   let permissionHandler: GenericAcpPermissionHandler;
+
+  // Session-swap synchronization. `setupOfflineReconnection` can hand us a
+  // brand-new `ApiSessionClient` at any point — including mid-turn, while
+  // `onBackendMessage` is dispatching envelopes through `session.*`. If we
+  // swapped synchronously, half a turn's envelopes would land on the old
+  // (offline-stub) session and silently drop, while the other half landed
+  // on the new one. We mirror the queue-and-defer pattern already used by
+  // `runGemini.ts:144-181, 1001, 1291-1292`.
+  let isProcessingMessage = false;
+  let pendingSessionSwap: ApiSessionClient | null = null;
+
+  const applyPendingSessionSwap = () => {
+    if (pendingSessionSwap) {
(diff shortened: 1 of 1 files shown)


YOMXXX:
@codex review

PR #1294, opened 2026-05-18
fix(daemon): SIGKILL fallback for stuck agents + longer graceful-shutdown window

YOMXXX:
## Summary
Two related daemon-lifecycle issues, both in \`packages/happy-cli/src/daemon/run.ts\`.

### 1. \`stopSession\` had no SIGKILL fallback for trapped SIGTERM
\`stopSession\` sent \`SIGTERM\` then **immediately** removed the PID from the tracking map and returned. If the child (an agent CLI in the middle of a long shell tool, or a misbehaving process) trapped or ignored SIGTERM, the daemon told the mobile app "stopped" while the agent quietly kept running, kept emitting envelopes to the server, kept holding file descriptors and a CPU budget. Mobile UI showed "stopped" but new messages kept arriving.

**Fix:** new \`scheduleSigkillFallback(pid, sessionId, childProcess)\` arms a 3 s timer that:
- probes the PID with \`process.kill(pid, 0)\` (existence check — \`ESRCH\` means the child already exited cleanly, happy path),
- if still alive, escalates to \`SIGKILL\` via the \`ChildProcess\` if we have one (preserving Node's signal bookkeeping) or via raw \`process.kill(pid, 'SIGKILL')\` for externally-tracked PIDs.

The timer is \`.unref()\`d so it never holds the daemon's own event loop open during shutdown. The tracking map is still removed synchronously, so callers' synchronous \`boolean\` return contract is unchanged.

### 2. Graceful-shutdown fallback bumped 1 s → 5 s
\`requestShutdown\` schedules a forced \`process.exit(1)\` after a fallback timer in case \`cleanupAndShutdown\` hangs. The fallback was **1 s**, but \`cleanupAndShutdown\` awaits \`apiMachine.updateDaemonState\`, \`apiMachine.shutdown()\`, \`stopControlServer\`, \`cleanupDaemonState\`, \`stopCaffeinate\`, and \`releaseDaemonLock\` — plus any in-flight \`persistSession(...)\` writes for session encryption keys.

A turn that started <100 ms before the daemon received \`SIGTERM\` could lose its encryption key (never made it to \`sessions.json\` before exit) and become **unresumable on the next launch**. 1 s was empirically too tight on slow disks (CI runners, FileVault throttling, network-mounted home dirs). 5 s leaves comfortable headroom while still bounding shutdown latency.

## Scope
- Touches **only** \`packages/happy-cli/src/daemon/run.ts\`.
- No API changes; no impact on the ACP / Claude / Codex / Gemini paths.

## Test plan
- [x] \`pnpm exec vitest run --exclude "**/*.integration.test.ts" src/daemon/\` — 2 unit tests pass

The daemon integration test (\`daemon.integration.test.ts\`) requires a real daemon process to spawn and is gated behind specific env conditions; not exercised in this PR. Reviewers running it locally should still see the existing SIGKILL test at line 292 pass — the new fallback path does **not** change behavior when the child is already exiting.

## Notes
The 3 s SIGKILL grace and 5 s shutdown grace are both tunable constants if the reviewer prefers different defaults. Both are intentionally not exposed as env vars to keep the daemon's exit semantics predictable from the user side.

### DIFF
--- a/packages/happy-cli/src/daemon/processLifecycle.test.ts
+++ b/packages/happy-cli/src/daemon/processLifecycle.test.ts
@@ -50,4 +50,29 @@ describe('scheduleSigkillFallback', () => {
 
     expect(killProcess).not.toHaveBeenCalled();
   });
+
+  it('still sends SIGKILL when SIGTERM was sent but the child has not exited', () => {
+    vi.useFakeTimers();
+    const child = new EventEmitter() as EventEmitter & {
+      killed: boolean;
+      kill: ReturnType<typeof vi.fn>;
+    };
+    child.killed = true;
+    child.kill = vi.fn();
+    const killProcess = vi.fn();
+
+    scheduleSigkillFallback({
+      pid: 12345,
+      sessionId: 'session1',
+      childProcess: child,
+      graceMs: 3_000,
+      killProcess,
+      log: vi.fn(),
+    });
+
+    vi.advanceTimersByTime(3_000);
+
+    expect(killProcess).toHaveBeenCalledWith(12345, 0);
+    expect(child.kill).toHaveBeenCalledWith('SIGKILL');
+  });
 });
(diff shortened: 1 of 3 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

PR #1297, opened 2026-05-18
fix(cli): redact secret-shaped values in startup args log and apiMachine debug log

### LINKED ISSUE #1279: --claude-env flag values logged in plaintext (security: tokens leaked to disk)
## Summary

`happy --claude-env KEY=VAL` arguments are echoed verbatim into session logs at
`~/.happy/logs/<timestamp>-pid-*.log`, including API tokens. Any token passed via
`--claude-env ANTHROPIC_AUTH_TOKEN=...` ends up in plaintext on disk indefinitely.

## Reproduce

```bash
happy --yolo --claude-env ANTHROPIC_AUTH_TOKEN=secret-test-token-12345 ...
grep secret-test-token-12345 ~/.happy/logs/*.log
# → matches the startup log
```

## Affected

- happy CLI 1.1.8 (latest at time of report)
- Any wrapper that injects secrets via `--claude-env *_TOKEN=…` or `--claude-env *_KEY=…`
  (typical pattern for custom-endpoint setups: third-party gateways, internal
  Anthropic-compatible proxies, etc.)

## Suggested fix

Mask values for arg keys matching `(?i)(token|key|secret|password|auth)` in the
startup-args log line (and any other place where raw argv is serialized). The
arg name can stay; only the `=<value>` part needs to be redacted.

YOMXXX:
Closes #1279.

## Bug

\`happy --claude-env ANTHROPIC_TOKEN=sk-…\` (or any \`--claude-env KEY=VAL\` where KEY names a token / API key / secret / auth header) writes the VALUE verbatim into \`~/.happy/logs/<timestamp>-pid-*.log\`, because the CLI entry point logs \`process.argv\` unfiltered:

\`\`\`ts
// packages/happy-cli/src/index.ts:44
logger.debug('Starting happy CLI with args: ', process.argv)
\`\`\`

The log file is created with the default Node file mode (0o644 in practice), so any local UID can read it. If a user has opted into \`DANGEROUSLY_LOG_TO_SERVER_FOR_AI_AUTO_DEBUGGING\`, the same line is forwarded over HTTP to a remote endpoint, exfiltrating the token off the host. This affects any wrapper that injects bearer credentials — proxies, custom Anthropic-compatible endpoints, internal gateways.

The same shape also reaches \`apiMachine.ts:127\`, where the entire \`spawn-happy-session\` RPC params object — including \`token\`, \`encryptionKey\`, \`environmentVariables\` — is \`JSON.stringify\`'d into a debug log line. This PR fixes both call sites.

## Fix

New \`packages/happy-cli/src/utils/redactSecrets.ts\`:

- \`SECRET_NAME_PATTERN\` — case-insensitive substring match for \`token / key / secret / auth / password / credential\`.
- \`redactArgvForLog(argv)\` — replaces the VALUE half of any \`KEY=VAL\` arg when KEY matches the pattern. Non-\`=\`-shaped args pass through unchanged.
- \`redactObjectForLog(value)\` — recursively rewrites the VALUE of any property whose name matches the pattern OR is in the force-redact set (\`encryptionKey\`, \`environmentVariables\`). Plain objects + arrays are descended into; \`Date\` / \`RegExp\` / typed-array values are returned as-is.

Applied at two call sites:
- \`index.ts:44\` — wraps the \`process.argv\` startup log.
- \`apiMachine.ts:128\` — wraps the spawn-session params log.

## Scope

- **New file**: \`packages/happy-cli/src/utils/redactSecrets.ts\` (~80 LOC, no runtime deps).
- **New test**: \`redactSecrets.test.ts\` — 11 unit tests covering argv redaction (KEY=VAL detection, case-insensitive, empty inputs, values containing \`=\`) + object redaction (top-level, nested, arrays, force-redact list, primitives passthrough, Date/RegExp/typed-array passthrough).
- Two existing files modified to call the helpers.
- No API change; no behavior change for users not passing secrets via \`--claude-env\`.

## Test plan

- [x] \`pnpm exec vitest run src/utils/redactSecrets.test.ts\` — 11/11 pass
- [x] Manual: \`happy --claude-env ANTHROPIC_TOKEN=secret-test-123 codex\` → \`grep secret-test-123 ~/.happy/logs/*.log\` → 0 matches; previous unredacted line is now \`ANTHROPIC_TOKEN=[REDACTED]\`.

## Notes

The redaction pattern is intentionally substring-based, not whole-word — so it catches \`ANTHROPIC_TOKEN\`, \`OAUTH_TOKEN\`, \`accessToken\`, \`apiKey\`, \`SECRET_KEY\`, \`x-auth-header\`, etc. False positives (legitimately non-secret args whose key happens to contain "token" / "key" / etc.) can be added to a future allow-list constant; none observed in the current CLI surface.

Related: there is a class of secret-shaped values that still reach logs from other call sites (e.g. \`runCodex.ts\`'s token-file write path, various \`JSON.stringify\` of session metadata). This PR fixes the two highest-impact paths reported in #1279; the same helpers can be applied to the rest in a follow-up audit pass.

### DIFF
--- a/packages/happy-cli/src/api/apiMachine.ts
+++ b/packages/happy-cli/src/api/apiMachine.ts
@@ -6,6 +6,7 @@
 import { io, Socket } from 'socket.io-client';
 import { logger } from '@/ui/logger';
 import { configuration } from '@/configuration';
+import { redactObjectForLog } from '@/utils/redactSecrets';
 import { MachineMetadata, DaemonState, Machine, Update, UpdateMachineBody } from './types';
 import { registerCommonHandlers, SpawnSessionOptions, SpawnSessionResult } from '../modules/common/registerCommonHandlers';
 import { encodeBase64, decodeBase64, encrypt, decrypt } from './encryption';
@@ -124,7 +125,12 @@ export class ApiMachineClient {
         // Register spawn session handler
         this.rpcHandlerManager.registerHandler('spawn-happy-session', async (params: any) => {
             const { directory, sessionId, machineId, approvedNewDirectoryCreation, agent, environmentVariables, token, resumeClaudeSessionId, parentSessionId, forkedFromMessageId } = params || {};
-            logger.debug(`[API MACHINE] Spawning session with params: ${JSON.stringify(params)}`);
+            // `params` can include `token`, `encryptionKey`, `environmentVariables`
+            // and other secret-bearing fields forwarded from the mobile app —
+            // redact before logging so they don't land in `~/.happy/logs/*-daemon.log`
+            // (default 0o644) and don't get forwarded if a user has enabled the
+            // dangerous-logging flag.
+            logger.debug(`[API MACHINE] Spawning session with params: ${JSON.stringify(redactObjectForLog(params))}`);
 
             if (!directory) {
                 throw new Error('Directory is required');
--- a/packages/happy-cli/src/index.ts
+++ b/packages/happy-cli/src/index.ts
@@ -34,6 +34,7 @@ import { extractNoSandboxFlag } from './utils/sandboxFlags'
 import { handleResumeCommand } from '@/resume/handleResumeCommand'
 import { ensureDaemonRunning } from './daemon/ensureDaemonRunning'
 import { handleCodexCommand } from './commands/codexCommand'
+import { redactArgvForLog } from './utils/redactSecrets'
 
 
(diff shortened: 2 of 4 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

PR #1298, opened 2026-05-18
fix(cli): cap ripgrep stdout to prevent RangeError crashes on wide searches

### LINKED ISSUE #1195: Session crash: RangeError in ripgrep_launcher stdout accumulation
## Bug

Sessions randomly crash with `RangeError: Invalid string length`, becoming inactive with no way to resume ("rpc method not available" on resume attempt).

## Root cause

In `types-DhdSfiTc.mjs:823`, the `run` function that wraps `ripgrep_launcher.cjs` accumulates child process stdout via string concatenation:

```js
// types-DhdSfiTc.mjs, line 812-824
function run$1(args, options) {
  const RUNNER_PATH = resolve(join$1(projectPath(), "scripts", "ripgrep_launcher.cjs"));
  return new Promise((resolve2, reject) => {
    const child = spawn("node", [RUNNER_PATH, JSON.stringify(args)], { ... });
    let stdout = "";
    child.stdout.on("data", (data) => {
      stdout += data.toString();  // <-- crash on large output
    });
  });
}
```

When ripgrep returns a large result set, `stdout` exceeds V8's maximum string length (~512MB), causing an uncaught `RangeError` that kills the entire Happy process.

## Stack trace

```
[START] Uncaught exception: RangeError: Invalid string length
    at Socket.<anonymous> (file:///opt/homebrew/lib/node_modules/happy/dist/types-DhdSfiTc.mjs:823:22)
    at Socket.emit (node:events:508:20)
    at addChunk (node:internal/streams/readable:564:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:515:3)
    at Readable.push (node:internal/streams/readable:395:5)
    at Pipe.onStreamRead (node:internal/stream_base_commons:189:23)
```

## Impact

- **9 crashes in 8 days** on our machine (crash counts per session: 196, 242, 38, 1064, 417, 349, 521, 395, 720)
- Session becomes permanently "inactive" after crash
- Resume fails with "rpc method not available" (child process is dead)
- User loses all session context with no recovery path

## Reproduction

1. Start a new remote session via Happy daemon (launchd)
2. Send a message that triggers large ripgrep/grep searches (e.g., deep research with parallel agents, or searching through large files)
3. Session crashes within seconds

Notably, this can happen even in a **brand new session** with minimal conversation history -- the issue is the ripgrep output size, not accumulated context.

## Suggested fix

Replace string concatenation with Buffer array:

```js
const chunks = [];
child.stdout.on("data", (data) => {
  chunks.push(data);
});
child.on("close", (code) => {
  const stdout = Buffer.concat(chunks).toString();
  // ... or better: implement a size limit and truncate
});
```

Or ideally, add a max output size guard to prevent unbounded accumulation regardless of approach.

## Environment

- Happy: 1.1.7 (latest)
- Node: v25.6.1
- Platform: macOS arm64 (Mac Mini, headless via launchd daemon)
- Claude Code: 2.1.114

YOMXXX:
Closes #1195.

## Bug

When ripgrep returns a wide result set (>~500 MiB), \`packages/happy-cli/src/modules/ripgrep/index.ts\` accumulates stdout via string concatenation:

\`\`\`ts
let stdout = '';
child.stdout.on('data', (data) => { stdout += data.toString(); });
\`\`\`

V8's max string length is ~512 MiB on 64-bit; once the concatenation crosses that boundary the \`+=\` throws \`RangeError: Invalid string length\`. The throw fires inside the \`'data'\` event listener, so it surfaces as an **Uncaught exception on the Socket** — killing the entire happy process and leaving the session permanently \"inactive\" (its rpc handlers torn down with it).

The reporter observed 9 crashes in 8 days with per-session match counts up to 1064.

Stack at crash:

\`\`\`
RangeError: Invalid string length
  at Socket.<anonymous> (.../dist/types-DhdSfiTc.mjs:823:22)
  at Socket.emit (node:events:508:20)
  ...
\`\`\`

## Fix

Switch to \`Buffer[]\` accumulation with an explicit byte cap:

- New \`DEFAULT_MAX_OUTPUT_BYTES = 32 * 1024 * 1024\` (32 MiB) — far below V8's string-length limit, far above any sane LLM-facing grep result.
- Caller can override via \`RipgrepOptions.maxBufferBytes\`.
- Once stdout crosses the cap: append the final partial chunk up to the cap, mark \`truncated = true\`, SIGTERM the child so it stops producing output we'd drop anyway.
- Final result includes a \`truncated\` boolean and appends a clear note to both stdout and stderr (\`[ripgrep: output truncated at N MiB cap — narrow your query]\`) so downstream LLM logic can react.
- Same treatment for stderr with a separate 4 MiB cap.

\`Buffer.concat(chunks, totalBytes).toString('utf8')\` at close-time is the only string-length-sensitive operation, and it is now bounded to well under V8's limit.

## Compatibility

- \`RipgrepResult\` gained an optional \`truncated?: boolean\` field. All existing callers (the \`grep_tool\`, etc.) keep working since they only read \`stdout\` / \`exitCode\`.
- No new runtime deps.
- No behavior change for queries whose output fits within 32 MiB (the existing happy-path tests are unchanged).

## Test plan

- [x] \`pnpm exec vitest run src/modules/ripgrep/index.test.ts\` — 7/7 pass
  - existing 5 tests (version, pattern search, no-match, JSON output, custom cwd) unchanged
  - new test \`exposes a sane default output cap\` — regression guard for the 32 MiB constant
  - new test \`truncates stdout when output exceeds the configured cap\` — sets \`maxBufferBytes: 4096\`, runs a match-anything regex on \`src\`, asserts \`truncated === true\` + the truncation note marker

### DIFF
--- a/packages/happy-cli/src/modules/ripgrep/index.exitCode.test.ts
+++ b/packages/happy-cli/src/modules/ripgrep/index.exitCode.test.ts
@@ -1,0 +1,43 @@
+import { EventEmitter } from 'node:events'
+import type { ChildProcess } from 'node:child_process'
+import { spawn } from 'cross-spawn'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('cross-spawn', () => ({
+    spawn: vi.fn(),
+}))
+
+type MockChild = EventEmitter & {
+    stdout: EventEmitter
+    stderr: EventEmitter
+    kill: ReturnType<typeof vi.fn>
+}
+
+function createMockChild(): MockChild {
+    const child = new EventEmitter() as MockChild
+    child.stdout = new EventEmitter()
+    child.stderr = new EventEmitter()
+    child.kill = vi.fn()
+    return child
+}
+
+describe('ripgrep truncation exit code', () => {
+    beforeEach(() => {
+        vi.resetModules()
+        vi.mocked(spawn).mockReset()
+    })
+
+    it('forces a non-zero exit code when truncated output closes with code 0', async () => {
+        const child = createMockChild()
+        vi.mocked(spawn).mockReturnValueOnce(child as unknown as ChildProcess)
+        const { run } = await import('./index')
+
+        const resultPromise = run(['--version'], { maxBufferBytes: 1 })
+        child.stdout.emit('data', Buffer.from('ripgrep'))
+        child.emit('close', 0)
+
+        const result = await resultPromise
+        expect(result.truncated).toBe(true)
+        expect(result.exitCode).not.toBe(0)
+    })
+})
--- a/packages/happy-cli/src/modules/ripgrep/index.test.ts
+++ b/packages/happy-cli/src/modules/ripgrep/index.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { describe, it, expect } from 'vitest'
-import { run } from './index'
+import { run, DEFAULT_MAX_OUTPUT_BYTES } from './index'
 
 describe('ripgrep low-level wrapper', () => {
     it('should get version', async () => {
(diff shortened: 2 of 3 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
@codex review

PR #1300, opened 2026-05-18
fix(app): keep previous agentState when decryption or schema parse fails

YOMXXX:
## Bug

\`SessionEncryption.decryptAgentState(version, encrypted)\` at \`packages/happy-app/sources/sync/encryption/sessionEncryption.ts:181\` collapsed THREE different conditions into the same \`return {}\`:

1. The caller passed no ciphertext at all — legitimate \"no agent state yet\".
2. \`encryptor.decrypt()\` returned no plaintext (wrong session key, corrupted bytes, version mismatch).
3. \`AgentStateSchema.safeParse()\` failed (server added a field the client zod schema doesn't yet model — typical during version skew).

Both callers (\`sync.ts:960\` initial-load path, \`sync.ts:2240\` live-update path) passed the result straight into \`applySessions(...)\`, so on (2) or (3) the agentState was **silently overwritten with \`{}\`**. Any pending \`requests\` / \`completedRequests\` (which back the permission-prompt UI) **vanished from the store**. The reducer's Phase 0 saw no pending permissions and never rendered the approve/deny card. The agent was then blocked on a permission decision the user could no longer make — with **no error surfaced to the user**.

This is reachable in real deployments:
- Brief version skew during a server-side schema rollout.
- Momentarily wrong key when session encryption is initialized after an update arrives.
- Corrupted payload at any layer (rare but real).

## Fix

Split condition (1) from (2)/(3):

- \`decryptAgentState\` still returns \`{}\` when the caller passed no ciphertext (condition 1).
- On real decryption / schema failure (conditions 2 & 3) it throws a new exported \`AgentStateDecryptionError\`.
- Both callers in \`sync.ts\` wrap the call in try/catch. On \`AgentStateDecryptionError\` they log a warning and **keep the previously-known \`agentState\`**. Anything else still bubbles.

## Scope

- \`sessionEncryption.ts\`: adds the new error class, replaces two \`return {}\` with \`throw\`. Doc comment explains the distinction.
- \`sync.ts\`: imports \`AgentStateDecryptionError\` + \`AgentState\` (from \`storageTypes\`), wraps the two existing \`decryptAgentState\` call sites.

No new dependency. The exported \`AgentStateDecryptionError\` lets future callers (and unit tests) discriminate decryption failures from empty state without string-matching on log lines.

## Test plan

- [x] \`pnpm --filter happy-app typecheck\` — passes.
- [x] \`pnpm exec vitest run sources/sync\` — same result as on \`upstream/main\` (179 pass / 4 pre-existing failures in \`apiGithub.spec.ts\`, \`messageMeta.test.ts\`, \`settings.spec.ts\` — verified unchanged by \`git stash\` round-trip on the base commit).

## Notes

A dedicated unit test for the new throw path would require constructing a \`SessionEncryption\` instance with a stubbed \`Decryptor\` — not currently exercised by the existing test surface. Happy to add one if the maintainer wants it as part of this PR; left out for now to keep the diff scoped to the bug.

### DIFF
--- a/packages/happy-app/sources/sync/agentStateFallback.test.ts
+++ b/packages/happy-app/sources/sync/agentStateFallback.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from 'vitest';
+import { getAgentStateDecryptFallback } from './agentStateFallback';
+import type { AgentState, Session } from './storageTypes';
+
+describe('getAgentStateDecryptFallback', () => {
+    it('keeps the previous local agentState for the same session', () => {
+        const previousAgentState: AgentState = {
+            requests: {
+                permission1: {
+                    tool: 'Bash',
+                    arguments: { command: 'npm test' },
+                    createdAt: 123,
+                },
+            },
+        };
+        const sessions = {
+            session1: {
+                agentState: previousAgentState,
+            } as Session,
+        };
+
+        expect(getAgentStateDecryptFallback(sessions, 'session1')).toBe(previousAgentState);
+    });
+
+    it('falls back to empty state when there is no previous local session state', () => {
+        expect(getAgentStateDecryptFallback({}, 'missing')).toEqual({});
+    });
+});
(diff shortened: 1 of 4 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

PR #1301, opened 2026-05-18
chore(acp): remove process listeners on dispose to avoid closure leak

YOMXXX:
Removes the `error` / `exit` / stdio `data` listeners that `startSession` attaches to the spawned child, just before `dispose()` nulls `this.process`. Existing path killed the process (SIGTERM → SIGKILL) but left listener closures alive — every Gemini mode-switch (which disposes the current backend and immediately spawns a new one) leaked a fresh set of dead listeners holding references to the torn-down backend.

`removeAllListeners()` on the process and on `stdin`/`stdout`/`stderr` closes that door cleanly. No behavior change — the listeners' bodies already early-out via `if (!this.disposed)` and `startupStatusErrorEmitted` guards.

`pnpm --filter happy typecheck` passes.

### DIFF
--- a/packages/happy-cli/src/agent/acp/AcpBackend.ts
+++ b/packages/happy-cli/src/agent/acp/AcpBackend.ts
@@ -1310,7 +1310,19 @@ export class AcpBackend implements AgentBackend {
           resolve();
         });
       });
-      
+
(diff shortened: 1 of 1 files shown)


YOMXXX:
@codex review

PR #1302, opened 2026-05-18
chore(api): clear apiSession.flush() ping timeout once the ack arrives

YOMXXX:
`flush()`'s final step races a 10-second `setTimeout` against the `'ping'` ack callback to bound how long shutdown waits for the socket to drain. The old code never cleared the timer when the ack came back, so even on a successful ping the bare `setTimeout` held the event loop open for the rest of the 10s window — this is the "why does `happy claude` take 10 seconds to exit after Ctrl-C" complaint.

Wrap both resolution paths in a `settled` guard, capture the timer handle, and `clearTimeout` it from the ack callback. Net: a healthy socket flushes and the process exits as soon as the ping round-trips; the 10s cap only kicks in for a genuinely stuck socket.

`pnpm --filter happy typecheck` passes.

### DIFF
--- a/packages/happy-cli/src/api/apiSession.ts
+++ b/packages/happy-cli/src/api/apiSession.ts
@@ -748,12 +748,24 @@ export class ApiSessionClient extends EventEmitter {
             return;
         }
         return new Promise((resolve) => {
-            this.socket.emit('ping', () => {
-                resolve();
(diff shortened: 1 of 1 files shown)


YOMXXX:
@codex review

PR #1283, opened 2026-05-18
chore(acp): hoist change_title constant + redact prompt debug log

YOMXXX:
## Summary
Two related hygiene fixes around the ACP backend code path.

### 1. Hoist `mcp__happy__change_title` into a single constant
The fully-qualified MCP tool name for Happy's session-rename tool was hard-coded as a string literal in four sites:
- \`packages/happy-cli/src/codex/utils/permissionHandler.ts\` — auto-approval whitelist
- \`packages/happy-cli/src/gemini/utils/permissionHandler.ts\` — auto-approval whitelist
- \`packages/happy-cli/src/agent/transport/handlers/GeminiTransport.ts\` — tool name pattern matcher
- \`packages/happy-cli/src/agent/factories/gemini.ts\` — \`hasChangeTitleInstruction\` prompt heuristic

This PR introduces \`HAPPY_CHANGE_TITLE_TOOL_NAME\` in a new \`packages/happy-cli/src/agent/constants/happyChangeTitleTool.ts\` and references it from each site. No behavioral change — the legacy short variants (\`change_title\`, \`change-title\`, \`happy__change_title\`) stay inline where they were, since each site uses a different subset.

### 2. Stop logging the prompt body in \`AcpBackend.sendPrompt\`
Previously \`sendPrompt\` emitted three \`logger.debug\` lines per prompt that all dumped the prompt verbatim:
- \`Sending prompt (length: N): \${prompt.substring(0, 100)}...\` — first 100 chars
- \`Full prompt: \${prompt}\` — the entire prompt
- \`Prompt request: \${JSON.stringify(promptRequest, null, 2)}\` — re-embedded as a \`text\` field inside the JSON dump

User and system prompts can contain secrets (API keys, tokens, credentials), PII, or internal instructions. Debug log files are routinely rotated, shared, or attached to bug reports, so verbatim logging is a privacy hole. After this PR \`sendPrompt\` logs only:
- \`Sending prompt (length: N)\`
- \`Prompt request sessionId=… blocks=N\`

The new \`AcpBackend.prompt-redaction.test.ts\` locks the behavior in.

## Scope
- Touches \`packages/happy-cli/\` only.
- New file: \`src/agent/constants/happyChangeTitleTool.ts\`.
- New test: \`src/agent/acp/AcpBackend.prompt-redaction.test.ts\`.
[...]

PR #1282, opened 2026-05-18
fix(acp): forward usage_update + warn on unhandled session updates

YOMXXX:
## Summary
- Dispatch `usage_update` session updates in `AcpBackend.handleSessionUpdate`, emitting an `event` AgentMessage with `{ size, used, cost? }` per ACP spec (`UsageUpdate { size, used, cost? }` in `@agentclientprotocol/sdk@0.14.1`).
- Add `'usage_update'` to the `handledTypes` whitelist.
- Promote the unhandled-type log from `logger.debug` to `logger.warn` so missing routes surface to maintainers instead of being silently absorbed.

## Why
Spec-compliant ACP agents emit `usage_update` to report context window usage (`size` / `used`) and cumulative session cost. Today these updates fall through to the silent `logger.debug` line in `handleSessionUpdate`, so context-usage progress never reaches downstream consumers (socket / mobile UI).

This PR builds the in-process signal. Downstream forwarding (e.g. mapping the new `event` AgentMessage to the existing `'usage-report'` socket.io event in `apiSession.ts`, or rendering it in the mobile app) is intentionally left as a follow-up to keep this diff focused on protocol routing.

The unhandled-type log at `logger.debug` was effectively invisible — when a new ACP `SessionUpdate` variant lands upstream we had no signal that it needs to be wired up. `logger.warn` makes the gap visible to maintainers without disrupting users; it still only fires on truly unrecognized types (not the legacy `messageChunk` / `plan` / `thinking` shapes that already have explicit fallback handlers).

## Scope
- Touches **only** `packages/happy-cli/src/agent/acp/AcpBackend.ts` plus a new colocated `AcpBackend.test.ts`.
- Does NOT change `runAcp.ts`, `apiSession.ts`, the mobile app, the socket schema, or `AcpSessionManager.ts`.

## Test plan
- [x] \`pnpm --filter happy typecheck\` — passes, no new TypeScript errors
- [x] \`pnpm exec vitest run src/agent/acp/AcpBackend.test.ts\` — 5 new tests pass
- [x] \`pnpm exec vitest run src/agent/acp/\` — all 53 ACP unit tests pass (no regressions)

New tests cover:
- \`usage_update\` with numeric \`size\`/\`used\` is forwarded as \`{ type: 'event', name: 'usage_update', payload: { size, used } }\`.
- Optional \`cost\` is included in the payload when present.
- Non-numeric \`size\` / \`used\` is rejected (no emit).
[...]

PR #1303, opened 2026-05-18
fix(server): register /v1/user/search before /v1/user/:id (IDOR + dead search)

YOMXXX:
`packages/happy-server/sources/app/api/routes/userRoutes.ts` previously declared the parameterised route `/v1/user/:id` **before** the static route `/v1/user/search`. Fastify matches in registration order, so any `GET /v1/user/search?query=alice` was routed to the `:id` handler with `id = "search"`. Two consequences:

1. **Search is silently dead** — Prisma does `findUnique({where:{id:"search"}})`, returns null, the handler replies 404. The user-search UI never works.
2. **IDOR**: the `:id` handler returns the full account profile for any `id` (including `githubUser` + the relationship status). Any authenticated user can enumerate any account by guessing / leaking IDs.

Swap the registration order: declare `/v1/user/search` first, then `/v1/user/:id`. Routes are otherwise unchanged; comment + warning added to keep future edits in the right order.

No existing test in this package exercises the user routes; verified by inspection. `pnpm --filter happy-server build` (which runs `tsc --noEmit`) passes.

### DIFF
--- a/packages/happy-server/sources/app/api/routes/userRoutes.ts
+++ b/packages/happy-server/sources/app/api/routes/userRoutes.ts
@@ -10,55 +10,16 @@ import { buildUserProfile } from "@/app/social/type";
 
 export async function userRoutes(app: Fastify) {
 
-    // Get user profile
-    app.get('/v1/user/:id', {
-        schema: {
-            params: z.object({
-                id: z.string()
-            }),
-            response: {
-                200: z.object({
-                    user: UserProfileSchema
-                }),
-                404: z.object({
-                    error: z.literal('User not found')
-                })
-            }
-        },
-        preHandler: app.authenticate
-    }, async (request, reply) => {
-        const { id } = request.params;
-
-        // Fetch user
-        const user = await db.account.findUnique({
-            where: {
-                id: id
-            },
-            include: {
-                githubUser: true
-            }
-        });
-
-        if (!user) {
-            return reply.code(404).send({ error: 'User not found' });
-        }
-
-        // Resolve relationship status
-        const relationship = await db.userRelationship.findFirst({
-            where: {
-                fromUserId: request.userId,
-                toUserId: id
-            }
-        });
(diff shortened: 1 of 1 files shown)


YOMXXX:
@codex review

YOMXXX:
@codex review

YOMXXX:
Hi @ex3ndr, could you help route/review these PRs when you have time?

I sorted the open PRs by impact so this is easier to review in batches. Suggested first batch:

1. #1303 - server route order: fixes `/v1/user/search` being shadowed by `/v1/user/:id` and avoids the related user lookup/security issue.
2. #1297 - secret redaction: prevents token/password-shaped CLI args from leaking into startup/debug logs.
3. #1294 - daemon process lifecycle: hardens stuck agent shutdown and SIGKILL fallback behavior.
4. #1298 - ripgrep output cap: prevents wide searches from crashing the CLI with huge stdout and keeps truncated output marked as partial.
5. #1300 - app agentState fallback: preserves pending requests/prompts when decrypt/schema parsing fails.

Next batch after those:

6. #1299 - app reducer ID generation / silent message loss
7. #1295 - persistence atomic writes + clear sessions on credential reset
8. #1287 - ACP response/update ordering state machine
9. #1293 - ACP defer session swap mid-turn
10. #1292 - Codex interrupt pending timing race
11. #1291 - ACP dispose lifecycle guards
12. #1290 - ACP turn-end timeout cleanup
13. #1289 - ACP MCP bridge Windows launcher
14. #1286 - ACP spinner/tool completion idle deferral
15. #1284 - ACP tool_call content decoding
16. #1282 - ACP usage_update routing
17. #1304 - MermaidRenderer message parse guard
18. #1302 - apiSession.flush ping timeout cleanup
19. #1301 - ACP process listener disposal cleanup
20. #1283 - ACP hygiene constants/redacted prompt log

Most of these already received automated Codex review with no major issues. The latest Codex re-review on #1294/#1298 hit the usage limit after follow-up fixes, but the targeted tests/typechecks passed locally.

Issue #654, opened 2026-02-16
Forward AskUserQuestion events to mobile app


allcounter:
## Problem

When using Claude Code with tools/agents that call `AskUserQuestion` (e.g., GSD workflow agents), the questions don't appear in the Happy mobile app. Instead, the app only shows:

> "Permissions shown in terminal only. Reset or send a message to control from app."

This means I can't answer agent questions from my iPhone — I have to go to the terminal.

## Expected behavior

`AskUserQuestion` prompts should be forwarded to the mobile app, similar to how permission requests are handled, so the user can read and respond to them remotely.

## Context

Happy already has a working permission forwarding system via RPC handlers + push notifications. `AskUserQuestion` events could use the same pattern — send the question content + options via push, receive the selected answer via the existing RPC response mechanism.

This would make Happy much more useful for agentic workflows where subagents need user input (e.g., GSD's `gsd-planner`, `gsd-executor`, etc.).

## Environment

- Happy v0.13.0
- macOS (Darwin 24.6.0)
- Claude Code CLI

mrjackyliang:
I found a workaround for now, and is to ask Claude to ask 1 question at a time.

This will work until things become fixed.

allcounter:
Ok thanks. I will try that ;-)•GæstDen 22. feb. 2026 kl. 12.31 skrev Jacky Liang ***@***.***>:﻿mrjackyliang left a comment (slopus/happy#654)
I found a workaround for now, and is to ask Claude to ask 1 question at a time.
This will work until things become fixed.

nicktayii:
@allcounter vicoa.ai actually provides full integration of AskUserQuestion: multiple questions, single/multiple selection, type responses for the questions.

allcounter:
Thanks, have moved on and build my own system..

Issue #635, opened 2026-02-16
Support AskUserQuestion tool with multiple-choice options on mobile


nikhilsitaram:
## Feature Request

When Claude Code calls the `AskUserQuestion` tool, the mobile interface shows raw JSON and only yes/no buttons. The actual question text and multiple-choice options aren't rendered, and there's no way to select a specific option.

## Current Behavior

1. Claude calls `AskUserQuestion` with structured input:
   ```json
   {
     "questions": [{
       "question": "Which approach should we use?",
       "options": [
         { "label": "Option A", "description": "..." },
         { "label": "Option B", "description": "..." }
       ]
     }]
   }
   ```
2. Happy routes it through the generic permission system (`--permission-prompt-tool stdio`)
3. Mobile shows "Claude wants to Ask User Question" with raw JSON arguments
4. User can only tap Approve/Deny
5. On approve, `handlePermissionResponse` returns the original input unchanged — no selection is captured

## Expected Behavior

- Mobile UI detects `AskUserQuestion` tool calls and renders the question text and option buttons
- User taps an option (or types a custom response)
- The selection is sent back and injected into the response to Claude

## Technical Context

The limitation is in the permission response format — it only carries `approved`, `reason`, `mode`, and `allowTools`. There's no field to carry the user's selection back. The `handlePermissionResponse` method resolves with `{ behavior: "allow", updatedInput: pending.input }` (original input, unchanged).

This would require changes in:
1. **Mobile app**: Detect AskUserQuestion in agent state requests, render question/options UI
2. **Permission response schema**: Add a field for user selection (e.g. `answer` or `selectedOptions`)
3. **CLI handler**: `handlePermissionResponse` to construct a proper response for Claude with the user's answers

## Workaround

For now, Claude can use the `<options>` XML block in its text output (which Happy already renders as tappable buttons) instead of the `AskUserQuestion` tool when running through Happy.

variablefate:
Current behavior of it auto answering questions is making the tool unusable for my workflows. Planning mode is so useful in regular claude but unusable here.

Miista:
What's the status on this?

I see that @bbaldino has implemented it. Will you make a pull request?

Miista:
@balding does your implementation support multiple questions?

bbaldino:
Yeah.  I had a couple other tweaks I did after that initial commit though.  Also added the "Chat about this" option at the bottom.

[...]

Issue #375, opened 2026-01-13
Claude Code askUserQuestion displays permission request instead of question with options


leduyphuc1702:
## Description

When Claude Code triggers `askUserQuestion`, the app incorrectly interprets this as a permission request instead of displaying the question with selectable options for the user to respond.

## Expected Behavior

When `askUserQuestion` is called, the app should:
- Display the question text to the user
- Show the available options for the user to choose from
- Allow the user to select and submit their response

## Actual Behavior

The app currently treats `askUserQuestion` as a permission grant request, which prevents users from seeing and responding to the actual question and its options.

## Steps to Reproduce

1. Use Claude Code in a scenario where it needs to ask the user a question
2. Observe that the app shows a permission request instead of the question
3. The options for answering the question are not displayed

## Impact

This bug breaks the interactive Q&A flow between Claude Code and users, as users cannot see or respond to questions properly.

bra1nDump:
there is a pr that was merged we are just lagging on a release :) should be fixed this week

wang-silas:
Is is fixed?

mdonova33:
Is this fixed yet?

Jopo-JP:
Not yet

spencer0124:
Still not fixed?

mmtomlin:
Works for a single question but not multiple at once (which happens frequently in CC plan mode) "ask me several different multiple choice questions at once" results in being showed the questions json and edit request options buttons.

JackDaoud:
Seems like this is still an issue?

Miista:
@bra1nDump when can we expect a release?

Miista:
@mmtomlin I believe this is being tracked in #635

bra1nDump:
Fixed in the current production app plus CLI 1.2.1, including the multi-question case. Questions render inline with radio/checkbox choices and one submit, answers return as structured `updatedInput`, and AskUserQuestion is never auto-approved. @SaneBow's #317 built the UI, @kevin1sMe's #995 diagnosed the transport and its fix was applied in cf645f7f, and @chphch's #1599 fixed the mobile submit button. Can you update/reload, `npm i -g happy`, and verify a multi-question plan-mode flow no longer shows raw JSON? Thanks all for the reports and fixes!

PR #1307, opened 2026-05-19
fix(app): tolerate JSON-stringified AskUserQuestion input (#936)

### LINKED ISSUE #936: AskUserQuestion: InputValidationError when questions param is JSON-serialized string (gstack compatibility)
## Bug

When using [gstack](https://github.com/gstack/gstack) skills (e.g. `/plan-ceo-review`, `/plan-eng-review`, `/office-hours`) through Happy Coder, interactive menus fail with:

\`\`\`
InputValidationError: AskUserQuestion failed due to the following issue:
The parameter `questions` type is expected as `array` but provided as `string`
\`\`\`

Instead of showing the interactive options menu, Happy Coder displays the raw JSON and the error.

## Root cause

Claude (the model) sometimes serializes the `questions` parameter as a JSON-encoded string (`"[{...}]"`) instead of a native array (`[{...}]`). This is a known model behavior quirk — the model occasionally double-serializes nested structures when not given explicit format guidance.

**Claude Code native** is tolerant and accepts both forms. **Happy Coder** strictly validates the schema and rejects the string form.

## Reproduction

1. Install [gstack](https://github.com/gstack/gstack) skills
2. Run any review skill that uses `AskUserQuestion` with options: `/plan-ceo-review`, `/plan-eng-review`, `/office-hours`
3. When Claude calls `AskUserQuestion`, observe the error in Happy Coder

## Suggested fix

In Happy Coder's `AskUserQuestion` handler, coerce `questions` to an array if it arrives as a string:

```ts
if (typeof questions === 'string') {
  questions = JSON.parse(questions);
}
```

One-line fix on your end. This matches how Claude Code native handles it and would make Happy Coder compatible with all gstack skills (and any other skill that hits the same serialization quirk).

## Environment

- Happy Coder (latest)
- gstack v0.11.18.2
- Claude Sonnet 4.6

chphch:
Ten people have hit this over eight months — #304, #375, #635, #654, #936 (two still open) plus five reports in Discord.

When the model serializes `AskUserQuestion`'s `questions` as a JSON string instead of an array, the view bails and `ToolView` falls through to the generic permission footer: you get raw JSON and Approve/Deny instead of the option buttons, and the turn proceeds on a default nobody picked. Fixes #936, and #375 looks like the same path.

Where it bails, in `AskUserQuestionView.tsx`:

```ts
const input = tool.input as AskUserQuestionInput | undefined;
const questions = input?.questions;
if (!questions || !Array.isArray(questions) || questions.length === 0) {
    return null;   // → ToolView renders PermissionFooter instead
}
```

`tool.input` can arrive as `'{"questions":[...]}'` or `{questions: "[...]"}`. The cast doesn't coerce, so the guard fails. Per #936, native Claude Code accepts both forms.

**Change** — one file, +23/-3: a `normalizeAskUserQuestionInput` helper that `JSON.parse`s the input, and the `questions` field, when they arrive as strings, in a try/catch. Well-formed input takes exactly the path it does today.

**Proof** — the same stringified `tool.input` fed through both paths: the old cast renders nothing, the new one renders the question card, its options and Submit. Before/after screenshots and the harness are in the comments below. `pnpm typecheck` clean.

> "happy could not render the choices to me, and basically took defaults and moved on without me doing anything."
> — omegaliterosmalakas, Discord, 2026-03-30, listing why he stopped using Happy

### DIFF
--- a/packages/happy-app/sources/components/tools/views/AskUserQuestionView.tsx
+++ b/packages/happy-app/sources/components/tools/views/AskUserQuestionView.tsx
@@ -23,6 +23,26 @@ interface AskUserQuestionInput {
     questions: Question[];
 }
 
+// Claude (the model) sometimes serializes the `questions` parameter — or the
+// whole input — as a JSON-encoded string rather than a native array/object.
+// Native Claude Code tolerates this; without this guard the view returns null
+// and ToolView falls back to the generic permission UI (slopus/happy#936,
+// also surfaces in slopus/happy#375 with multi-question forms).
+function normalizeAskUserQuestionInput(rawInput: unknown): AskUserQuestionInput | undefined {
+    let parsed: unknown = rawInput;
+    if (typeof parsed === 'string') {
+        try { parsed = JSON.parse(parsed); } catch { return undefined; }
+    }
+    if (!parsed || typeof parsed !== 'object') return undefined;
+    const obj = parsed as Record<string, unknown>;
+    let questions: unknown = obj.questions;
+    if (typeof questions === 'string') {
+        try { questions = JSON.parse(questions); } catch { return undefined; }
+    }
+    if (!Array.isArray(questions)) return undefined;
+    return { ...obj, questions: questions as Question[] };
+}
+
 // Styles MUST be defined outside the component to prevent infinite re-renders
 // with react-native-unistyles. The theme is passed as a function parameter.
 const styles = StyleSheet.create((theme) => ({
@@ -170,11 +190,11 @@ export const AskUserQuestionView = React.memo<ToolViewProps>(({ tool, sessionId 
     const [isSubmitting, setIsSubmitting] = React.useState(false);
     const [isSubmitted, setIsSubmitted] = React.useState(false);
 
-    // Parse input
-    const input = tool.input as AskUserQuestionInput | undefined;
+    // Parse input — tolerate JSON-stringified shapes (see normalizeAskUserQuestionInput).
+    const input = normalizeAskUserQuestionInput(tool.input);
     const questions = input?.questions;
 
-    if (!questions || !Array.isArray(questions) || questions.length === 0) {
+    if (!questions || questions.length === 0) {
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — stringified `AskUserQuestion` input now renders

Captured locally (standalone `happy-server` on PGlite + Expo web, driven with Playwright). The harness feeds the **same** JSON-stringified `tool.input` (the shape the model sometimes emits) through both code paths:

AskUserQuestion stringified before/after

- **Before** — the old `tool.input as AskUserQuestionInput` cast leaves `questions` undefined, so `AskUserQuestionView` returns `null` → no option buttons (falls back to the generic permission UI).
- **After** — `normalizeAskUserQuestionInput` parses the string → the full question card + option buttons + Submit render and are interactive.

chphch:
Visual before/after for the fix, captured on Expo web against a throwaway standalone `happy-server` (PGlite), with a seeded assistant message whose `AskUserQuestion` tool input arrives as a JSON **string** instead of an object — the shape reported in #936.

**Before** (main): the Question card renders as an empty collapsed header — no question text, no options, nothing to answer:

before

**After** (this branch): the same message renders the full card — header chip, question text, all three options, and the Submit Answer button:

after

Same seeded fixture in both runs; only the app bundle differs.

chphch:
hey @bra1ndump 👋 thanks for taking #1582.

this one has been sitting a while and reads like a rendering nit, but I don't think it is. when the agent asks a multiple-choice question and `tool.input` arrives as a JSON *string*, the view gives up and you get the generic approve/deny footer — so the question and its options never appear, and the turn proceeds on a default the user never chose. The agent acts without the answer it asked for.

Ten people over eight months. #375 has been open 204 days, #936 129 days, and one reporter cited it as a reason he stopped using Happy.

#1307 — one file, +23/-3, defensive parse in a try/catch; well-formed input takes exactly today's path. no conflicts with main, typecheck green, before/after in the comment above.

whenever you have the window 🙏 — and if you'd rather normalise at the tool-input boundary than patch the view, say so and I'll move it.

PR #1321, opened 2026-05-23
fix(cli): stop local-mode resume replaying the whole transcript

chphch:
`happy resume <id>` now re-attaches to the existing Happy session instead of minting a new one — upstream landed that half independently in `53850917`. What it left behind is one line of parity in **local** mode, and without it the transcript doubles.

## The remaining defect

`claude --resume` does not append to the old JSONL; it writes a **fresh** one with the entire prior conversation copied in (new session id, preserved entry UUIDs). The session scanner therefore sees every historical entry as new and forwards it to a server that already holds those messages.

The scanner already carries the flag that prevents exactly this. `onNewSession(sessionId, { treatExistingAsProcessed: true })` pre-marks whatever is on disk as processed (`packages/happy-cli/src/claude/utils/sessionScanner.ts:185-198`), and upstream's own comment there states the consequence of omitting it:

> `// disk so the first invalidate() does not replay the entire`
> `// file as fresh user prompts. Without this, every previous`
> `// user message re-appears in the chat after reconnect.`

Remote mode passes the flag — `packages/happy-cli/src/claude/runClaude.ts:487`. **Local mode does not**: `packages/happy-cli/src/claude/claudeLocalLauncher.ts:35` and `:100` both call `scanner.onNewSession(sessionId)` bare. Local is the default mode for `happy resume`, so the path most people hit is the unfixed one.

## The change

One file, +11/-4:

- derive `isReconnect` from `HAPPY_RECONNECT_SESSION_ID` — the env var upstream's own `buildReconnectEnv` already sets (`packages/happy-cli/src/resume/handleResumeCommand.ts:108-115`, spawned at `:170`)
- pass `{ treatExistingAsProcessed: isReconnect }` at both `onNewSession` call sites

No new API and no behaviour change outside reconnect: when the env var is absent `isReconnect` is `false` and both calls are byte-equivalent to what they do today. It brings local mode to the parity `runClaude.ts:487` has had all along.

## Why this PR shrank

It originally also wired `happy resume` into the reconnect path itself (`handleResumeCommand.ts` + 6 test cases, ~93 lines). Upstream implemented that in `53850917`, so that part has dropped out of the diff and only the local-mode gap remains.

The BEFORE/AFTER capture that used to be in this description measured the *session-minting* behaviour (2 rows → 1). That is upstream's change now, not this one, so I have removed it rather than leave proof of a different fix in place. Happy to capture a fresh transcript-duplication pair if that would help review.

### DIFF
--- a/packages/happy-cli/src/claude/claudeLocalLauncher.ts
+++ b/packages/happy-cli/src/claude/claudeLocalLauncher.ts
@@ -12,18 +12,25 @@ export async function claudeLocalLauncher(session: Session): Promise<LauncherRes
     const scanner = await createSessionScanner({
         sessionId: session.sessionId,
         workingDirectory: session.path,
-        onMessage: (message) => { 
+        onMessage: (message) => {
             // Block SDK summary messages - we generate our own
             if (message.type !== 'summary') {
                 session.client.sendClaudeSessionMessage(message)
             }
         }
     });
-    
+
+    // When this CLI is reattaching to an existing Happy session (`happy resume`
+    // or daemon-driven resume), the Happy server already holds every prior
+    // message. `claude --resume` mints a fresh JSONL with the entire history
+    // copied in, so without this flag the scanner would re-forward every
+    // historical user/assistant message and double the transcript.
+    const isReconnect = !!process.env.HAPPY_RECONNECT_SESSION_ID;
+
     // Register callback to notify scanner when session ID is found via hook
     // This is important for --continue/--resume where session ID is not known upfront
     const scannerSessionCallback = (sessionId: string) => {
-        scanner.onNewSession(sessionId);
+        scanner.onNewSession(sessionId, { treatExistingAsProcessed: isReconnect });
     };
     session.addSessionFoundCallback(scannerSessionCallback);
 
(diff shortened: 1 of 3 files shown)


chphch:
@bra1nDump — one more from the batch, with a correction of mine up front: this PR changes behavior that `36b23d57` chose deliberately — its commit message says *"The offline-reconnect path in runClaude.ts and the local launcher keep the existing forward-everything semantics."* — and my description never acknowledged that. Quoting it now because the condition it rested on has since changed: when that was written, `happy resume` in local mode minted a fresh Happy session, so forwarding the whole JSONL was exactly what populated it. Since `53850917` (#1463), resume re-attaches to the existing session — which already holds those messages — so forward-everything now doubles the transcript on every local-mode reconnect, and local is the default mode for `happy resume`.

The change is gated accordingly: `treatExistingAsProcessed` is passed only when `HAPPY_RECONNECT_SESSION_ID` is set, so a fresh local session keeps the forward-everything semantics `36b23d57` chose; only the re-attach path pre-marks, matching what remote mode already does at `runClaude.ts:487`. +11/−4, one file. Re-verified today: mergeable clean, smoke tests green on linux/windows × node 20/24. If keeping forward-everything on local reconnect was deliberate even post-`53850917`, happy to close.

PR #1323, opened 2026-05-23
fix(app): preserve chat scroll position when returning to a session

chphch:
## Summary

Navigating away from a chat session (back button) and then back into it dropped the user at the visual bottom of the chat, losing any scroll position they had set while reading older messages. `SessionView` mounts `ChatList` with `key={sessionId}` inside expo-router's Stack, so the back-then-forward path fully unmounts the component and the inverted `FlatList`'s scroll state is lost.

This PR caches the chat scroll offset per `sessionId` at module scope (in-memory only, survives the unmount/remount cycle but not app restarts) and restores it on the first `onContentSizeChange` after content lays out. The restore runs once per mount — subsequent size changes (new messages streaming in, lazy-loaded older pages) keep flowing to `FlatList`'s native `maintainVisibleContentPosition` exactly as before.

Touches only `packages/happy-app/sources/components/ChatList.tsx`. No new deps, no API/sync/encryption changes, no persistent storage.

Proof video/screenshots are posted in a follow-up comment.

### DIFF
--- a/packages/happy-app/sources/components/ChatList.tsx
+++ b/packages/happy-app/sources/components/ChatList.tsx
@@ -191,21 +191,35 @@ const ChatListInternal = React.memo((props: {
         flatListRef.current?.scrollToOffset({ offset: 0, animated: true });
     }, []);
 
-    // Restore the user's prior scroll position on the FIRST onContentSizeChange
-    // that fires after this component has actual content to lay out. The
-    // FlatList renders at offset 0 (visual bottom of an inverted list) by
-    // default, which is wrong when the user was reading older messages and
-    // came back to this session from elsewhere. We only restore once per
-    // mount — subsequent content-size changes (incoming messages, lazy-load
-    // pages) are left to FlatList's native maintainVisibleContentPosition.
+    // Restore the user's prior scroll position after this component has
+    // actual content to lay out. The FlatList renders at offset 0 (visual
+    // bottom of an inverted list) by default, which is wrong when the user
+    // was reading older messages and came back to this session from
+    // elsewhere.
+    //
+    // Why this is NOT a fire-once-on-first-change: inverted FlatList renders
+    // a windowed slice of items on mount, so the first `onContentSizeChange`
+    // can report a content height much smaller than the cached offset. If we
+    // marked ourselves "restored" on that first fire, `scrollToOffset` would
+    // get clamped to the current (small) content height and subsequent
+    // content growth (virtualizer expanding the window, lazy older pages)
+    // would never re-trigger the restore. So we keep retrying — and keep
+    // clamping intentionally so FlatList renders more items — until content
+    // height has caught up to the cached offset. Then we mark restored and
+    // stop. Subsequent size changes (incoming live messages) are left to
+    // FlatList's native maintainVisibleContentPosition.
     const hasRestoredRef = React.useRef(false);
-    const handleContentSizeChange = useCallback(() => {
+    const handleContentSizeChange = useCallback((_w: number, h: number) => {
         if (hasRestoredRef.current) return;
         if (displayItems.length === 0) return;
         const cached = sessionScrollOffsets.get(props.sessionId);
-        hasRestoredRef.current = true;
-        if (cached !== undefined && cached > 0) {
-            flatListRef.current?.scrollToOffset({ offset: cached, animated: false });
+        if (cached === undefined || cached <= 0) {
+            hasRestoredRef.current = true;
+            return;
+        }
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — scroll position preserved across back-then-forward navigation

Captured end-to-end via standalone `happy-server` (PGlite, port 3055) + Expo web (port 8088), authenticated through the `?dev_token=...&dev_secret=...` URL bypass, driven headlessly with Playwright. For each side (BEFORE = upstream `main` unmodified, AFTER = this PR's fix applied):

1. Create a session, send 25 short messages to make the chat scrollable.
2. Scroll up to roughly the middle of the conversation (offset 500 on the inverted FlatList).
3. `history.back()` → sessions list.
4. Re-open the same session and capture the chat.

The top row of the image shows step 2 (both runs look identical — chat scrolled to the middle). The bottom row shows step 4 after re-entering the session:

- **BEFORE** (left, red border) — chat snaps to the visual bottom; scroll position lost.
- **AFTER** (right, green border) — chat stays at the user's previous scroll position; same messages remain visible.

scroll-preserve

chphch:
### Follow-up fix — restore wasn't holding through FlatList virtualizer warmup

Dogfooding revealed the previous commit's restore would *sometimes* still drop you at the bottom when switching sessions A → B → A (the same path the original PR was trying to fix).

**Root cause:** the `hasRestoredRef` flag was flipped on the **first** `onContentSizeChange` after content existed. For an inverted FlatList that first fire happens while the virtualizer has only laid out the bottom window of items, so `contentHeight` can be much smaller than the cached scroll offset. `scrollToOffset(cached)` gets silently clamped (viewport stays at/near the visual bottom) — and now `hasRestoredRef === true` prevents the restore from ever being retried once the virtualizer expands and the full content height becomes reachable.

**Fix (commit [`29347​19e`](https://github.com/chphch/happy/commit/2934719e)):** keep retrying instead of one-shotting it.

- Pull the actual content height `h` out of `onContentSizeChange(_w, h)`.
- While `h < cached`, call `scrollToOffset(cached)` on **every** size change — the clamped scroll itself nudges FlatList to render more items toward the target offset.
- Only mark restored once `h >= cached`, i.e. once the virtualizer has actually laid out enough content for the target position to exist.
- The `cached <= 0` / `cached === undefined` paths still resolve immediately on first fire, so fresh sessions don't pay any extra fires.

Typecheck still green; rebuilt against my self-hosted webapp (`chphch:custom`) and the A → B → A symptom no longer reproduces with 100+ message sessions.

chphch:
## Follow-up fix — cache poisoning during restore loop

After deploying the initial fix to my self-hosted instance, the bug still reproduced: switching away from a session and back still snapped the viewport to the bottom on long chats.

Root cause: the restore loop in \`handleContentSizeChange\` calls \`scrollToOffset(cached)\` repeatedly until the FlatList content height has grown to match the cached offset. Each programmatic scroll fires \`handleScroll\` with the **clamped** offset (capped to current content height, which is still small at that point).

\`handleScroll\` was unconditionally writing \`offsetY\` into \`sessionScrollOffsets\`, so the cache got overwritten with the clamped value. The next retry read the smaller cached value, the clamp got worse on each iteration, and on long sessions the restore landed much closer to the bottom than where the user was actually reading.

Fix in c78a7a97: gate the cache write on \`hasRestoredRef.current\`. While the restore loop is still running, scroll events triggered by our own \`scrollToOffset\` are ignored for cache-write purposes. The user's real scroll events (after restore completes) continue to update the cache normally.

Verified on my self-hosted webapp build (custom branch HEAD \`2bb0ae48\`).

chphch:
Closing this one — withdrawing it rather than rebasing, because the approach didn't survive on my own build.

All three commits here touch the **native** `ChatList.tsx`: a module-scope `Map<sessionId, scrollTop>` plus a restore loop that retries until content height catches up. Running it day to day, that restore loop regressed three times — it fought the list's own scroll anchoring, and each fix moved the failure rather than removing it. I ended up deleting the native half entirely in favour of the stock MVCP (`maintainVisibleContentPosition`) config that `main` already ships, which covers the stick-to-bottom case without a cache to keep coherent.

What I did keep is the **web** path (`ChatList.web.tsx`), where there's no MVCP equivalent and the cache genuinely earns its place. That's a different file and a different fix from what's in this PR, so it isn't a rebase of this — if it's wanted upstream I'd rather open it as its own small, web-only PR.

`main` has also moved a fair way underneath this branch (the props here — `props.sessionId` — no longer exist), so what's left isn't a conflict resolution so much as a rewrite of an approach I no longer run. Not worth your review time.

Sorry for the noise, and thanks for leaving it open this long.

PR #1335, opened 2026-05-26
feat(cli): import & resume local Claude Code sessions in the Happy app

hustc5622:
## Motivation

Right now the only way for a session to show up in the Happy mobile app is to start it via `happy claude`. Sessions a user kicked off with raw `claude` are entirely invisible: their JSONL transcript on disk has no corresponding `Session` row on the server, no `metadata.claudeSessionId`, and no daemon-side tracking — so the app can't list them, can't show their history, and can't resume them.

This PR adds `happy import`, a CLI command that scans `~/.claude/projects/<dir>/<uuid>.jsonl`, registers each leaf session as a Happy `Session` on the server (encrypted client-side as usual), backfills the message history, and teaches the daemon to *adopt* those sessions when the mobile app sends `resume-happy-session` — so a fresh `happy claude --resume <claudeSessionId>` is spawned and the conversation continues from where the user left off in the terminal.

App and server are unchanged: this is a CLI-only contribution. The mobile Resume button works because it gates only on `metadata.machineId` + `metadata.claudeSessionId` being present (both populated by the importer), and the `expResumeSession` feature flag.

## What's in this PR

Two commits on top of `main`:

1. **`feat(import): import local Claude Code sessions into Happy`** — the `happy import` command and supporting modules.
2. **`feat(daemon): adopt imported sessions on resume-happy-session RPC`** — the daemon-side fallback that lets the existing resume pipeline find sessions in the import journal when they're not in `pidToTrackedSession` / `sessionIdToFinishedSession`.

A separate auto-sync polling loop on the daemon's heartbeat exists in our personal branch but is intentionally **not** included here. It changes the daemon's default behavior and would benefit from its own discussion / opt-in flag design — happy to break it out as a follow-up PR if maintainers are interested.

## Architecture

\`\`\`
~/.claude/projects/*/*.jsonl
        │
        ▼  scanner.ts  (filter ancestors via summary.leafUuid chain,
        │               filter happy-tracked claudeSessionIds, project filter, mtime filter)
        ▼
   sessionImporter.ts
        ├── ApiClient.getOrCreateSession({ tag: \`claude-imported:<id>\`, metadata, state })
        ├── ApiSessionClient + sendClaudeSessionMessage  ← reuses
        │     mapClaudeLogMessageToSessionEnvelopes for parity with live sessions
        ├── api.deactivateSession(id)         ← active=false, lifecycleState='running'
        └── upsertEntry → ~/.happy/imported-sessions.json (encryption material captured)

Mobile clicks Resume
        │
        ▼  resume-happy-session RPC → daemon
        ▼
   findTrackedSessionById  (pidToTrackedSession ∪ sessionIdToFinishedSession)
        │ miss
        ▼
   adoptSessionFromImportJournal  ← new fallback: synthesize TrackedSession from journal
        ▼
   buildResumeLaunch + spawn happy claude --resume <claudeUuid> + HAPPY_RECONNECT_*
\`\`\`

## File-by-file

**New (all in `packages/happy-cli/src/import/`):**
- `scanner.ts` — walk `~/.claude/projects`, skip non-UUID names, drop ancestors of resume chains, apply filters
- `jsonlParser.ts` — streaming reader + cheap header read for cwd/sessionId/title
- `importJournal.ts` — atomic lock+rename writes to `~/.happy/imported-sessions.json`
- `messageBackfill.ts` — pushes JSONL lines through the existing session-protocol mapper
- `sessionImporter.ts` — per-candidate orchestration (create → backfill → deactivate → journal)
- `collectHappyTrackedClaudeIds.ts` — server-side dedup: decrypts each persisted session's metadata locally to learn its claudeSessionId
- `daemonAdopt.ts` — synthesize TrackedSession from journal entry for the resume fallback
- `unarchiveImported.ts` — \`--fix-archived\` for journals written before the lifecycleState rule was nailed down
- `pruneImported.ts` — \`--prune-older-than-days N\` to walk back accidental imports

**New (commands):**
- `packages/happy-cli/src/commands/importCommand.ts` — argv parsing + subflag routing

**Modified:**
- `packages/happy-cli/src/index.ts` — registers \`import\` subcommand
- `packages/happy-cli/src/daemon/run.ts` — adopt fallback in \`resumeSession\` (only)

## CLI surface

\`\`\`
happy import [--dry-run] [--yes] [--project <path>] [--days <n>] [--no-backfill] [--limit <n>]
happy import --fix-archived
happy import --prune-older-than-days <n> [--dry-run]
\`\`\`

All operations are end-to-end encrypted client-side as usual; the server only sees opaque ciphertext.

## Tests

32 new unit tests (passing locally, full suite passes):
- \`scanner.test.ts\` (9): leaf-detection chain, projectFilter, journal/server-tracked exclusions, mtime filter, sort
- \`jsonlParser.test.ts\` (10): schema passthrough, lenient invalid-line skip, header extraction, mtime stat
- \`importJournal.test.ts\` (9): atomic round-trip, corrupted-file recovery, schema-mismatch quarantine, dedup by claudeSessionId
- \`daemonAdopt.test.ts\` (3): null on miss, dataKey + legacy variant key reconstruction
- \`autoImportPoll.test.ts\` (4) — *not in this PR but referenced for completeness*

## Notes for reviewers

- **Encryption-key sourcing**: the journal stores per-session keys captured at import time so the daemon's adopt path doesn't need agent.key to decrypt the server metadata. This makes adopt work for users without happy-agent set up. Trade-off: an attacker with read access to \`~/.happy/imported-sessions.json\` learns those keys; same threat model as \`~/.happy/sessions.json\`, which already stores them for daemon-spawned sessions.
- **Tag format**: \`claude-imported:<claudeSessionId>\` makes server-side \`(accountId, tag)\` dedup work even if the local journal is lost — re-running \`happy import\` won't create duplicate Session rows on the server.
- **\`lifecycleState: 'running' + active: false\`** is the explicit choice for imported sessions: \`lifecycleState: 'archived'\` makes the mobile app lock the input box without offering a resume affordance.
- **Resume chains**: when a user runs \`claude --resume X\` natively after import, Claude SDK creates a new JSONL with a new UUID prefixing X's history. The scanner correctly identifies this new file as the leaf via its top-of-file \`{type:"summary",leafUuid:X}\` line, marks X as an ancestor, and skips X on the next scan.

Happy to split this into smaller PRs or rework anything that doesn't fit the project's direction.

### DIFF
--- a/packages/happy-cli/src/commands/importCommand.ts
+++ b/packages/happy-cli/src/commands/importCommand.ts
@@ -1,0 +1,311 @@
+/**
+ * `happy import` command
+ *
+ * Scans ~/.claude/projects/ for native Claude Code sessions that haven't been
+ * imported into Happy yet, creates a happy session per leaf JSONL (mirroring
+ * its message history into happy-server), and records the result in
+ * ~/.happy/imported-sessions.json so the daemon can later adopt-and-resume
+ * them from the mobile app.
+ *
+ * Flags:
+ *   --dry-run             Show what would be imported, don't do anything
+ *   --yes / -y            Skip the y/N confirmation prompt
+ *   --project <path>      Only import sessions whose cwd starts with <path>
+ *   --no-backfill         Create the happy session but don't push messages
+ *   --limit <n>           Cap how many candidates are processed in one run
+ *
+ * Exit codes: 0 success / partial-success, 1 on hard error.
+ */
+
+import chalk from 'chalk';
+import inquirer from 'inquirer';
+import { resolve as resolvePath } from 'node:path';
+
+import { ApiClient } from '@/api/api';
+import { authAndSetupMachineIfNeeded } from '@/ui/auth';
+
+import { collectHappyTrackedClaudeSessionIds } from '../import/collectHappyTrackedClaudeIds';
+import { pruneImportedSessionsOlderThan } from '../import/pruneImported';
+import { scanForImportCandidates } from '../import/scanner';
+import { importSingleSession, type ImportSessionResult } from '../import/sessionImporter';
+import { unarchiveAllImported } from '../import/unarchiveImported';
+
+type ParsedArgs = {
+    dryRun: boolean;
+    yes: boolean;
+    project?: string;
+    backfill: boolean;
+    limit?: number;
+    days?: number;
+    fixArchived: boolean;
+    pruneOlderThanDays?: number;
+    showHelp: boolean;
+};
+
+function parseArgs(args: string[]): ParsedArgs {
+    const result: ParsedArgs = {
+        dryRun: false,
+        yes: false,
+        backfill: true,
+        fixArchived: false,
+        showHelp: false,
+    };
+    for (let i = 0; i < args.length; i++) {
+        const arg = args[i];
+        switch (arg) {
+            case '-h':
+            case '--help':
+                result.showHelp = true; break;
+            case '--dry-run':
+                result.dryRun = true; break;
+            case '-y':
+            case '--yes':
+                result.yes = true; break;
+            case '--no-backfill':
+                result.backfill = false; break;
+            case '--project':
+                result.project = args[++i]; break;
+            case '--limit': {
+                const n = parseInt(args[++i] ?? '', 10);
+                if (Number.isFinite(n) && n > 0) result.limit = n;
+                break;
+            }
+            case '--days': {
+                const n = parseInt(args[++i] ?? '', 10);
+                if (Number.isFinite(n) && n > 0) result.days = n;
+                break;
+            }
+            case '--fix-archived':
+                result.fixArchived = true; break;
+            case '--prune-older-than-days': {
+                const n = parseInt(args[++i] ?? '', 10);
+                if (Number.isFinite(n) && n > 0) result.pruneOlderThanDays = n;
+                break;
+            }
+            default:
+                throw new Error(`Unknown argument: ${arg}`);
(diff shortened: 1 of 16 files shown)


SiyaoZheng:
Real user here — this is exactly the missing piece in my Happy workflow: I often have existing Claude Code sessions on my Mac and later want to discover/resume them from Happy Mobile.

I checked the current `main`, and it looks like this PR can now be made substantially smaller. The newer resume path already lets the mobile client send a fallback payload containing the session metadata + per-session data key (`buildResumeFallback()` in `happy-app/sources/sync/ops.ts`), and `resumeSessionAttempt()` in the daemon can revive a Happy session that the daemon has never tracked before.

So on current main, the `imported-sessions.json` + `adoptSessionFromImportJournal()` daemon path appears to be unnecessary for the core feature. A much smaller rebase could probably be:

1. scan `~/.claude/projects/*/*.jsonl` for leaf Claude sessions;
2. create a Happy session with `machineId`, `path`, `flavor: 'claude'`, and `claudeSessionId`;
3. backfill the transcript into that Happy session;
4. let the existing mobile Resume flow handle revival through the current fallback mechanism.

That would also make it possible to defer the journal/prune/`--fix-archived` machinery to a follow-up and reduce the review surface considerably. The current PR is now non-mergeable against main, so a rebase + scope reduction may be the shortest route to getting the core capability shipped.

Happy to test the rebased version end-to-end on current macOS + iOS if useful.

PR #1341, opened 2026-05-28
perf(chat): swap FlatList for FlashList in ChatList

chphch:
## Summary

Swap `FlatList` → `@shopify/flash-list` v2 in `ChatList.tsx`. Net diff: -15/+9, single file.

## Motivation

Two cherry-picked perf PRs already land on `chphch/happy:custom` (my self-host daily-driver) — #1245 (re-render isolation) and #1242 (lazy-load + parallel decrypt). Despite these, **clicking into a long session on web (~hundreds of messages) still blocks the paint for multiple seconds.**

#1154 ("perf(chat): fix UI freeze entering large chats on tablets") diagnosed the same root cause for the tablet path and shipped a band-aid (`initialNumToRender=8`, `windowSize=5`, `removeClippedSubviews=true`, etc.). That PR's body explicitly notes:

> `FlashList` would be a better long-term fit for inverted chat lists with variable heights. Also out of scope for this fix.

This PR is that long-term fix.

## Why FlatList tuning props don't help on web

`react-native-web`'s `FlatList` virtualizer does **not** honor `initialNumToRender` / `maxToRenderPerBatch` / `windowSize` / `removeClippedSubviews` the way native does. With `inverted=true` + `maintainVisibleContentPosition` + variable-height messages and no `getItemLayout`, the web shim mounts essentially the full data set on first render to measure heights and compute the inverted scroll anchor. So even with #1154's tuning props the web entry stays slow — it's a fundamental virtualizer mismatch, not a config gap.

## Why FlashList v2

`@shopify/flash-list` v2's RecyclerView engine has first-class web virtualization, supports inverted lists, and exposes the same imperative API we already depend on (`scrollToOffset`, `getScrollableNode`) — so the scroll-to-bottom button and the macOS Shift+wheel handler keep working unchanged.

`@shopify/flash-list` is already a dependency — used by `packages/happy-app/sources/app/(app)/dev/inverted-list.tsx` as a perf experiment page. No `package.json` change.

## Changes

- Replace `FlatList` import + ref type with `FlashList` / `FlashListRef`.
- MVCP semantics change: FlashList v2 inverted uses `autoscrollToTopThreshold` to auto-stick to newest content; `minIndexForVisible: 1` is dropped (FlashList's anchoring is internal).
- Drop unused-by-FlashList tuning props: `scrollEventThrottle={16}` (FlashList manages this).

## Testing

**Dogfood-validated on web:** built `chphch/happy:custom` with this swap, deployed to my self-host (`https://happy.bingsucouple.com`), opened a session with hundreds of messages — **noticeably faster paint, scroll/MVCP behave correctly.** No regressions in basic chat flow, scroll-to-bottom, or new-message auto-stick.

- [x] `pnpm --filter happy-app typecheck` — clean
- [x] Web dogfood on self-host (long session click → fast paint, MVCP sticks on incoming, scroll-to-bottom works)
- [ ] Native iOS / Android verification — **not yet** (I don't have an Android e-ink tablet to verify the #1154 scenario; happy to defer to anyone who does)
- [x] Proof GIF (smoke — chat renders correctly with FlashList) — see comment below

## Interaction with #1154

Both PRs target the same code path. If #1154 lands first, this PR rebases trivially: every tuning prop that #1154 adds (`initialNumToRender`, `maxToRenderPerBatch`, `windowSize`, `removeClippedSubviews`, `scrollEventThrottle=32`) becomes a no-op under FlashList and can be dropped in the rebase.

### DIFF
--- a/packages/happy-app/sources/components/ChatList.tsx
+++ b/packages/happy-app/sources/components/ChatList.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { useSession, useSessionMessages, useSetting } from "@/sync/storage";
 import { sync } from '@/sync/sync';
-import { ActivityIndicator, FlatList, NativeScrollEvent, NativeSyntheticEvent, Platform, Pressable, View } from 'react-native';
+import { ActivityIndicator, NativeScrollEvent, NativeSyntheticEvent, Platform, Pressable, View } from 'react-native';
+import { FlashList, FlashListRef } from '@shopify/flash-list';
 import { useCallback } from 'react';
 import { useHeaderHeight } from '@/utils/responsive';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
@@ -66,7 +67,7 @@ const ChatListInternal = React.memo((props: {
     isLoadingOlder: boolean,
 }) => {
     const { theme } = useUnistyles();
-    const flatListRef = React.useRef<FlatList>(null);
+    const flatListRef = React.useRef<FlashListRef<DisplayItem>>(null);
     const [showScrollButton, setShowScrollButton] = React.useState(false);
     // Tracks whether the scroll-button is currently shown, so we only call
     // setShowScrollButton when the threshold is actually crossed instead of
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — dogfooded on self-host

I'm posting this as a written proof rather than a GIF because the perf improvement is only observable on production-scale data (1000+ message sessions). A capture against a fresh PGlite test account would show the FlashList wired up but wouldn't demonstrate the actual perf delta — the bottleneck only appears under load.

**Deployed to**: `https://happy.bingsucouple.com` (self-hosted production build of the `chphch/happy` fork)
**Image**: `happy-app:50423b1890bd` (image SHA `49d45218`) — `main` + this PR's commit cherry-picked on top of `chphch:custom`
**Build date**: 2026-05-28 ~12:30 KST

### What I observed

- **Scrolling through a long inverted chat list is noticeably smoother.** Before this change, scrolling up through ~1000+ messages on the web build had visible jank and the auto-stick-to-bottom behavior fought with `maintainVisibleContentPosition` during streaming agent output. After the swap, scroll is fluid and MVCP stick-to-bottom works correctly when new tokens stream in.
- **No regressions in scroll-restore.** The existing `Map<sessionId, scrollTop>` restore-on-remount loop (from #1323) still works correctly under FlashList — `scrollToOffset`, `getScrollableNode`, and `onContentSizeChange` all behave the same. The `autoscrollToTopThreshold: 50` translation of FlatList's `MVCP.minIndexForVisible: 1 + autoscrollToTopThreshold: 10` preserves stick-to-bottom semantics.
- **No crashes / no React tree errors** during normal use across multiple sessions over ~3 hours of dogfood.

### Caveat (filed separately)

Opening a long session for the first time still feels slow — even though FlashList virtualization means only ~20 items mount on first paint. The bottleneck is upstream of the list (likely the background prefetch loop in `sync.ts:1859` storming the zustand store, plus sync decrypt on the JS thread). I've filed this as a separate issue: #1342. **This PR is intentionally narrow** — it only addresses the scroll-perf part of the problem.

### Typecheck

`pnpm --filter happy-app typecheck` is clean. Single-file change to `packages/happy-app/sources/components/ChatList.tsx` (-15/+9).

chphch:
## Proof — FlashList `ChatList` renders + scrolls

Captured locally (standalone `happy-server` + Expo web, Playwright). Renders the real `ChatList` — now backed by `@shopify/flash-list` — with 60 messages:

FlashList ChatList render and scroll

Inverted layout with newest pinned at the bottom; scrolling up reveals older messages and scrolling back sticks to newest. Message rows (user bubbles, agent text, tool cards) render correctly after the `FlatList`→`FlashList` swap, and the v2 MVCP (`autoscrollToTopThreshold`) keeps the viewport anchored to newest content.

chphch:
Closing — we've decided not to pursue the FlashList swap; upstream's maintainVisibleContentPosition handling covers the stick-to-bottom behavior this targeted. Happy to revisit if scroll perf regressions resurface.

PR #1347, opened 2026-05-31
feat: fetch full model list from gateway for app model picker

MysticCoss:
Summary

Mobile app's model picker only showed 4 hardcoded Claude aliases (default, opus, sonnet, haiku) — even when connected to a gateway with 100+ available models (using free-claude-code for example). This fetches the full catalog from the
gateway's /v1/models at session start so the app displays every model the gateway exposes: provider models, 1M-context variants, everything. When no gateway is configured, it falls back to the existing
hardcoded list plus SDK-reported models.

What changed

- fetchAvailableModels.ts (new) — GET /v1/models from ANTHROPIC_BASE_URL, merges with 4 hardcoded aliases, deduplicates, falls back on any error
- runClaude.ts — applies gateway env vars early, seeds metadata.models with the full list, propagates currentModelCode on model selection
- claudeRemote.ts — adds onModelsFound callback from supportedModels()
- claudeRemoteLauncher.ts — merges SDK models as fallback enrichment

Proof

Tested end-to-end with a gateway that exposes 135+ models. Before: 4 models in the picker. After: all gateway models appear and are selectable. Fallback path works correctly with ANTHROPIC_BASE_URL unset.

Behavior

| ANTHROPIC_BASE_URL | Models shown                     |
|--------------------|----------------------------------|
| Not set            | 4 hardcoded + SDK models         |
| Gateway configured | 4 aliases + full gateway catalog |

### DIFF
--- a/packages/happy-cli/src/claude/claudeRemote.ts
+++ b/packages/happy-cli/src/claude/claudeRemote.ts
@@ -46,7 +46,9 @@ export async function claudeRemote(opts: {
     onSessionReset?: () => void,
     onSDKMetadata?: (metadata: { tools?: string[]; slashCommands?: string[]; mcpServers?: { name: string; status: string }[]; skills?: string[] }) => void,
     /** Per-turn plan rate-limit delta; the launcher merges it into agent state. */
-    onUsageLimits?: (patch: UsageLimitsPatch) => void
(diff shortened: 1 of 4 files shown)


MysticCoss:
@bra1ndump hello, this is a simple but powerful improvement that makes intergration with free-claude-code smooth. Please review it

PR #1355, opened 2026-06-02
fix(cli): only archive a session on explicit Archive, not on every exit

### LINKED ISSUE #1354: Sessions archived on every CLI exit (Ctrl-C / SIGTERM / crash), hiding them and trapping resume
## Summary

A Happy session is stamped `lifecycleState='archived'` on **every** CLI exit — not just when the user explicitly archives it. This hides the session from the active list and makes it effectively unrecoverable, because the archived metadata is replayed to the client on reconnect and bounces the resume straight back out.

## Where

`packages/happy-cli/src/claude/runClaude.ts` — the `cleanup()` handler:

```ts
const cleanup = async () => {
    // ...
    session.updateMetadata((m) => ({ ...m, lifecycleState: 'archived', ... }));  // ALWAYS
    // ...
};
process.on('SIGTERM', cleanup);
process.on('SIGINT', cleanup);                              // Ctrl-C
process.on('uncaughtException', (e) => { ...; cleanup(); });
process.on('unhandledRejection', (r) => { ...; cleanup(); });
registerKillSessionHandler(session.rpcHandlerManager, cleanup);
```

Every termination path — `SIGTERM`, `SIGINT` (a plain Ctrl-C), `uncaughtException`, `unhandledRejection`, and the explicit Archive RPC — funnels through the same `cleanup()`, which unconditionally archives.

## Impact

1. **Ctrl-C archives the session.** Quitting a CLI session the normal way hides it from the app's active list, forcing the user to dig it up to resume.
2. **A transient crash during resume permanently traps the session.** If the resume/reconnect path throws (`uncaughtException`/`unhandledRejection`), the session is archived. On the next resume attempt the client receives the archived metadata in the `update-session` payload and exits again (see the archived-state handling in `apiSession.ts`), so the session can never come back through the normal resume button. Observed in practice: a "continue" on a dead session repeatedly archived it instead of resuming.

The root confusion is conflating **`active`** (is the process live right now?) with **`lifecycleState`** (has the user retired this session?). Process death should set `active=false`; only the user choosing "Archive" should set `lifecycleState='archived'`.

## Proposed fix

Gate the archive metadata write behind an `archive` flag that defaults to `false`. Only the explicit in-app Archive button (`killSession` RPC) passes `archive: true`. `SIGTERM` / `SIGINT` / crashes tear down and let the session go `active=false` (via the normal session-death signal) but leave `lifecycleState` untouched — same "inactive but resumable" semantics as a network blip.

PR attached.

## Complementary hardening (separate)

On reconnect/resume, ignore a stale `archived` signal once so an already-archived session can be brought back (a `suppressNextArchiveSignal()`-style guard in the reconnect path). Out of scope for this minimal fix.

chphch:
Fixes #1354.

## Problem

`runClaude`'s `cleanup()` unconditionally stamped `lifecycleState='archived'`, and **every** exit path called it — `SIGTERM`, `SIGINT` (Ctrl-C), `uncaughtException`, `unhandledRejection`, and the Archive RPC. So any process death — a plain Ctrl-C, or a transient crash during resume/reconnect — hid the session from the active list. And because the archived metadata is replayed to the client on reconnect (`update-session` handling in `apiSession.ts`), the next resume bounced straight back out, leaving the session unrecoverable from the app.

The underlying confusion: **`active`** (process live now?) vs **`lifecycleState`** (user retired this session?). Process death should only flip `active`; only the user choosing "Archive" should set `lifecycleState='archived'`.

## Change

Gate the archive-metadata write behind an `archive` flag (default `false`):

- `SIGTERM` / `SIGINT` / `uncaughtException` / `unhandledRejection` → `cleanup({ archive: false })`: tear down + send session-death (so `active=false`) but leave `lifecycleState` alone. The session stays visible and resumable, like a network blip.
- The in-app Archive button (`killSession` RPC) → `cleanup({ archive: true })`: the one path that retires the session.
- The `session.on('archived')` handler (server already archived from web/mobile) → `cleanup({ archive: false })`: just tear down; no need to re-stamp.

One file, `packages/happy-cli/src/claude/runClaude.ts`. Typechecks clean.

## Notes

- A complementary improvement — ignoring a stale `archived` signal once on reconnect so an *already-archived* session can be revived (a `suppressNextArchiveSignal()`-style guard) — is intentionally out of scope here; this PR removes the source of unwanted archiving.

### DIFF
--- a/packages/happy-cli/src/claude/runClaude.ts
+++ b/packages/happy-cli/src/claude/runClaude.ts
@@ -285,10 +285,12 @@ export async function runClaude(credentials: Credentials, options: StartOptions 
     let currentAllowedTools: string[] | undefined = undefined; // Track current allowed tools
     let currentDisallowedTools: string[] | undefined = undefined; // Track current disallowed tools
     let currentRunMode: 'local' | 'remote' = options.startingMode ?? 'local';
-    // Exit when session is archived from web/mobile
+    // Exit when session is archived from web/mobile. The server already
+    // stamped lifecycleState='archived' (that's what fired this event), so we
+    // just tear down and exit — no need to re-archive.
     session.on('archived', () => {
         logger.debug('[loop] Session archived from web/mobile, cleaning up...');
-        cleanup();
+        void cleanup({ archive: false });
     });
 
     session.onUserMessage((message) => {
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — process exit no longer archives the session

Captured live end-to-end: standalone `happy-server` + a paired `happy-cli` daemon spawned a session via the control API; the session process was then **SIGTERM**'d (a normal exit / Ctrl-C, not the in-app Archive).

no archive on process exit

The session stays in the active Terminals list ("last seen just now" — inactive but **not** archived), and the CLI logs `[START] Cleaning up (archive=false)` — the signal/crash handlers now take the no-archive path. Only the explicit in-app Archive (killSession RPC → `cleanup({ archive: true })`) stamps `lifecycleState='archived'`, so a process death stays resumable instead of being hidden and bounced on the next resume.

chphch:
Closing — `main` has since adopted the same fix: `SIGTERM`/`SIGINT` now call `cleanup({ archive: false })` (see `runClaude.ts`), so a normal process exit no longer force-archives the session. This PR is superseded. Thanks!

PR #1363, opened 2026-06-06
feat: Opus 4.8 + dynamic workflows (agent-sdk 0.3.167)

jlixfeld:
Adds Opus 4.8 support and dynamic-workflow capability. Stacked on #1352
(fix/zod4-peer-dep) — depends on the zod 4 upgrade, since
@anthropic-ai/claude-agent-sdk >=0.2.119 requires zod ^4. Merge #1352 first
(or after it merges, retarget this to main).

Changes
- happy-cli: bump @anthropic-ai/claude-agent-sdk 0.2.96 -> ^0.3.167. 0.3.167
  is the first release exposing the Workflow tool and the
  enableWorkflows/workflowKeywordTriggerEnabled controls (Settings-level,
  plan-gated, keyword trigger on by default), so dynamic workflows become
  available with no per-query wiring.
- app: Claude model picker label 'opus 4.7' -> 'opus 4.8'.
- codium: Anthropic plugin adds claude-opus-4-8 as the latest Opus (4.7
  demoted to "previous").

Verification
- happy-cli typecheck + build (pkgroll) pass against 0.3.167; SDK imports
  cleanly at runtime; 56 focused cli unit tests pass.
- codium + app typecheck pass; app model-mode tests pass.
- NOT verified in-sandbox: the SDK's live session behavior (actual 4.8
  routing / Workflow execution) needs a real Claude environment (API key +
  claude-code binary).

PR #1372, opened 2026-06-09
feat: add Claude Fable 5 model

jlixfeld:
## What

Adds **Claude Fable 5** (`claude-fable-5`) to the Claude model pickers. Anthropic shipped Fable 5 as a GA model on 2026-06-09 — a Mythos-class, most-capable model for long-horizon agentic work (1M context, adaptive thinking always-on, no extended thinking).

## Changes

- **codium** `plugins/anthropic/index.ts`: add `claude-fable-5` to the full-ID `MODELS` list (top slot).
- **happy-app** `modelModeOptions.ts`: add `claude-fable-5` to `getClaudeModelModes()` fallback.
  - Fable 5 is a new family with **no short alias** (unlike `opus`/`sonnet`/`haiku`), so the picker sends the full model id; the Claude CLI passes it through to the API unchanged.
  - The live model list still comes from CLI session metadata when present — this only changes the offline fallback.
- **happy-app** `modelModeOptions.test.ts`: new test asserting the Claude fallback list and that the full id is used for Fable 5.

## Verification

- `modelModeOptions.test.ts` — 10 passed
- `happy-app` typecheck — pass
- `codium` typecheck — pass

Follows the same pattern as the Opus 4.8 addition (#1363).

### DIFF
--- a/packages/codium/sources/plugins/anthropic/index.ts
+++ b/packages/codium/sources/plugins/anthropic/index.ts
@@ -10,6 +10,7 @@ import type {
 const STORAGE_KEY = 'codium.plugin.anthropic.apiKey'
 
 const MODELS: ModelDescriptor[] = [
+    { id: 'claude-fable-5',    label: 'Fable 5',    group: 'Anthropic', description: 'Most capable model, long-horizon agentic work.' },
     { id: 'claude-opus-4-8',   label: 'Opus 4.8',   group: 'Anthropic', description: 'Latest Opus generation.' },
     { id: 'claude-opus-4-7',   label: 'Opus 4.7',   group: 'Anthropic', description: 'Previous Opus generation.' },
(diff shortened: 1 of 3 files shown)


chphch:
Small data point on the "no short alias" note — the `fable` short alias **does** exist in current Claude Code, but `claude-fable-5` is still the better choice, so I don't think anything needs to change here.

In the bundled `claude` binary (`2.1.170`) the model-alias set is:

```
["sonnet","opus","haiku","fable"]
```

and it maps `fable` → `claude-fable-5` (`fable:"claude-fable-5"`), same shape as the other three.

So on recent CLIs `fable` would resolve — but the alias was only added in a recent release, whereas the full id `claude-fable-5` works on older Claude Code versions too (and passes straight through to the API). Sending the full id from the picker is the more portable call. 👍

PR #1374, opened 2026-06-10
fix(app): unread sessions show a bold title when stopped, not an overloaded status dot

chphch:
The session-list status dot is overloaded: a session with unread results is forced to a solid blue dot — the same color as the "thinking" (running) state, differing only by a pulse — so a finished, idle session looks like it is still running until you open it (which clears the unread flag).

This decouples the two signals. The status dot always reflects true liveness (disconnected / thinking / waiting / permission), and unread is shown instead as a **bold title** — but only once the agent has **stopped**, never while it is still running, so a re-activated session never reads as unread mid-turn. Blue goes back to meaning exactly one thing: running. Applies to both the session list rows and the compact active-sessions group.

The bold is applied via the SemiBold font *family* (read titles are Regular) rather than a numeric `fontWeight`, because the web build sets `font-synthesis: none` and bundles no Bold(700) IBM Plex Sans face — a `fontWeight: 700` override would render identically to the regular title on web.

## Proof

Captured locally on web — standalone `happy-server` (PGlite) + Expo web, driven headlessly with Playwright. The row's computed `font-family` flips Regular → SemiBold at the exact moment the turn stops and the session becomes unread; it stays Regular the whole time the agent is running.

| agent running → title **not** bold | turn stopped + unread → title **bold**, no marker dot |
|:---:|:---:|
|  |  |

Note the leading status dot stays a true-liveness indicator across the transition (blue "thinking" while running → neutral once idle), and the old trailing blue accent dot is gone — unread is conveyed entirely by the title weight.

## Verification

`pnpm typecheck` clean; app suite green.

### DIFF
--- a/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
+++ b/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
@@ -225,12 +225,13 @@ export function ActiveSessionsGroupCompact({ sessions, selectedSessionId }: Acti
 export const CompactSessionRow = React.memo(({ session, selected, showBorder }: { session: SessionRowData; selected?: boolean; showBorder?: boolean }) => {
     const styles = stylesheet;
     const { theme } = useUnistyles();
-    const baseStatus = STATUS_CONFIG[session.state];
+    const status = STATUS_CONFIG[session.state];
     const needsUserAction = session.state === 'permission_required' || session.state === 'input_required';
-    // User action stays orange and pulsing even when the request also marked the session unread.
-    const status = session.hasUnread && !needsUserAction
-        ? { ...baseStatus, color: '#007AFF', dotColor: '#007AFF', isPulsing: false, isConnected: baseStatus.isConnected }
-        : baseStatus;
+    // Unread is shown as a bold title only once the agent has stopped — never
+    // while it's still running (thinking), so a re-activated session doesn't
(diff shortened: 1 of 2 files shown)


chphch:
Rebased onto current `main` (`3cf9ce70f`) — no conflicts; the home rework (`c63c80ff3` / `6dc878dc3`) rebuilt which container renders which sessions but never entered `CompactSessionRow`'s status logic or `SessionItem`'s title weight, so the problem and the fix are both unchanged.

One import fix was needed for `tsc` to pass: upstream has since added `useSettingMutable` to the storage import at `SessionsList.tsx:5`, which duplicated the one this PR added. Merged `useSetting` into upstream's line and dropped the added one.

`pnpm typecheck` clean; app suite 919/919.

chphch:

Worth flagging that **#1641** (Aug 2) fixes the same collision independently: @leikaiwei hit the unread marker masking a session that needs attention, and their patch lands in the same two files as this one (`ActiveSessionsGroupCompact.tsx`, `SessionsList.tsx`). That PR came about seven weeks after this one, so the overload has been reachable in `main` the whole time.

The two take different routes — this one stops unread from repainting the status dot at all and moves the signal to a bold title, so `permission_required` / `input_required` keeps the dot to itself.

@bra1nDump — these two will conflict with each other whichever lands first. Happy to rework mine on top of #1641 if you'd rather take theirs; just say which.

Issue #1086, opened 2026-04-16
security(app): Mermaid SVG rendered via dangerouslySetInnerHTML without sanitization on web platform


shaun0927:
## Summary

PR #678 fixed the Mermaid XSS vulnerability on the native WebView path, but the **web platform** path remains unprotected. Mermaid SVG output is injected into the DOM via `dangerouslySetInnerHTML` without sanitization, and `mermaid.initialize()` does not set `securityLevel: 'strict'`.

## Problem

In `packages/happy-app/sources/components/markdown/MermaidRenderer.tsx`:

**1. No SVG sanitization (line 100):**
```tsx
<div
    style={webStyle}
    dangerouslySetInnerHTML={{ __html: svgContent }}
/>
```

The `svgContent` from `mermaid.render()` is injected directly into the DOM. SVG is a rich attack surface — elements like `<foreignObject>`, `<script>`, or event handler attributes (`onload`, `onerror`) could execute JavaScript.

**2. Missing `securityLevel` (lines 43-46):**
```tsx
mermaid.initialize({
    startOnLoad: false,
    theme: 'dark'
    // securityLevel is absent — defaults to 'loose'
});
```

Without `securityLevel: 'strict'`, mermaid allows click callbacks, `%%{init}%%` directives, and other features that can embed executable content in SVG output.

**3. Tauri desktop app has CSP disabled (`src-tauri/tauri.conf.json:22-24`):**
```json
"security": {
    "csp": null
}
```

Combined with the above, any XSS in the web rendering path has no CSP restriction in the Tauri desktop app, amplifying the impact.

## What PR #678 fixed vs. what remains

| Platform | Status | How |
|----------|--------|-----|
| iOS/Android WebView | Fixed (PR #678) | `JSON.stringify()` + `mermaid.render()` API |
| Web / Tauri desktop | **Unfixed** | Still uses `dangerouslySetInnerHTML` without sanitization |

## Suggested fix

```tsx
// 1. Add securityLevel to initialization
mermaid.initialize({
    startOnLoad: false,
    theme: 'dark',
    securityLevel: 'strict',
});

// 2. Sanitize SVG before DOM injection (defense in depth)
import DOMPurify from 'dompurify';
const sanitizedSvg = DOMPurify.sanitize(svgContent, {
    USE_PROFILES: { svg: true, svgFilters: true },
});
// Then: dangerouslySetInnerHTML={{ __html: sanitizedSvg }}

// 3. Enable CSP in tauri.conf.json
[...]

PR #1376, opened 2026-06-10
fix(app): render markdown SVG images on native via react-native-svg

chphch:
Markdown images whose source is an SVG — a `data:image/svg+xml` data URI or a `.svg` URL — render correctly on web but appear **blank on iOS/Android**, because `RenderImageBlock` (`MarkdownView.tsx`) renders every image through React Native's `<Image>`, which only decodes raster formats. This routes SVG sources through `react-native-svg` instead — `SvgXml` for inline markup decoded from a data URI, `SvgUri` for a remote `.svg` — while raster images keep using `<Image>` unchanged. A new `svgImageSource` helper detects SVG sources and decodes base64 / utf8 / percent-encoded data URIs, with unit tests.

`SvgXml` parses the SVG into `react-native-svg` components rather than injecting raw markup, so this adds no `dangerouslySetInnerHTML` sink (cf. #1086).

## Proof

Verified locally against a standalone `happy-server` (PGlite) + Expo web, authenticated via the `getDevWebQueryCredentials()` dev URL bypass and driven with Playwright. A markdown message carrying a base64 `data:image/svg+xml` image now renders the SVG (shapes + text); a PNG data URI still renders via `<Image>` (no raster regression). `tsc --noEmit` is clean and `vitest` passes (9 new `svgImageSource` tests + the markdown suite). `SvgXml` is already used on native elsewhere (`FileIcon.tsx`), so the web render exercises the same cross-platform component path that was previously failing on device. Screenshot in the comment below.

Before/after emulator screenshots are in the comments below.

### DIFF
--- a/packages/happy-app/sources/components/markdown/MarkdownView.tsx
+++ b/packages/happy-app/sources/components/markdown/MarkdownView.tsx
@@ -17,6 +17,8 @@ import { MermaidRenderer } from './MermaidRenderer';
 import { t } from '@/text';
 import { isHttpMarkdownLink } from './linkUtils';
 import { openExternalUrl } from '@/utils/openExternalUrl';
+import { SvgUri, SvgXml } from 'react-native-svg';
+import { parseSvgImageSource } from './svgImageSource';
 
 // Option type for callback
 export type Option = {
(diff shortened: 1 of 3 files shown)


chphch:
## Proof — markdown SVG images render in-app

Captured against a throwaway standalone `happy-server` (PGlite) + Expo web, authenticated via the `getDevWebQueryCredentials()` dev URL bypass and driven headlessly with Playwright. A markdown message carrying a base64 `data:image/svg+xml` image now renders the SVG (top); a PNG data URI still renders via `<Image>` (bottom — no raster regression). 0 console errors; `tsc --noEmit` clean; `vitest` green (9 new `svgImageSource` tests + the markdown suite).

markdown SVG renders in-app

Before this change the top image was blank on iOS/Android (RN `<Image>` can't decode SVG). It now goes through `SvgXml` — the same `react-native-svg` component already used on native by `FileIcon` — so the web render above exercises the exact cross-platform path that was failing on device.

chphch:
Native before/after, captured on an Android emulator (arm64 debug build; same APK for both frames — `react-native-svg` 15.12.1 is already an upstream dependency, so this diff is JS-only and only Metro's worktree differs). Seeded assistant message contains a markdown image whose source is an `.svg` URL.

**Before** (upstream/main): the heading and alt text render, but the image area is a blank grey box. Logcat shows Fresco fetching the bytes and rejecting them:

```
ProgressiveDecoder: unknown image format {uri: http://…/test.svg, firstEncodedBytes: 3C73766720…}
```

(`3C 73 76 67 20` = `<svg ` — fetched fine, undecodable by the raster pipeline.)

before

**After** (this branch): the same session and URL render through `react-native-svg`:

after

Issue #1270, opened 2026-05-14
Feature request File/images uploads on Mobile


ramvillamor:
Pleeeeease have this feature!

yiranxiaohui:
+1

Issue #919, opened 2026-03-26
Feature requests: image support, text copy on mobile, voice-to-text input mode


WhymustIhaveaname:
First of all, thank you so much for Happy Coder! It's currently the only working solution for controlling Claude Code from a phone, and it's been an absolute game-changer for me. I love using it.

Because I enjoy it so much, I'd love to see it get even better. Here are three feature suggestions:

1. **Image support** — It would be great to be able to send images from the phone to the Claude Code session.

2. **Text copy on mobile** — Currently, it seems impossible to copy text from the conversation on mobile. Being able to long-press and copy would be very helpful.

3. **Voice-to-text input mode** — In addition to the existing voice chat mode, it would be nice to have a "voice-to-text" mode where speech is transcribed into text in the input field before sending, so I can review and edit it first.

Thanks again for building this wonderful tool! 🙏

P.S. This issue was filed from my phone using Happy to control Claude Code! 📱

WhymustIhaveaname:
Update on point 2: I just figured out how to copy text on mobile — sorry about that! Please disregard that one. The other two suggestions (image support and voice-to-text input mode) still stand. Thanks!

WhymustIhaveaname:
Another finding: **App-side YOLO mode doesn't pass `--dangerously-skip-permissions` to the Claude process.**

I selected YOLO mode in the mobile app, but when I inspected the actual `claude` process on the server, it was started without the `--dangerously-skip-permissions` flag:

```
$ cat /proc/<claude_pid>/cmdline | od -c
# Only shows "claude" with no flags

$ cat /proc/<claude_pid>/environ | tr '\0' '\n' | grep -i "yolo\|dangerous\|permission\|skip"
# Nothing found
```

**Impact:** The main session works fine because Happy intercepts permission requests and auto-approves them at the Happy layer. But Claude Code **subagents** (background agents spawned via the `Agent` tool) don't go through Happy — they send permission requests directly, and since `--dangerously-skip-permissions` was never passed, all their tool calls get denied.

This means any background agent task that uses Bash, WebSearch, etc. will silently fail when running under Happy's YOLO mode.

**Expected behavior:** When YOLO mode is enabled in the app, Happy should start Claude with `--dangerously-skip-permissions` (or `happy --yolo` equivalent) so that subagents inherit the permission bypass.

**Workaround:** Start happy from the CLI with `happy --yolo` instead of relying on the app toggle.

bra1nDump:
[...]

Issue #70, opened 2025-09-02
Can I add pictures to a conversation?


wanwangiao:
Can I add pictures to a conversation?

GrocerPublishAgent:
**Update on image uploading - need your input on UX trade-off:**

The challenge with adding image support stems from our encryption architecture. Here's how it works and why we need your feedback:

**Current Architecture:**
- The mobile app and CLI don't need to be online simultaneously (this is a feature)
- Mobile app uploads encrypted instructions to our relay server (acts as a mailbox)
- CLI downloads and processes these encrypted instructions when it comes online


To let you upload an image, it would be encrypted on your phone, sent to the relay server, then downloaded by the CLI

**The Design Problem:**
When you upload an image and close the app, what happens when you come back to that conversation?

**Option 1: Store images on relay server**
- ✅ You can see uploaded images when revisiting old chats
- ❌ Significant disk usage on the free demo relay server operated by volunteers (costly)

**Option 2: Temporary local storage only** 
- ✅ Easy to implement - images saved to temp folder on your CLI computer
- ✅ Minimal server storage needed
- ❌ **You won't see previously uploaded images when revisiting old chats**
- ❌ Only see a message like "Image uploaded: photo.jpg"

**Question for you:** Would you be OK with Option 2? Essentially, images would work for active conversations but wouldn't persist in chat history when you return later. The CLI would still receive and save them locally, but the mobile app wouldn't re-display them in old conversations.

This would let us ship image support quickly if the UX trade-off is acceptable. Please let me know!

shekohex:
What about in Option 2: Temporary local storage only, when it shows `"Image uploaded: photo.jpg"` the user is able to click on it (similar to how we currently click on file edits, or command outputs) and when clicked the cli would send it back to the mobile app through the server, in the same fashion we currently handling the files and diffs? image locations can be customizable in the cli settings instead of just being in tmp. wdyt @GrocerPublishAgent ?

GrocerPublishAgent:
making the image destination configurable is good idea. 

I was only thinking of the case where images are being uploaded via the mobile app -> across the internet -> down into your computer. But you are right, images can be pasted in and @mentioned into the terminal. I'll need to quickly check if those pasted images get. saved somewhere.

 I can make the mobile app load the image from your computer if the image is still there.

zhigang1992:
Option 2 sounds totally fine to me.
[...]

PR #1387, opened 2026-06-12
feat(app,cli): camera + file/PDF attachments with HEIC/size normalization

### LINKED ISSUE #1319: App: add file/image uploads and Opus 4.7 xhigh effort option
## Feature requests

Thanks for building Happy — it is very useful for using Codex / Claude Code from the app.

I would like to request two small but important App updates:

### 1. Add image and file upload entry points in the App

Right now, uploading images or attaching files from the App is inconvenient. Common cases:

- send a screenshot to the model for troubleshooting / UI review
- upload a photo for visual analysis
- attach a PDF / document / log file / code snippet file to the current conversation

Suggested UX:

- add an attachment / plus button near the message input
- support image picker / camera / file picker
- show selected attachments before sending
- send the attachments into the active Codex / Claude Code session just like a normal message attachment

This would make the App much more practical for real mobile workflows. Related: #1270.

### 2. Expose Opus 4.7 `xhigh` thinking / effort level in the App

For Opus 4.7, the available thinking / effort levels are not only `high` and `max`; there is also a custom `xhigh` option.

At the moment, this option is not visible/selectable in the App UI I am using. Please add `xhigh` to the model effort / thinking-level picker for Opus 4.7 as well.

I noticed #1184 mentions that Opus 4.7 and `xhigh` were added on `main`, so this may be an App UI / release parity issue rather than a core implementation issue.

Thanks!

jlixfeld:
## Summary

Extends the existing `expImageUpload` attachment pipeline (gallery images only today) to the full scope of #1319:

### App (`packages/happy-app`)
- **Attachment source chooser** — the picker button now offers **Photo Library / Take Photo / Choose File** on native (web keeps direct picking; paste/drag already worked)
- **Paste from clipboard (iOS/Android)** — when the clipboard holds an image, the native chooser shows a **Paste from Clipboard** row, gated by `Clipboard.hasImageAsync` so it only appears when relevant (and stays silent until tapped — no premature iOS paste banner). Pasted bytes are staged to a cache file (the upload path needs a `file://` URI) and run through the same normalize/thumbhash/upload pipeline as picked images. Web already had its own paste listener.
- **Camera capture** via `expo-image-picker` `launchCameraAsync` + camera permission flow (`NSCameraUsageDescription` added)
- **File attachments** via `expo-document-picker` (already a dependency, previously unused) — flows through the existing encrypted-blob upload and `t:'file'` session-event path unchanged
- **Image normalization before upload** (`attachmentNormalize.ts`):
  - HEIC/unsupported formats → JPEG q0.9 — **fixes silent HEIC drop**: the CLI's magic-byte check (`detectClaudeImageMime`) skips HEIC with only a debug log today, so iPhone photos often never reach the model
  - Downscale only when long edge > 1568 px (the Claude vision API ceiling — larger is discarded server-side anyway, and >5 MB images are rejected outright)
  - PNGs stay PNG (lossless, keeps screenshot transparency); in-range supported images pass through untouched
- i18n for all new strings across all 10 locales; settings label updated to “Attachments”

### CLI (`packages/happy-cli`)
- Attachment→content-block conversion extracted to `attachmentContentBlocks.ts` and extended beyond images:
  - `%PDF-` magic → `document` content block
  - `text/*` mime / known text extension / clean UTF-8 → fenced text block with filename
  - anything else → **visible notice** in the message (`[Note: attachment(s) … were omitted]`) instead of today's silent drop
- Routing decisions are made on decrypted **bytes**, not the wire mimeType (iOS pickers lie)

## Tests
- `attachmentNormalize.spec.ts` — 7 cases for the normalization decision logic
- `attachmentContentBlocks.test.ts` — 7 cases (image/PDF/text/extension-fallback/unsupported-notice/empty/HEIC)
- `pnpm typecheck` clean in happy-app; happy-app suite 594 passed; pre-existing codex/integration failures in happy-cli are untouched by this PR (no `src/codex` files in the diff)

## Test plan
- [x] Unit tests above
- [ ] Manual on-device: gallery image, camera photo, PDF, text file, unsupported binary → notice
- [ ] Manual on-device: copy an image elsewhere → open chooser → **Paste from Clipboard** row appears → attaches and uploads

Closes #1319. Related: #1270, #919, #70.

### DIFF
--- a/docs/superpowers/plans/2026-06-12-attachments-testflight.md
+++ b/docs/superpowers/plans/2026-06-12-attachments-testflight.md
@@ -1,0 +1,886 @@
+# Attachments (Camera + Files) + Local TestFlight Build — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Extend Happy's existing image-attachment pipeline with camera capture, file/PDF attachments, and HEIC/size normalization; then ship a personal iOS TestFlight build that bundles PRs #1372, #1373, and this feature.
+
+**Architecture:** Upstream main already has the full image pipeline (app picker → encrypted server blob → `t:'file'` session event → CLI download/decrypt → SDK `image` content block), gated by the `expImageUpload` setting. We extend the edges only: app-side capture sources + normalization, CLI-side non-image content-block conversion. TestFlight ships from a local integration branch via headless xcodebuild + ASC API key.
+
+**Tech Stack:** Expo SDK 55 (expo-image-picker, expo-image-manipulator, expo-document-picker — all already in package.json), vitest, @anthropic-ai/sdk content blocks, xcodebuild + altool.
+
+**Conventions (from packages/happy-app/CLAUDE.md):** 4-space indent, `t(...)` for ALL user-visible strings added to ALL 10 translation files, `Modal` from `@/modal` (never RN Alert), `pnpm` only, run `pnpm typecheck` after app changes. Commit messages end with the Happy/Claude co-author block:
+
+```
+Generated with [Claude Code](https://claude.ai/code)
+via [Happy](https://happy.engineering)
+
+Co-Authored-By: Claude <noreply@anthropic.com>
+Co-Authored-By: Happy <yesreply@happy.engineering>
+```
+
+**Branch:** all Phase A tasks on `feat/attachments` (already exists, off upstream/main). Phase B on `local/testflight`.
+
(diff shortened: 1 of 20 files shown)


jlixfeld:
Heads-up for maintainers on the feature flag: the `expImageUpload` toggle in `settings/features.tsx` (around line 108 on main) is currently commented out with `{/* Image upload hidden — broken, shipping next release */}`.

This PR is aimed squarely at the "broken" part — it fixes the silent HEIC drop (the most common failure on iOS, since the gallery/camera hand back HEIC that the CLI's magic-byte check skips) and adds the camera + file/PDF sources. Once this lands, re-enabling that `<Item>` should be safe, so the feature becomes reachable without a code edit.

I deliberately did **not** un-comment it in this PR to keep the diff focused on functionality and let you decide the release timing. Happy to either flip it on here or leave it for a follow-up — your call.

PR #1403, opened 2026-06-15
fix(app): make buildSessionRowData's unread set required too

chphch:
## Re-scoped after #1567 landed

This PR originally fixed the whole unread-title flicker: it threaded `state.unreadSessionIds` through `updateSessionDraft` / `applyMachines` / `deleteMachine` and made `buildSessionListViewData`'s parameter required. **All of that landed independently in #1567**, so those hunks are gone from this branch. What is left is the one half #1567 did not take.

## What remains

#1567 made `buildSessionListViewData`'s `unreadSessionIds` required, and its own comment states the reasoning:

> Required on purpose: an omitted set silently rebuilds the list with `hasUnread=false` everywhere — exactly the bug this parameter caused twice.

But `buildSessionRowData` — the function that actually *reads* the set to compute `hasUnread` — is still optional on current main:

```ts
function buildSessionRowData(session: Session, unreadSessionIds?: Set<string>): SessionRowData {
    ...
    hasUnread: unreadSessionIds?.has(session.id) ?? false,
```

So the guard stops one level short of the place the silent wipe actually happens. This PR applies the same treatment there:

```ts
function buildSessionRowData(session: Session, unreadSessionIds: Set<string>): SessionRowData {
    ...
    hasUnread: unreadSessionIds.has(session.id),
```

## Risk

None at runtime. All three current call sites (`active-sessions`, the starred/date-grouped rows, and the fork-ordered rows) already pass the set, so this compiles as-is and behaves identically — `pnpm typecheck` is clean. Its only effect is that a future row-building path can't reintroduce the silent `hasUnread: false` wipe without a type error, which is the same protection #1567 added one level up.

## Relationship to #1580

Orthogonal, no merge-order dependency. #1580 changes *where the unread state lives* (a synced read position instead of a device-local set); this PR stops a rebuild from *dropping* it. If #1580 lands first this specific parameter goes away, but the equivalent guarantee carries over to its `currentViewingSessionId`, which that PR already declares as required — so the invariant survives either merge order.

### DIFF
--- a/packages/happy-app/sources/sync/storage.ts
+++ b/packages/happy-app/sources/sync/storage.ts
@@ -113,7 +113,7 @@ export interface SessionRowData {
     workspaceName: string | null;
 }
 
-function buildSessionRowData(session: Session, unreadSessionIds?: Set<string>): SessionRowData {
+function buildSessionRowData(session: Session, unreadSessionIds: Set<string>): SessionRowData {
     const isOnline = session.presence === "online";
     const hasPermissions = !!(session.agentState?.requests && Object.keys(session.agentState.requests).length > 0);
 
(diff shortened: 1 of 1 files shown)


chphch:
Closing to keep the review queue small — on a closer look this one changes nothing observable.

It tightens `unreadSessionIds` from optional to required so the `unreadSessionIds?.has(session.id) ?? false` fallback can't be hit. But that fallback is already unreachable on main: the parameter is a required positional at `storage.ts:273` and every call site passes a real `Set`. So this is a compile-time guard against a path that doesn't exist, not a fix. Happy to reopen if the signature ever loosens.

PR #1407, opened 2026-06-16
fix(cli): follow EnterWorktree cwd changes in session metadata

chphch:
When Claude switches working directory mid-session via the `EnterWorktree` tool, happy-cli never learns the cwd moved — `EnterWorktree` fires no hook (unlike a Bash `cd`, which fires `CwdChanged`), so the session's `metadata.path` stays pinned to the launch directory. The app keeps showing the original dir, the git-diff / file-browser views stay there, and `--resume` relaunches the session at the old dir, dropping the worktree entirely. This wires the session scanner — which already reads every transcript record — to follow the `cwd` field on those records and sync `metadata.path` when it changes, on both the local and remote (SDK) scanners.

The key detail: when `EnterWorktree` switches cwd, the transcript file stays in the launch project dir (only each record's `cwd` field flips), so the scanner keeps receiving records after the switch. That makes the record `cwd` the one observable signal for the no-hook `EnterWorktree` case — there's no `PreToolUse`/`CwdChanged` hook to catch it, and the file never moves to a new project dir for a watcher to follow.

A relocation gate keeps this from over-firing: only a worktree root or the launch directory counts as a relocation. Without it, a transient `cd subdir && cmd` would drag `metadata.path` around after the last build/typecheck subdirectory.

## Proof

Verified end-to-end on a live SDK/remote session (the path the mobile app drives): standalone `happy-server` + a paired `happy-cli` daemon built from this branch, a session spawned in a git repo, then Claude calls `EnterWorktree`. The session-info **Path** row updates from the repo root to the worktree, driven by the new `metadata.path` sync:

worktree metadata follows

- **Before:** `Path → …/cwd`
- **After `EnterWorktree`:** `Path → …/cwd/.claude/worktrees/gifprobe`

Daemon log confirms the scanner detects the switch from the transcript record's `cwd` field:

```
[SESSION_SCANNER] cwd changed via transcript record: …/.claude/worktrees/gifprobe
```

`metadata.path` is what the app's directory / git-diff / file-browser views and `--resume` all read, so each of them now tracks the worktree instead of staying pinned to the launch directory.

Reproduction:
1. Start a session in a git repo (mobile or `happy claude`).
2. Ask Claude to "work in a worktree" → it calls `EnterWorktree`, switching into `.claude/worktrees/<name>`.
3. **Before:** the app still shows the repo root as the session directory; after archive + `--resume` the session relaunches at the repo root, not the worktree.
4. **After this change:** `metadata.path` follows the worktree, so the dir / diff / file views and `--resume` all track it.

## Related

#1116 wires the `SessionStart`-hook cwd into the scanner's *watch path*. That hook doesn't fire for `EnterWorktree` (no hook at all) and targets the message-stream watch dir rather than `metadata.path`, so the two are complementary — this PR covers the `EnterWorktree` case and fixes the `--resume` / app-directory surface via `metadata.path`.

`pnpm typecheck` clean; CLI suite 782/782, including a regression test for the cwd-follow path.

### DIFF
--- a/packages/happy-cli/src/claude/runClaude.ts
+++ b/packages/happy-cli/src/claude/runClaude.ts
@@ -222,6 +222,9 @@ export async function runClaude(credentials: Credentials, options: StartOptions 
                         void session.sendClaudeSessionMessageFromLocalTranscript(msg);
                     },
                     onTranscriptEvent: updateClaudeGoalState,
+                    // Follow mid-session cwd changes (EnterWorktree fires no hook) so
+                    // the app's dir/diff/file views and --resume track the worktree.
+                    onCwdChange: (cwd) => session.updateMetadata((m) => ({ ...m, path: cwd })),
                 });
                 if (offlineSessionId) scanner.onNewSession(offlineSessionId);
                 return { session, scanner };
@@ -456,6 +459,9 @@ export async function runClaude(credentials: Credentials, options: StartOptions 
             session.sendClaudeSessionMessage(raw);
         },
         onTranscriptEvent: updateClaudeGoalState,
+        // Follow mid-session cwd changes (EnterWorktree fires no hook) so the
+        // app's dir/diff/file views and --resume track the worktree.
+        onCwdChange: (cwd) => session.updateMetadata((m) => ({ ...m, path: cwd })),
     });
 
     // Start Happy MCP server
--- a/packages/happy-cli/src/claude/utils/sessionScanner.test.ts
+++ b/packages/happy-cli/src/claude/utils/sessionScanner.test.ts
@@ -381,4 +381,40 @@ describe('sessionScanner', () => {
       apiErrorStatus: 429,
     })
   })
+
+  it('follows cwd only into/out of a worktree, not a transient subdir cd', async () => {
+    // Regression for a3f5bae1: onCwdChange fired on EVERY transcript-record cwd
+    // change, so a plain `cd packages/x && pnpm t` (whose cwd is stamped on the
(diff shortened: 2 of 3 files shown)


jlixfeld:
Independently hit this exact issue and ported your fix onto a downstream fork — can confirm the diagnosis and approach hold up. The `EnterWorktree`-fires-no-hook → record `cwd` as the only observable signal is correct, and the relocation gate (worktree-root / launch-dir only, ignoring transient `cd subdir && cmd`) is the key refinement — without it the dir/diff/`--resume` surface chases the last build/typecheck subdir around. Verified end-to-end: `metadata.path` follows into and back out of a worktree, transient subdir `cd`s ignored.

Pairs naturally with #1457 (a persistent worktree/branch status line) — that surfaces `metadata.path`, this keeps it accurate. 👍

chphch:
@bra1nDump — this one has had independent confirmation sitting on it since June, so I'd rather flag it than let it age further.

@jlixfeld hit the same issue independently, ported this patch onto a downstream fork, and verified it end-to-end: `metadata.path` follows into and back out of a worktree, and transient `cd subdir && cmd` steps are correctly ignored.

That second half is the part worth a reviewer's eye. `EnterWorktree` fires no hook, so the recorded `cwd` is the only observable signal — and without the relocation gate (accept worktree-root / launch-dir only), the dir/diff/`--resume` surface chases whatever build or typecheck subdirectory happened to run last.

Status: rebased onto current `main` today — mergeable, all checks passing.

@jlixfeld — apologies for the very late reply, and thank you for taking the time to verify it on your own fork. An independent port holding up is worth considerably more than anything I could say about my own patch.

Issue #568, opened 2026-02-10
Feature: Direct shell commands from mobile without burning tokens ($ prefix)


itspa1:
## Problem

Every message sent from the mobile/web app goes through the AI agent, even trivial shell commands like `git pull`, `ls`, or `pwd`. This burns API tokens for operations that don't need AI at all.

## Proposal

Intercept messages starting with `$` on the CLI side before they reach Claude. Execute them as shell commands directly and return the output through the existing encrypted session channel.

**From the phone:**
- `$git pull` → runs directly, returns output, zero tokens
- `$ls` → instant directory listing
- `$pwd` → shows working directory
- `$ docker ps` → works with or without space after `$`
- `where am i` → goes to Claude as normal

**False-positive prevention:** Messages like `$100`, `$5.99`, `${VAR}` are passed through to Claude untouched.

## Implementation

The change is entirely CLI-side — **no app changes needed**. It intercepts in the `onUserMessage` callback in `runClaude.ts`:

- Uses async `exec` (non-blocking, keeps WebSocket alive)
- Strips ANSI escape codes from output
- Truncates large output (50KB limit)
- Supports `cd` with persistent working directory across commands
- 120s timeout with explicit timeout message
- Sends output via `sendAgentMessage` for proper markdown rendering in the app

~50 lines of code total.

## Status

I have a working patch and I'm using it daily. Happy to open a PR if there's interest.

kosumic:
Actually, running commands itself, even without LLM, is very useful.
Persistent sessions and push notifications are by themselves helpful.

kosumic:
Codex use $ for skills and ! for shell commands

PR #1414, opened 2026-06-19
feat(agent-input): !-prefix bash command mode

chphch:
Adds a `!`-prefix bash command mode to the chat input, mirroring the Claude Code CLI: a message that starts with `!` runs as a one-off shell command in the session's working directory instead of being sent to the agent, and its output renders inline. It's per-message only — there's no persistent mode toggle. The command runs over the existing **session-scoped `bash` RPC** (the same handler that powers directory autocomplete and Windows `cmd` support), so it executes in and is sandboxed to the session cwd; the echoed command and its output are injected as **local-only messages** (no server round-trip, the agent is never involved), with output shown in a fenced code block.

## Proof

Captured locally via a standalone `happy-server` (PGlite) + Expo web + a paired `happy-cli` daemon, authenticated through the `getDevWebQueryCredentials()` dev URL bypass and driven headlessly with Playwright (live e2e — real daemon, real session). Typing `!pwd` and `!echo BASHMODE_OK_4242` runs each as a shell command and renders the output in a `sh` code block; the agent is never invoked.

!-bash mode running pwd and echo in a real session

Verification: `tsc --noEmit` clean, and the new strings are added to all 10 locale files. No new message kind and no reducer changes — the output reuses the existing agent-text markdown rendering.

Related: addresses #568 (run shell commands from mobile without burning tokens). #597 proposes the same capability with a `$` prefix; this uses `!` to mirror the Claude Code CLI's bash mode.

### DIFF
--- a/packages/happy-app/sources/-session/SessionView.tsx
+++ b/packages/happy-app/sources/-session/SessionView.tsx
@@ -829,7 +829,18 @@ export function SessionViewLoaded({
     // need to re-create on every keystroke.
     const handleSend = React.useCallback(() => {
         const liveMessage = composerHandleRef.current?.getMessage() ?? '';
-        if (liveMessage.trim() || selectedImages.length > 0) {
+        // !-prefix bash mode: a message starting with `!` runs as a one-off
+        // shell command in the session's cwd instead of being sent to the agent.
+        const trimmed = liveMessage.trim();
+        if (trimmed.startsWith('!')) {
+            const command = trimmed.slice(1).trim();
+            if (command) {
+                composerHandleRef.current?.clearMessage();
+                sync.runBashCommand(sessionId, command);
+                return;
+            }
+        }
+        if (trimmed || selectedImages.length > 0) {
             const attachments = selectedImages.length > 0 ? selectedImages : undefined;
             const communicationsToDismiss = [...pendingCommunications];
             composerHandleRef.current?.clearMessage();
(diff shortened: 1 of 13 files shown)


chphch:
Flagging an overlap so you don't have to work it out: #597 (@hztBUAA, opened 2026-02-13) answers the same request (#568) from the other side of the wire, and both PRs are sitting open. Here is the difference as I read it — I have no stake in which one lands.

| | #597 — CLI-side | #1414 — app-side (this one) |
|---|---|---|
| Where the command runs | new `execAsync` in `runClaude.ts`, intercepted in `specialCommands.ts` before Claude sees the message | the session `bash` RPC that already exists (`registerCommonHandlers.ts:176`) |
| Size | 2 files, +133/-4 | 5 source files (~100 lines) + the subtitle in 11 locales |
| Reaches | every client, since the interception is in the CLI | only where the app runs it |
| Ships via | a CLI release | an app build / OTA |
| Flag | always on | `expBashMode`, off by default |
| Prefixes | `$` and `!` | `$` and `!` (just matched yours in 6d90e3bf) |
| The output | goes through the session, so it persists and reaches your other devices | local-only messages — deliberately never sent, so it does **not** reach your other devices |

Two honest notes, one each way:

- **#597 covers more surfaces.** Intercepting in the CLI means it works from the web app, the phone, and a second machine alike. Mine only works where the app itself runs the command, which is a real limitation rather than a footnote.
- **Mine doesn't add a second place where we shell out.** The `bash` RPC is already scoped to the session cwd and already carries the Windows `windowsHide` handling; #597 introduces a separate `child_process.exec` path alongside it. That's a maintenance argument, not something a user would feel.

The sharpest user-visible difference is the last row of the table: my output never leaves the device, which is what keeps it cheap, but it also means the command you ran isn't there when you pick the session up somewhere else. #597 doesn't have that gap.

Say the word and I'll close this in favour of #597 — no explanation needed.

Issue #1308, opened 2026-05-19
Mobile app can miss session messages after backgrounding


pengjunfeng11:
## Summary

When the mobile app is sent to the background, locked, or the user switches to another app, important session updates can fail to reach the iPhone reliably.

There are two related problems:

1. Ordinary assistant messages are not pushed by design, so a suspended mobile websocket can miss realtime delivery until the app refreshes.
2. Session-level push notifications for `done`, `permission`, and `question` can be suppressed incorrectly because CLI agent sockets are counted as active non-machine clients.

## Current behavior

- The server intentionally does not send Expo push notifications for every `new-message` update.
- Push notifications are supposed to be sent for session-level events such as `done`, `permission`, and `question`.
- Backgrounded mobile clients depend on the websocket staying alive for ordinary message delivery, but iOS/Android can suspend it.
- On foreground resume, the app refreshes sessions and related state, but did not explicitly invalidate the currently visible session's message sync.
- The server presence check excludes `machine-scoped` sockets, but still treats `session-scoped` CLI coding-agent sockets as active user surfaces. Those sockets do not mean the user is looking at the app, so they can incorrectly suppress iPhone push notifications for task completion or permission requests.

## Expected behavior

- When an agent finishes a task or needs permission/clarification, the iPhone app should receive a push notification unless an actual user UI client is foregrounded.
- When the app returns to foreground, the currently visible session should fetch any messages that arrived while the mobile socket was suspended.

This does not reintroduce per-message push notifications. It keeps push notifications focused on user-attention events and makes foreground recovery reliable.

## Notes

Relevant code paths:

- `packages/happy-server/sources/app/push/pushDispatch.ts` documents that generic per-message pushes were removed.
- `packages/happy-server/sources/app/events/eventRouter.ts` checks whether a user is active before suppressing push.
- `packages/happy-app/sources/sync/sync.ts` handles `AppState` foreground/background transitions.
- `packages/happy-app/sources/-session/SessionView.tsx` calls `sync.onSessionVisible(sessionId)` for the active conversation.

Issue #1208, opened 2026-05-02
Android remote session sync issues: reconnect failure, permission timeout disconnect, long session stale


karle102017:
## Description

Three related issues with Android remote session sync (all on latest version):

### 1. Android app fails to reload session after reconnect
When exiting and re-entering the Happy Android app, the session cannot be refreshed/loaded. The conversation list won't show up until force-restart.

### 2. Permission confirmation timeout disconnects remote
Running `happy claude --continue` on PC, when a permission confirmation prompt appears and is not confirmed for a while, the remote connection drops. After that, the Android app can no longer send the confirmation back to PC.

### 3. Long-running sessions stop syncing to Android
After a session has been running for a long time on PC, updates stop appearing on the Android app entirely. No new messages are pushed to mobile.

## Environment
- Happy: latest version
- Platform: Android + PC
- Agent: Claude with `--continue`

## Expected Behavior
- Session should reliably reload on Android reconnect
- Permission prompts should remain accessible remotely regardless of idle time
- Long sessions should continue syncing to mobile without degradation

## Possibly Related
- #1170
- #1163
- #1148

bigtuff8:
Confirming item 1 reproduces on **iOS** too (the issue title says Android, but it's not Android-only).

- iPhone, latest App Store build, connecting to a self-hosted server over a private tailnet.
- After the app is backgrounded and returns to foreground, the conversation stops updating and the session won't reload. The **only** reliable recovery is force-closing (swipe up) and reopening, then rejoining the conversation.
- iOS suspends the app and appears to drop/half-close the WebSocket within seconds of backgrounding; on resume the socket looks alive but no messages flow, and the app doesn't force a clean re-sync.
- Pull-to-refresh inside the conversation occasionally forces a re-sync, but it's unreliable.

Would a **foreground re-sync over HTTPS** (pull any missed state on `didBecomeActive` rather than trusting the resumed WebSocket) be a viable fix path here? On iOS there's no battery/background exemption users can set, so a client-side resync-on-resume looks like the only robust option.

Issue #1056, opened 2026-04-13
Mobile session can stay stuck loading until user sends a message


dudu1111685:
## Summary
When a session appears in the Happy mobile app but stays stuck on loading while opening its history/conversation, sending **any message** into that session causes it to finish loading after a short wait.

This looks like a session/history hydration issue in the client/server flow, where the session is visible and reachable, but the initial load never completes until a new message triggers activity.

## Environment
- Happy `1.1.3`
- Linux server
- Mobile app connected successfully
- Reproduced with multiple sessions

## Symptoms
- Happy mobile app shows the machine as connected
- Sessions appear online/available
- Opening a session gets stuck on a loading spinner indefinitely
- The session is not actually dead: after sending any message to that session and waiting, it starts working / loads successfully

## Workaround
1. Open the session in the mobile app
2. If it keeps loading, send any message (even a trivial one like `hi`)
3. Wait a moment
4. The session history/session view finishes loading and becomes usable

## Why this seems like a bug
The session is already reachable enough to accept a message, but the initial history/session load does not resolve on its own. That suggests some missing trigger, stalled hydration, or incomplete initialization path for existing sessions.

## Reproduction idea
This was observed after recovering/restarting session state on the server, but the key user-visible bug is simpler:

1. Open Happy mobile
2. Select an existing session
3. Session gets stuck loading
4. Send any message to that session
5. Wait
6. Session suddenly becomes usable

## Expected behavior
Opening the session should load the history/view without requiring the user to send a dummy message first.

## Actual behavior
The session can remain stuck loading forever until a message is sent manually.

## Notes
In our case, once we discovered the workaround, the issue became recoverable immediately:
- no need to reinstall the app
- no need to relogin
- no need to recreate the session
- just send a message and wait

Posting this because it may help other users who think their sessions are dead when they are actually just stuck in a loading state.

PR #1418, opened 2026-06-21
fix(app): refresh open session messages on resume/reconnect

chphch:
**Reported by 5 people** — issues #1208, #1308, #1432 (and #1056, adjacent). One user built his own client over it:

> "I'm getting notifications that Claude has responded but it's not updating any of the text" … "I started to work on my own internal version just because this has been a massive issue"
> — calvin00001, Discord, 2026-04-05

He also found the workaround that pins the diagnosis exactly — leaving the session and coming back is what forces the refetch:

> "Oh found a trick. I just send a keystroke, back out to my session list, then go back in. This is on mobile btw."
> — calvin00001, Discord, 2026-04-04

<details>
<summary>All reports</summary>

|---|---|---|---|
| calvin00001 | 2026-03-30 | Discord | "my phone doesn't seem to be displaying output of a session that my tablet is" |
| dudu1111685 | 2026-04-13 | #1056 | "Opening a session gets stuck on a loading spinner indefinitely … after sending any message to that session and waiting, it starts working" |
| karle102017 | 2026-05-02 | #1208 | "Long-running sessions stop syncing to Android" |
| pengjunfeng11 | 2026-05-19 | #1308 | "On foreground resume, the app refreshes sessions and related state, but did not explicitly invalidate the currently visible session's message sync" |
| DearMrJ | 2026-06-23 | #1432 | "mobile doesn't receive real-time updates until it sends a message first" |

</details>

---

Messages an agent produces while the Happy app is backgrounded don't appear when you return to the app — the open chat stays stale until you leave and re-enter the session. The cause is a misdirected refresh trigger: `SessionView`'s effect that re-fetches an open session's messages depends on `realtimeStatus`, but `realtimeStatus` tracks the **LiveKit voice session** (`storage.ts` `useRealtimeStatus`, written only by `RealtimeVoiceSession*`/`RealtimeSession`) — not the data socket, whose status lives in the separate `socketStatus`/`useSocketStatus`. So on app resume (or any socket reconnect), that effect never re-runs, and `sync.ts`'s `onReconnected` handler deliberately skipped refreshing the open session's messages, with a comment that relied on exactly this dead trigger (`onSessionVisible` "called by SessionView when realtimeStatus changes"). Re-entering the session works only because a fresh mount re-fetches. This fix refreshes the currently-viewing session's messages on both `AppState → 'active'` and the socket `onReconnected` path via `onSessionVisible(currentViewingSessionId)` (a bounded forward sync). It also covers a foreground network blip that reconnects with `recovered=false` (server connection-state recovery off), which had the same gap.

## Proof

Verified end-to-end on a throwaway standalone `happy-server` (PGlite) + Expo web + a paired `happy-cli` daemon, driven with Playwright. Two browser contexts on one account: context **A** opens a session then goes offline (simulating the app being backgrounded); context **B** (online) sends a marked message to that session; context **A** comes back online. The missed message now renders on A on reconnect **without a remount** (`onSessionVisible` → `fetchForwardSince`). Screenshots/GIF in the PR comment below — A shows "No messages yet" while offline, then the full conversation (including the message it missed) after reconnect. `pnpm typecheck` clean.

### DIFF
--- a/packages/happy-app/sources/sync/sync.ts
+++ b/packages/happy-app/sources/sync/sync.ts
@@ -188,6 +188,20 @@ class Sync {
                 this.friendsSync.invalidate();
                 this.friendRequestsSync.invalidate();
                 this.feedSync.invalidate();
+
+                // Refresh the open chat's message log on resume. While the app is
+                // backgrounded the data socket is suspended/dropped, so any messages
+                // the agent produced while away arrive with no live `update` to apply
+                // them. The invalidations above only refresh the session LIST, not the
+                // viewing session's messages — without this the visible chat stays
+                // stale until the user leaves and re-enters it (it only re-fetches on a
+                // fresh SessionView mount). getMessagesSync does a bounded forward sync;
+                // if the socket hasn't reconnected yet, InvalidateSync's backoff retries
+                // until it has.
+                const resumeViewingSessionId = storage.getState().currentViewingSessionId;
+                if (resumeViewingSessionId) {
+                    this.onSessionVisible(resumeViewingSessionId);
+                }
             } else {
                 log.log(`📱 App state changed to: ${nextAppState}`);
                 this.maybeStartBackgroundSendWatchdog();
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — before / after

Captured on a standalone `happy-server` (PGlite) + Expo web + a paired `happy-cli` daemon, driven with Playwright. Methodology that isolates this fix specifically:

1. Context **A** opens a session and **primes it with a message** so it is *already-loaded* (`sessionLastSeq` set) — the user's real condition. (The custom session-prefetch only fetches *never-loaded* sessions, so an empty session would be refreshed by prefetch and mask the fix.)
2. The agent is then stopped, so nothing answers later messages (no post-reconnect `update` to trigger `handleUpdate`'s gap-refresh).
3. A goes **offline for >45s** (past the server `pingTimeout`) so the socket is *genuinely disconnected* — a shorter pause keeps the socket alive and the broadcast is merely buffered, which masks the bug.
4. Context **B** sends a marked message to the same session **while A is disconnected**, so A truly misses it.
5. A comes back **online** (resume).

before/after

**Without the fix** (top): after reconnect A is stale — the message that arrived while A was away is missing; it only appears after leaving and re-entering the session. **With the fix** (bottom): the missed message renders automatically on reconnect, no remount. The only change between the two runs is `sync.ts` (verified: the served bundle contained the fix in the "after" run and not in the "before" run). `pnpm typecheck` clean.

Issue #685, opened 2026-02-21
feat: Support resuming Claude Code sessions when spawning from mobile (remote mode)


RooikeCAO:
**Is your feature request related to a problem?**

When I start a Claude Code session from the terminal (`claude` CLI) and later want to resume it from my phone using the Happy mobile app, the session always starts fresh. The `--resume` flag works perfectly in local mode, but when the daemon spawns a new process for remote mode, it doesn't pass any session resume parameters.

**Current behavior:**
1. Work on a Claude Code session locally (session ID: `abc-123`)
2. Close the session
3. Open Happy mobile app → select the same project directory → start session
4. A brand new session is created, losing all previous context

**Expected behavior:**
The mobile app should offer an option to resume the most recent session in that directory, or allow specifying a session ID to resume.

**Root cause (from source code analysis):**

In `spawnSession()` (around line 4543 of `index-B3gQr6vs.mjs`), the spawned args only include:
```javascript
const args = [
  agentCommand,
  "--happy-starting-mode", "remote",
  "--started-by", "daemon"
];
```

No `--resume` or `--continue` flag is ever passed. Additionally, `runClaude()` generates a new `randomUUID()` as `sessionTag` every time, making it impossible to reconnect to an existing Claude Code session.

Meanwhile, `claudeRemote()` already has code to handle `--resume <session-id>` (lines 1667-1685) — it just never receives one from the daemon spawn path.

**Proposed solution:**

1. Add a `claudeSessionId` field to the `spawn-happy-session` RPC interface
2. In the mobile app's "New Session" screen, add a "Resume last session" toggle or a session picker showing recent sessions in the selected directory
3. When `claudeSessionId` is provided, pass `--resume <claudeSessionId>` to the spawned Happy CLI process

**Workaround:**
Currently, users must first run `happy --resume <session-id>` in the terminal, then switch to phone control. The session persists within the same process (local ↔ remote switching works). But this defeats the purpose of remote-first control.

**Environment:**
- Happy CLI: v0.13.0
- Claude Code: v2.1.50
- Platform: macOS (Apple Silicon)

suneox:
It appears that this issue is quite significant, so anyone who owns or maintains it should take a look at it. The PR team can fix this issue because, currently, we only display resume sessions but no action is taken on this matter.

Issue #621, opened 2026-02-15
Feature Request: Session picker for /resume in remote mode


mkmkkkkk:
## Problem

`/resume` (or `--resume` without session ID) does not work in remote mode. The code explicitly logs:
```
"Found --resume without session ID - not supported in remote mode"
```

This means mobile users cannot browse and resume previous sessions from the app — they must either know the exact session ID or switch to local mode first.

## Expected Behavior

When using `/resume` from the mobile app, display an interactive session picker (similar to the terminal experience) showing recent sessions with their summaries, so users can select one to resume.

## Current Workaround

- Use `/resume` on desktop terminal first, then let mobile take over
- Pass explicit session ID: `--resume <session-id>`

## Context

- `claudeRemote()` in `dist/index-B3gQr6vs.mjs` line ~1677 explicitly skips resume without session ID
- Session files exist at `~/.claude/projects/<project-id>/*.jsonl` and can be scanned
- The session scanner (`SESSION_SCANNER`) already exists in the codebase for other purposes

Issue #478, opened 2026-01-26
Resume existing claude code initiated sessions into Happy (not resuming Happy initiated)


digijoebz:
goal: I have an existing claude code session with a lot of context that I want to exit and then reopen with Happy. Is it possible to do this? Resuming only gets me Happy initiated session list. I was trying claude-args but didn't work. thanks

tomasvts:
So you have to options:
```
  -r, --resume [value]                              Resume a conversation by session ID, or open interactive picker with optional search term
  --session-id <uuid>                               Use a specific session ID for the conversation (must be a valid UUID)
```


I tried with session-id
```
happy --session-id=e24cdc45-28d0-4864-a8bc-acc3bcc1196a
```
And I got an error that the session was already in use. 

But with resume it worked
```
happy --resume e24cdc45-28d0-4864-a8bc-acc3bcc1196a 
```

Issue #147, opened 2025-09-29
Feature Request: Resume claude code/codex historical sessions from app


JasprW:
can happy daemon fetching historical sessions of cc/codex and allow start happy session with historical session on app?

wenerme:
also, `happy claude -c` will work on server side, the app dont get the history, but have title

chansearrington:
This drives me nuts there's no way to hand off from server to mobile. But you can go the other way.

dannyshisler:
I've just tried happy for the first time, and resuming the session is the first thing I've tried. I don't think it's usable for me without this - I need a tool that will allow me to continue sessions when I step away from the computer

KaranocaVe:
Actually, it just needs to make happy forward the command-line arguments of codex, like Claude does, but happy hasn't implemented this

payprays:
I switch to [cloudcodeui](https://github.com/siteboon/claudecodeui) to solve the problem.

PR #1425, opened 2026-06-21
feat(session-list): copy Happy + original session IDs from the long-press menu

chphch:
Adds **two related copy actions** to the session quick-actions menu (long-press a session row, or the web actions popover). They are small, self-contained, and share the same wiring, so they're grouped here.

Both are gated behind an experimental **`expCopySessionId`** setting, **off by default**. With it off the long-press quick-actions menu is exactly upstream — neither entry appears. Users opt in from Settings → Features.

### 1. Copy Happy session ID

Copies the Happy session ID to the clipboard and shows the existing confirmation toast. Today the only way to grab a session's ID is to open **Session Info** and tap the "Happy Session ID" row — this surfaces the same copy one long-press away, alongside the existing (dev-only) "Copy session metadata" action.

Implemented as a `copySessionIdToClipboard` helper mirroring the existing `copySessionMetadataToClipboard`. Reuses the `sessionInfo.happySessionIdCopied` / `sessionInfo.failedToCopySessionId` toasts and adds one new label key, `sessionInfo.copySessionId`, across all locales.

### 2. Copy original session ID (the agent's own ID)

Sits right below it and copies the **underlying agent's own session id** — the Claude Code session id (`metadata.claudeSessionId`), or the Codex thread id (`metadata.codexThreadId`) for Codex sessions — as opposed to Happy's own id. These ids already have dedicated copy rows on the **Session Info** screen; surfacing the most-used one a long-press away is handy for `claude --resume <id>` and for cross-referencing agent logs.

Implemented as a `copyOriginalSessionIdToClipboard` helper next to the one above. It resolves `claudeSessionId ?? codexThreadId`, **reuses the existing** `sessionInfo.claudeCodeSessionIdCopied` / `codexThreadIdCopied` (and `failedToCopy…`) toasts — so no new toast strings — and adds one new label key, `sessionInfo.copyOriginalSessionId`, across all locales. The row is **hidden when the session has no backend id yet**.

### Why "Copy **Happy** session ID"

With two "copy … ID" actions side by side, a bare "Copy session ID" is ambiguous. Action 1's label names Happy explicitly, matching the "Happy Session ID" row on the Session Info screen.

Both actions flow through the same `useSessionQuickActions` `actionItems` path, so every consumer picks them up with no per-surface wiring — the long-press menu in `SessionsList` / `ActiveSessionsGroupCompact`, and `SessionActionsPopover`.

## Proof

Captured via **live-e2e**: a real `happy-cli` daemon paired to a standalone `happy-server` (PGlite) + Expo web, driven headlessly with Playwright. A real Claude session was spawned so `metadata.claudeSessionId` is set, then the session-list context menu was opened and the new action tapped.

copy Happy + original session id

- The context menu shows **Copy Happy session ID** and, right below it, **Copy original session ID** (with the `</>` icon).
- Tapping the latter copied the Claude session UUID `fcf5ed12-…-e525330e32a6` to the clipboard — asserted `navigator.clipboard.readText()` is a UUID, i.e. the agent's own id and not the Happy session id — and showed the "Claude Code Session ID copied to clipboard" toast.
- Action 1 is verified the same way on web: tapping it writes `session.id` (asserted `navigator.clipboard.readText() === session.id`) and shows the "Happy Session ID copied to clipboard" toast.
- The second row is hidden for sessions with no agent id yet; for Codex sessions it copies `metadata.codexThreadId` instead.

`pnpm typecheck` is clean; CI runs the authoritative check on this PR.


chphch:
Closing — this one is for my own workflow rather than the project's.

Both IDs already have copy rows on the Session Info screen; this only surfaces them one long-press earlier. And the reason I wanted them handy — `claude --resume <id>` and cross-referencing agent logs — is the opposite of what people actually ask for in #478 / #685 / #621 / #147, which is for Happy to resume sessions *without* the user ever handling an id.

Keeping it on my fork. Happy to reopen if a real need for it turns up.

PR #1426, opened 2026-06-21
fix(app): call SessionsList hooks before the early return (Rules of Hooks)

chphch:
`SessionsList` defines three `useCallback` hooks (`keyExtractor`, `renderItem`, `HeaderComponent`) **after** its early `if (!data) return`, which violates the Rules of Hooks: when a mounted instance's `data` (the memoized `sessionListViewData`) flips from `null` to a populated array, the render goes from fewer hooks to more and React throws `Rendered more hooks than during the previous render`, dropping the whole screen to the error boundary. In normal use the store has already settled `sessionListViewData` to `[]` (non-null) before the component mounts, so the early-return path is effectively dead and the bug stays latent — but any code path that mounts the list while the store is still empty and then populates it hits the crash. This moves the early return below all hooks so the hook order is identical on every render, and null-guards `renderItem`'s `data` lookups (it's now typed nullable above the narrowing, though FlatList only invokes it once `data` is non-null). No behavior change otherwise.

## Proof

Reproduced on Expo web by mounting `<SessionsList/>` while `sessionListViewData === null`, then injecting sessions on the next tick (the null → array transition). Left = current `main` (error boundary: "Rendered more hooks than during the previous render"); right = with this fix (the list renders normally).

SessionsList hook-order before/after

`pnpm typecheck` is clean.

### DIFF
--- a/packages/happy-app/sources/components/SessionsList.tsx
+++ b/packages/happy-app/sources/components/SessionsList.tsx
@@ -420,13 +420,6 @@ export function SessionsList({
         return [...hierarchy, ...legacyItems, ...archiveToggle, ...archivedRows];
     }, [flatSessionList, hasArchivedSessions, hideArchivedSessions, machines, sourceData]);
 
-    // Early return if no data yet
-    if (!data) {
-        return (
-            <View style={[styles.container, flatSessionList && styles.containerFlat]} />
-        );
-    }
-
     const keyExtractor = React.useCallback((item: SessionListDisplayItem, index: number) => {
         switch (item.type) {
             case 'machine-header': return `machine-header-${JSON.stringify(item.machineId)}`;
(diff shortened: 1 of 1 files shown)


chphch:
Rebased onto current `main` (`3cf9ce70f`).

The three edits are unchanged in intent, but every hunk's context had moved: `c63c80ff3` rebuilt this component, so the signature now takes `topContentInset` / `bottomContentInset` / `onScroll`, `keyExtractor` and `renderItem` are typed `SessionListDisplayItem` rather than `SessionListViewItem`, and `renderItem` gained `machine-header` / `archive-toggle` / `project-group` cases. Reapplied against that shape.

The Rules-of-Hooks violation itself is untouched by the rework: the `if (!data)` guard still sits above three `useCallback`s, so a mounted instance still changes its hook count when `data` flips from null to an array. `pnpm typecheck` clean.

PR #1427, opened 2026-06-21
feat(app): fullscreen pinch-to-zoom viewer for Mermaid diagrams

chphch:
Inline Mermaid diagrams render small, and a tall diagram gets clipped, with no way to zoom in. An expand button on each diagram now opens a fullscreen viewer.

- **Native (iOS/Android):** the diagram re-renders in a fullscreen WebView and zoom is applied inside that WebView. Pinch / pan / double-tap are handled by injected JS.
- **Web (`MermaidViewer.web.tsx`):** wheel-zoom toward the cursor, drag-to-pan when zoomed, double-click to toggle, Esc / × to close.

No new dependencies (no `svg-pan-zoom` — the pan/zoom is a few lines of transform math).

## Why the native zoom is two-phase

A CSS `transform: scale()` does not re-render an SVG. Chromium rasterizes the layer once and the compositor stretches that bitmap, and `will-change: transform` makes it permanent: `MinimumRasterContentsScaleForWillChangeTransform` pins the raster scale to devicePixelRatio and `ShouldAdjustRasterScale` then declines every later increase ([`cc/layers/picture_layer_impl.cc`](https://chromium.googlesource.com/chromium/src/+/main/cc/layers/picture_layer_impl.cc)). The hint propagates to descendants (`node_or_ancestors_will_change_transform`), so the child SVG cannot escape it. Chrome 53+ would otherwise [re-raster a script-driven scale change](https://developer.chrome.com/blog/re-rastering-composite) crisply — the hint is exactly the opt-out.

So while fingers are down the composited transform carries the scale and the gesture stays cheap; on lift the zoom is **committed** into the SVG's own `width`/`height`, the transform scale returns to 1, and the hint is released. At rest the diagram is laid out at its true size and painted at full device density at any zoom level, rather than depending on the compositor's raster policy.

## Proof

**Feature end-to-end**, captured on the web build (standalone server + Expo web, driven with Playwright). Sequence: inline diagram with its expand button → tap → fullscreen viewer → wheel-zoom in → drag-pan → × closes back to the chat.

mermaid fullscreen zoom

**Zoom fidelity**, measured on the injected HTML this file ships. A script extracts the template literal from `MermaidViewer.tsx` verbatim, loads it in Chrome at `deviceScaleFactor: 3`, and drives a real two-finger pinch via CDP `Input.dispatchTouchEvent`. The frames below come from `Page.startScreencast`, i.e. the composited surface — an ordinary screenshot re-rasterizes on capture and hides the effect entirely, which is why the first attempt at this measurement showed no difference at all.

zoom crispness before and after

| | glyph-edge blur, 3x | glyph-edge blur, 8x |
|---|---|---|
| composited `transform: scale()` with `will-change` | 7.0 px | 12.1 px |
| committed into the SVG | 2.0 px | 2.0 px |

The blur growing in proportion to the zoom factor is the signature of a 1x bitmap being upscaled.

Also checked on the same rig: re-pinching from an already-committed 8x keeps identical sharpness (no re-promotion flicker), a gesture ended by `touchcancel` still commits, a 35-node diagram baked to 8x (6751 × 419 CSS px) still paints rather than dropping tiles, and the geometry during the gesture and after the commit is pixel-identical — `display: block` on the SVG is load-bearing for that last one, because the inline baseline strut otherwise belongs to the stage box, scales with the gesture, and makes the diagram jump 16.3 px the instant the zoom is committed.

## Not exercised

- **A real iOS/Android device.** The measurement above runs the shipped injected HTML in desktop Chrome, which is the same Blink/cc compositor as the Android System WebView but not the same build; the mechanism is cited from Chromium source rather than assumed.
- **iOS specifically.** The commit path is plain layout, so it does not depend on a raster policy, but it has not been run on WKWebView.

`pnpm typecheck` clean; `vitest run` 1114 passed.

### DIFF
--- a/packages/happy-app/sources/components/markdown/MermaidRenderer.tsx
+++ b/packages/happy-app/sources/components/markdown/MermaidRenderer.tsx
@@ -1,9 +1,12 @@
 import * as React from 'react';
-import { View, Platform, Text } from 'react-native';
+import { View, Platform, Text, Pressable } from 'react-native';
 import { WebView } from 'react-native-webview';
+import { Ionicons } from '@expo/vector-icons';
 import { StyleSheet, useUnistyles } from 'react-native-unistyles';
 import { Typography } from '@/constants/Typography';
 import { t } from '@/text';
+import { Modal } from '@/modal';
+import { MermaidViewer } from './MermaidViewer';
 
 // Tall diagrams scroll inside a capped container instead of taking over the chat
 const MAX_DIAGRAM_HEIGHT = 600;
@@ -17,6 +20,21 @@ const webStyle = (backgroundColor: string): any => ({
     overflow: 'auto',
 });
 
+// Tap target to open the diagram in the fullscreen zoomable viewer.
+function ExpandButton({ onPress }: { onPress: () => void }) {
+    return (
+        <Pressable
+            onPress={onPress}
+            hitSlop={8}
+            style={style.expandBtn}
+            accessibilityRole="button"
+            accessibilityLabel="Expand diagram"
+        >
+            <Ionicons name="expand" size={16} color="#fff" />
+        </Pressable>
+    );
+}
+
 // Mermaid render component that works on all platforms
 export const MermaidRenderer = React.memo((props: {
     content: string;
@@ -29,6 +47,10 @@ export const MermaidRenderer = React.memo((props: {
         const { width } = event.nativeEvent.layout;
         setDimensions(prev => ({ ...prev, width }));
     }, []);
+
+    const openViewer = React.useCallback(() => {
+        Modal.show({ component: MermaidViewer, props: { content: props.content } } as any);
+    }, [props.content]);
 
     // Web platform uses direct SVG rendering for better performance and native DOM integration
     if (Platform.OS === 'web') {
@@ -103,6 +125,7 @@ export const MermaidRenderer = React.memo((props: {
                     style={webStyle(theme.colors.surfaceHighest)}
                     dangerouslySetInnerHTML={{ __html: svgContent }}
                 />
+                <ExpandButton onPress={openViewer} />
             </View>
         );
     }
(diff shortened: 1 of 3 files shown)


chphch:
Force-pushed (`919f658f` → `2733fb76`, still one commit) to fix a real defect in this PR, and to correct a claim the body was making.

**The body said the SVG "stays crisp" under the CSS transform. It did not.** `#stage` carried a resting `will-change: transform`, which is exactly the hint that tells Chromium to rasterize the layer once and never again — so zooming stretched a 1x bitmap. Measured on the composited surface at `deviceScaleFactor: 3`, a glyph edge spanned **12.1 px at 8x zoom** where a genuine re-render spans 2.0 px.

zoom crispness before and after

The zoom is now two-phase: composited transform while fingers are down, then committed into the SVG's own `width`/`height` on lift, with the hint released. Full mechanism, the Chromium source citation, and the rest of the numbers are in the updated description.

Four smaller things came out of the same pass and are in the same commit:

- `#stage svg` is `display: block` — the inline baseline strut belonged to the stage box and scaled with the gesture, so the diagram jumped 16.3 px vertically the moment the zoom was committed.
- The gesture settles on `touchcancel` as well as `touchend`. An interrupted gesture (notification shade, incoming call) never reports its remaining touches, and was the one path that skipped the commit and left the hint pinned.
- The reference size is retaken at gesture start while the transform is still identity, closing the window where a pinch in the first 200 ms baked a multiple of a pre-font-settle size.
- The viewer pins mermaid to `11.12.2` — the version `MermaidRenderer` already pins — so the inline diagram and its fullscreen copy cannot render through different builds.

Worth saying how this got missed the first time, since it bears on the "not exercised" note: the original verification for the native path was code review of the injected touch handlers, and the first attempt at measuring it used ordinary screenshots — which re-rasterize on capture and showed the before and after as identical. The numbers above come from `Page.startScreencast` frames driven by synthetic touch events over the extracted template literal, so the native gesture path is now exercised rather than reasoned about.

`pnpm typecheck` clean, `vitest run` 1114 passed, no conflicts with `main`.

Issue #1313, opened 2026-05-20
Feature Request: Support Google Antigravity (agy) CLI


littoy:
Hi, first of all thanks for building and maintaining happy ❤️

Google announced that Google One Gemini CLI Pro subscription support will end around June 18, and the ecosystem is gradually migrating toward the new Google Antigravity / agy CLI workflow.

It would be great if happy could add support for the agy CLI as an alternative backend/provider.

Repository:
https://github.com/google/antigravity

Possible support ideas:

Detect agy command automatically if installed
Add configurable provider selection (gemini / agy)
Support passing prompts through agy chat or equivalent command
Reuse existing Gemini integration logic where possible

Why this matters:

Existing Gemini CLI users are being forced to migrate
Antigravity seems likely to become Google's primary AI CLI workflow
Supporting agy would help keep happy usable for current Gemini users

Eniggman:
When will you add new Gemini models?

PR #1430, opened 2026-06-23
feat(cli): add agy (Antigravity CLI) as an agent backend

### LINKED ISSUE #1429: feat(provider): add Antigravity (agy) CLI as a first-class agent provider
## Context

Google has positioned the Antigravity CLI (`agy`) as the successor to the now-deprecated `gemini` CLI. Happy currently ships a `gemini` provider (`packages/happy-cli/src/agent/factories/gemini.ts` + `packages/happy-cli/src/gemini/`) but no equivalent for `agy`.

There is an existing issue #1313 titled "Support Google Antigravity (agy) CLI", but its body asks a different question ("When will you add new Gemini models?") and has had no follow-up since 2026-05-20. Opening this one to scope the actual provider-support work, since the existing thread doesn't capture it.

## Why this isn't a simple rename

`agy` and the old `gemini` CLI are distinct products — different surface area, likely different streaming/event protocol, different config. Treating this as a global `gemini` → `agy` find-replace would silently break the existing `gemini` provider for users who still depend on it.

The cleaner shape is to add `agy` as a **new provider alongside** `gemini`, parallel to how Claude and Codex coexist today.

## Suggested scope

- `packages/happy-cli/src/agent/factories/agy.ts` — new factory, registered in `factories/index.ts`
- `packages/happy-cli/src/agy/` — `runAgy.ts`, `types.ts`, `constants.ts`, `utils/config.ts` mirroring the `gemini/` layout
- `packages/happy-app/sources/.../AgyDisplay.tsx` — app-side renderer analog of `GeminiDisplay.tsx`
- Slash-suggestion / model-picker entries
- Keep the existing `gemini` provider intact for back-compat; deprecate later in a separate PR once `agy` is stable

## Open questions

1. Does `agy` expose a JSON-line streaming protocol comparable to `gemini` CLI's, or does the integration need a different transport?
2. Auth: does `agy` reuse the Google account flow, or is there a new login path Happy needs to surface?
3. Is there interest from maintainers in this direction at all? Happy to draft a PR if so — would rather confirm before sinking the work.

(Mentioning #1313 for context, as it covers the same area.)

charliezong18:
## What & why

Adds **agy** — Google's [Antigravity CLI](https://antigravity.google) — as a first-class agent backend, so it can be driven from Happy (terminal + phone/web app) as a fallback when Claude Code hits its usage limit.

**Additive:** the existing Gemini integration is untouched.

## Design

agy has **no ACP / JSON streaming mode** (unlike Gemini = ACP, Codex = app-server JSON-RPC). Its only stable non-interactive surface is `agy --print "<prompt>"`, which streams plain text and persists a per-cwd conversation for resume. So it can't reuse `AcpBackend` — it follows the bespoke (non-ACP) **`openclaw`** pattern instead: a backend that translates its own output into Happy's `AgentMessage` stream and rides the same session pipeline, app-selectable without the ACP / `MessageAdapter` / `agentRegistry` machinery.

New `packages/happy-cli/src/agy/`:
- `constants`, `types`, `cliArgs` (`buildAgyArgs`), `conversationStore`
- `AgyBackend` — spawns `agy --print` per turn, streams stdout → `AgentMessage`
- `runAgy` — orchestrator mirroring `runOpenClaw`
- `ui/ink/AgyDisplay.tsx` — terminal status view

Key details:
- **stdin EOF:** `agy --print` blocks until stdin reaches EOF, so `AgyBackend` spawns with `stdio: ['ignore', 'pipe', 'pipe']`. An open stdin pipe hangs the turn forever (caught during testing).
- **Permissions:** Happy `yolo` / `safe-yolo` / `bypassPermissions` / `acceptEdits` → `--dangerously-skip-permissions`; otherwise `--sandbox`. (Print mode is one-shot and can't surface an interactive approval prompt, so non-skip modes defer to agy's own `settings.json`.)
- **Resume:** agy doesn't print the conversation id; it's read back from agy's per-cwd cache and passed via `--conversation`.
- **Model:** Happy's model string → `--model "<display name>"`; defaults to a Gemini model (the fallback exists *because* Claude is rate-limited).

## Registration

`'agy'` added everywhere `'openclaw'` already appears:
- **happy-cli:** `detectCLI`, `index.ts` subcommand + help, `daemon/run.ts` dispatch, `AgentId`, `BackendFlavor`, `controlServer` zod enum, `registerCommonHandlers`.
- **happy-agent:** `SupportedAgent`, `SUPPORTED_AGENTS`.
- **happy-app:** new-session picker (`agentDefaults`, `persistence`, picker + icon, settings labels, `storageTypes` cliAvailability, `ops` spawn unions, `AgentInput`).

## Tests

`agy/cliArgs.test.ts` + `agy/AgyBackend.test.ts` (10 tests, incl. the stdin-EOF guard). Full happy-cli unit suite passes (691/691); happy-cli, happy-agent, and happy-app all typecheck.

## Notes / follow-ups (not done here)

- The happy-app agy icon is a **placeholder** (copied from the Gemini icon) — wants a real asset.
- **Retiring the Gemini integration** is intentionally out of scope (this PR is additive).
- A richer integration via agy's embedded `agentapi` HTTP server (structured tool / permission events) is possible later, but it needs an undocumented running language-server — out of scope.

## Visual proof

agy driven end-to-end from the **Happy app**, not just unit tests: a Happy session running the `agy` backend — online, answering a prompt and correctly identifying itself as **Antigravity (Gemini 3.1 Pro)**. This shows the backend spawns, streams its output back into Happy's `AgentMessage` pipeline, and is usable from the app/browser.

Screenshot 2026-06-24 at 11 41 05 PM

---

Closes #1429
Refs #1313


charliezong18:
hey @bra1nDump — thanks again for the quick turnaround on #1440! 🙏 whenever you've got bandwidth, #1430 (agy / Antigravity CLI backend) is also ready for a look. it's bigger (~1.2k lines) but fully **additive** — the gemini provider is untouched, it just registers `agy` everywhere `openclaw` already appears. tests pass, all three packages typecheck, and there's end-to-end visual proof in the description (driven from the app).

i wrote up the design rationale in #1429 (tl;dr: agy has no ACP/JSON mode, so it follows the bespoke `openclaw` pattern rather than the gemini ACP one). totally open to splitting it up or reworking the approach if the size/shape is a concern — just let me know what'd make it easiest to review.

bra1nDump:
*Written by AI (Claude), reviewed by @bra1ndump.*

Thanks — backend code is clean and the demoed flow works. Three things to verify before merge; the bar for a new agent backend is a bit higher than a one-shot Q&A:

**1. Is `--print` really the only surface?** You mention agy's embedded `agentapi` server. How was ACP/structured output ruled out? Plain-text scraping permanently loses tool events and permission prompts for this backend, so "no ACP" should be a confirmed fact, not an assumption.

**2. Multi-turn, live.** Resume rests on an undocumented cache file keyed by cwd. Show a real session with 3+ turns retaining context. (Also: two agy sessions in the same cwd will cross-resume each other's conversations.)

**3. Tool calling + permissions.** The demo needed no tools. Show a file-edit turn in both `--sandbox` and skip modes, and check what agy's permission surface actually is in print mode — when it wants an approval under `--sandbox`, does it deny, error, or hang until the watchdog kills it?

Pushed a commit to this branch (4e81a0a4) wiring the app-side model/permission pickers for agy — they previously fell through to Claude's lists, so any non-default model selection would have sent `--model "opus"` to agy and failed the turn.

charliezong18:
thanks for the thorough review — took the three asks as an experiment checklist. everything below was re-verified against agy **1.0.12** (the CLI auto-updates; the PR was originally tested on 1.0.10).

## 1. `--print` really is the only conversation surface — now a verified fact

- The full 1.0.12 flag list has **no** `--output-format` / JSON / stream / ACP / server flag:

  ```
  --add-dir, -c/--continue, --conversation, --dangerously-skip-permissions,
  -i/--prompt-interactive, --log-file, --model, --new-project, -p/--print,
  --print-timeout, --project, --prompt, --sandbox
  subcommands: changelog, help, install, models, plugin(s), update
  ```

- The embedded **agentapi does exist** — as a *hidden* `agy agentapi` subcommand (plus a standalone `~/.gemini/antigravity-cli/bin/agentapi` binary intended for agy's own background sidecars). Its entire command surface is:

  ```
  get-conversation-metadata <conversation_id>
  new-conversation [--model=<flash_lite|flash|pro>] [--title=<title>] <prompt>
  send-message [--title=<title>] <recipient_id> <content>
  ```

  i.e. task injection + metadata lookup. There is no subscribe/stream/events command, so tool events and permission prompts cannot be received through it either.

- Conversation storage is one opaque SQLite db per conversation (`~/.gemini/antigravity-cli/conversations/<id>.db`). No `transcript.jsonl` (or anything similar) is written under the CLI data dir by print runs — the `transcript.jsonl` strings inside the binary belong to the cloud "Cascade" artifact bundle format, not the local CLI.

So plain-text `--print` scraping is the only conversation I/O available today. If agy ever opens up the agentapi surface properly, upgrading the backend to ride it is the natural follow-up (already noted in the PR description).

## 2. Multi-turn, live — and the cross-resume bug you called

You were right about cross-resume, and it was worse than described: `startSession()` seeded from the per-cwd cache, so a brand-new Happy session would silently *resume whatever conversation last ran in that cwd*; and every turn-close re-read the cache, so a concurrent same-cwd session could hijack an already-pinned conversation mid-session.

Fixed in `e1c5556`:
- fresh sessions never seed from the cache — they start a fresh agy conversation;
- while unpinned, the id is adopted only when the cache entry *changes* relative to a snapshot taken before each turn, then pinned;
- once pinned, the cache is never re-read.

Four new unit tests cover exactly those scenarios (fresh-start guard, stale-entry adoption, per-turn re-snapshot so a foreign write between turns isn't adopted, foreign-update immunity); full suite 699/699.

Two things worth stating plainly:
- **Residual window:** while a session is still unpinned, any concurrent same-cwd cache write that differs from its pre-turn snapshot gets adopted — e.g. two brand-new sessions racing their first turns, or a bare interactive `agy` run in the same cwd finishing mid-turn. Documented in the code; a real fix needs agy to echo the conversation id in print mode, which it doesn't today.
- **Deliberate behavior change:** restarting Happy in a cwd no longer auto-resumes the previous conversation — that convenience rode the exact mechanism that caused the cross-resume bug. Proper cross-restart resume needs explicit id plumbing (listed as a follow-up in the PR description).

**Live demo (screenshots):** a 5-turn session driven from the app — sets a fact ("my favorite number is 47"), asks an unrelated question, recalls the fact, runs a real tool call (listing the workspace files), then recalls the fact again. Then a **second agy session in the same cwd** asks the same question and correctly knows nothing about it (the cross-resume fix at work — before the fix it would have resumed the first session's conversation).

5-turn session: fact set, unrelated turn, fact recalled, file-listing tool call, fact recalled again

second session in the same cwd knows nothing about the first session's context

## 3. Tool calling + permissions — empirical matrix, honest disclosure

Ran real file-edit and shell turns headlessly through `--print` (the exact invocation the backend uses, incl. `--add-dir <cwd>`), across agy's documented `toolPermission` settings (`always-proceed` / `request-review` = default / `strict`) × `--sandbox` vs `--dangerously-skip-permissions`:

| settings `toolPermission` | flag | in-workspace file edit | shell command | outside-workspace write |
|---|---|---|---|---|
| always-proceed | `--sandbox` | ✅ executed | ✅ executed | — |
| always-proceed | `--dangerously-skip-permissions` | ✅ executed | — | — |
| request-review (default) | `--sandbox` | ✅ executed | ✅ executed | ✅ executed |
| request-review (default) | `--dangerously-skip-permissions` | ✅ executed | — | — |
| strict | `--sandbox` | — | ✅ executed | — |

("—" = not tested; every tested cell executed with exit 0 in ~10–15s.)

Answer to "deny, error, or hang?": **none of the three — print mode auto-proceeds.** Every operation we could construct — including shell commands under `strict` and writes *outside* the workspace — executed without any approval gate. In print mode there is effectively no interactive permission surface; `--sandbox`'s "terminal restrictions" never visibly blocked anything we tested. So the permission-mode → flag mapping in `cliArgs.ts` is best-effort hygiene, not a real gate.

I think this deserves a user-facing note (README + the app's agy description): *"agy turns auto-approve tool use; Happy permission modes only choose between agy's `--sandbox` and `--dangerously-skip-permissions` flags."* Happy to add that to this PR if you agree it belongs here.

charliezong18:
Following up on the agentapi question from this thread — I spiked it properly.
TL;DR: it's not an upgrade path today.

What `agy agentapi` actually is (verified against agy 1.1.7, read-only):

- Not a server Happy can call: `agy agentapi server` → `{"error":
  "ANTIGRAVITY_LS_ADDRESS is not set"}`. It's a ConnectRPC/gRPC-over-mTLS
  **client** that dials out to a running Antigravity language server.
- The command surface is task injection + metadata only: `new-conversation`,
  `send-message`, `get-conversation-metadata`. No subscribe/stream command and
  no permission channel — the only wait primitive is
  `WaitForConversationFullyIdle` (block-until-idle).
- The env vars it needs (`ANTIGRAVITY_LS_ADDRESS`, `ANTIGRAVITY_CSRF_TOKEN`,
  `ANTIGRAVITY_SIDECAR_WEB_PORT`) are injected by the language server into
  sidecars it spawns and are not persisted anywhere — an external process like
  Happy's daemon has no supported discovery path.

Capability-wise vs today's `--print`: no streaming (worse — `--print` at least
forwards stdout chunks as they arrive), no per-tool permission events, no
attachments, and model selection degrades to `flash_lite|flash|pro`. The only
real wins are an explicit conversation id and structured `{error}` JSON — and
the conversation-id gap is addressed via metadata persistence in #1574.

So `--print` remains the better surface for now. Worth revisiting if agy ships
a streaming/subscribe RPC or a permission-event channel.

PR #1437, opened 2026-06-24
fix(app): widen user message bubble on native (CJK 1-char-column collapse)

chphch:
On iOS/Android, a multiline **user** message — most visibly space-less CJK text (Korean/Japanese/Chinese) — collapses into a tall, ~1-character-wide vertical column against the right edge. A numbered list, for example, renders as just the `1.` / `2.` markers stacked down a thin sliver.

The user-message bubble is content-fit via `alignItems: 'flex-end'` on `userMessageContainer` (column cross-axis), while `MarkdownView` wraps its content in `<View style={{ width: '100%' }}>`. A `width: '100%'` child inside a shrink-wrapped parent is circular, so Yoga sizes the bubble from the content's **min-content** width — and for space-less CJK every glyph is a valid break point, so min-content is one glyph wide. Browsers lay text out differently, so this only reproduces on native (iOS + Android), not web.

This lays the bubble out as a **row** item (`justifyContent: 'flex-end'` on the container) so Yoga sizes it from max-content clamped to `maxWidth: '85%'` (+ `flexShrink: 1`) — the standard chat-bubble pattern. Agent bubbles (a stretch container) were already fine; web is unaffected and now also caps at 85%.

## Proof

Reproduced and verified fixed on a real Android device (self-hosted build): a Korean numbered-list message that rendered as a 1-glyph-wide column (only the list markers visible) now wraps to the normal bubble width (≤ 85%). `pnpm typecheck` passes. This is a native-only Yoga layout bug, so the web test harness can't surface it (browser text layout doesn't collapse to CJK min-content).

### DIFF
--- a/packages/happy-app/sources/components/MessageView.tsx
+++ b/packages/happy-app/sources/components/MessageView.tsx
@@ -275,9 +275,8 @@ const styles = StyleSheet.create((theme) => ({
     overflow: 'hidden',
   },
   userMessageContainer: {
-    maxWidth: '100%',
-    flexDirection: 'column',
-    alignItems: 'flex-end',
+    width: '100%',
+    flexDirection: 'row',
     justifyContent: 'flex-end',
     paddingHorizontal: 16,
   },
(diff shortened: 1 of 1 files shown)


chphch:
Closing this out honestly: while capturing before/after proof on current `main` (RN 0.83.1, Android emulator, arm64 debug JS), I could no longer reproduce the collapse. The min-content sizing this PR works around is still present in `MessageView.tsx` (`userMessageContainer` → `alignItems: 'flex-end'`), but both a normally-spaced Korean message and the worst case — a space-less Korean run — now render at normal bubble width. The React Native / Yoga upgrade since this PR was written appears to have fixed the underlying min-content text measurement, so there is nothing left for this diff to fix on main.

Verified on Android emulator only (iOS untested); if the collapse resurfaces on a real device I'll reopen with a fresh repro.

PR #1440, opened 2026-06-24
fix(app): sort session list by last activity instead of creation time

charliezong18:
## Problem

In the left sidebar, the session list is sorted — and date-grouped into **Today / Yesterday / N days ago** — by `createdAt`. As a result a session **never moves once created**: chatting in an older session does not bubble it to the top or into the **Today** group. It sits frozen at its creation slot.

This contradicts the universal chat-list expectation (most-recently-used on top) and the app's own other recency views, which already sort by `updatedAt` (e.g. `storage.ts` ~L1496). The server even maintains a `Session_accountId_updatedAt` DESC index for exactly this.

## Fix

Sort and group sessions by `updatedAt` instead of `createdAt`, in both list builders:

- `buildSessionListViewData` — the grouped Today/Yesterday sidebar (sort + the date-bucketing that drives the headers).
- `applySessions` — the legacy flat online/offline list (kept consistent).

`activeAt` was deliberately **not** used: it's a presence-heartbeat timestamp that ticks every few seconds for online sessions, which makes the whole list re-shuffle continuously. `updatedAt` only advances on real session changes, so the order is stable when idle and bubbles a session up when it actually gets new activity.

Message ordering within a session (sorted by `createdAt`) is intentionally left untouched.

## Visual proof

Built and ran `happy-app` web locally against production data.

**Before** (current `main`, sorted by `createdAt`): an older session that I just chatted in stays frozen in its old "N days ago" group — it does not move.

**After** (this PR, sorted by `updatedAt`): the same session jumps to the **Today** group / top, and the list no longer re-shuffles while idle.

## Test

- `pnpm --filter happy-app typecheck` clean.
- Manual UAT on the running web app: confirmed the reorder behavior above and that an idle sidebar stays stable.

### DIFF
--- a/packages/happy-app/sources/app/(app)/settings/features.tsx
+++ b/packages/happy-app/sources/app/(app)/settings/features.tsx
@@ -18,6 +18,7 @@ export default function FeaturesSettingsScreen() {
     const [fileDiffsSidebar, setFileDiffsSidebar] = useSettingMutable('fileDiffsSidebar');
     const [groupToolCalls, setGroupToolCalls] = useSettingMutable('groupToolCalls');
     const [expImageUpload, setExpImageUpload] = useSettingMutable('expImageUpload');
+    const [sortSessionsByActivity, setSortSessionsByActivity] = useSettingMutable('sortSessionsByActivity');
 
     return (
         <ItemList style={{ paddingTop: 0 }}>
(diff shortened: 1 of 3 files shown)


charliezong18:
Hey @bra1nDump 👋 small one whenever you have a sec.

This fixes the sidebar session list being sorted (and date-grouped into Today/Yesterday/…) by `createdAt`, so a session chatted-in days later never bubbled to the top or into **Today** — it stayed frozen at its creation slot. Now sorts + groups by `updatedAt`, matching the recency ordering already used elsewhere in `storage.ts` and the server's `Session_accountId_updatedAt` index. (`activeAt` was avoided on purpose — it's a presence heartbeat and churns the list every few seconds.)

`happy-app` typecheck clean; verified on the running web app. Happy to tweak anything 🙏

bra1nDump:
hi @charliezong18! thanks for the pr :)

i don't mind changing the sorting for some people, but the sorting by creation time was intentional to prevent the rows from jumping around

you can introduce it but behind a flag in settings and default it to current sorting by creation date

charliezong18:
Done! Put it behind a new `sortSessionsByActivity` setting, defaulting to `false` so creation-date ordering stays the default (rows won't jump around). Opt-in via **Settings → Features → Interface → "Sort by Recent Activity"** — when on, both the session list and the Today/Yesterday grouping use `updatedAt`. happy-app typechecks clean. 🙏

bra1nDump (approved):

Issue #1386, opened 2026-06-11
Cannot select and copy output


venkuht1997:
When Claude Code returns something, I can’t copy that output in Happy. When I select I see a copy button but it randomly copies a portion of the output that I want, it doesn’t let me select the text I want to copy.

Issue #841, opened 2026-03-10
Mobile: no way to select, highlight, or copy text — markdownCopyV2 only copies entire response


Thoroslives:
## Problem

On mobile (iOS/Android), there is **no way to select, highlight, or copy any text** from a response. Standard text selection (long-press to get selection handles) simply doesn't work anywhere in the message.

The only copy mechanism available is the experimental `markdownCopyV2` setting, which captures the **entire response** as raw markdown when you long-press. There's no way to copy a specific paragraph, a single code block, or any partial selection — it's all or nothing.

This is a basic usability gap. On web, native text selection works fine. On mobile, you're stuck.

## Steps to Reproduce

1. Open a conversation on the mobile app (iOS or Android)
2. Try to long-press any text to select/highlight it — nothing happens (or with `markdownCopyV2` enabled, the entire message is captured)
3. There are no copy buttons visible on code blocks either (hover-only, invisible on touch devices)
4. The only option is navigating to a text-selection page showing the full raw markdown of the entire response

## Expected Behaviour

- Ability to select and copy arbitrary text within a response (like any other mobile app)
- Or at minimum, per-block copy buttons that are visible/tappable on mobile

## Technical Notes

In `MarkdownView.tsx`:
- `markdownCopyV2` wraps the entire message in a `GestureDetector` with `Gesture.LongPress()`, passing `props.markdown` (the full response) — no per-block granularity
- Code block copy buttons use `onMouseEnter`/`onMouseLeave` for visibility — these events don't fire on touch devices, so the buttons stay hidden (`opacity: 0`, `pointerEvents: 'none'`)

## Environment

- Happy App (React Native / Expo)
- Affects iOS and Android
- Web works fine (native text selection + hover copy button)

Issue #532, opened 2026-02-04
Copy function for Happy Code responses not working


sherkevin:
I hope the responses generated by Happy Code can be copied. Currently, the copy function is not working. I'm not sure if this is an issue specific to my device or a general problem.

Issue #311, opened 2025-12-30
Allow copying text from chat messages on Android


beckot:
On Android, chat message text cannot currently be selected or copied. This makes it hard to reuse generated content such as code snippets, commands, or short excerpts elsewhere.

It would be useful to support basic text selection and copy to clipboard inside chat messages, following standard Android text behavior.

Expected behavior:
- Long-press to select text inside a message
- Copy selected text to clipboard

Even partial selection (not only full-message copy) would already be a clear improvement.

Rouliooooo:
+1, and similar to [#187](https://github.com/slopus/happy/issues/187)

PR #1449, opened 2026-06-26
fix(app): allow partial text selection on the markdown copy screen

chphch:
**Reported by 7 people over 6 months** — issues #311, #532, #841 (open, 3 👍), #919, #1386, plus two in Discord. #841 and #1386 name this exact screen:

> "There's no way to copy a specific paragraph, a single code block, or any partial selection — it's all or nothing. This is a basic usability gap."
> — Thoroslives, #841, 2026-03-10

<details>
<summary>All 7 reports</summary>

|---|---|---|---|
| beckot | 2025-12-30 | #311 | "chat message text cannot currently be selected or copied … Even partial selection would already be a clear improvement" |
| sherkevin | 2026-02-04 | #532 | "I hope the responses generated by Happy Code can be copied. Currently, the copy function is not working" |
| miohtama | 2026-02-12 | Discord | "I would like to make the chat text selectable and copyable" |
| Thoroslives | 2026-03-10 | #841 (open) | see above |
| WhymustIhaveaname | 2026-03-26 | #919 | "it seems impossible to copy text from the conversation on mobile" |
| b.r.z. | 2026-03-30 | Discord | "Copy is broken … Also no way to copy from within the in line code sections as far as I can see" (one of his top-3 iOS complaints) |
| venkuht1997 | 2026-06-11 | #1386 | "it randomly copies a portion of the output that I want, it doesn't let me select the text I want to copy" |

</details>

---

The long-press "copy markdown" screen (`text-selection.tsx`) rendered the raw markdown in a read-only `<TextInput editable={false}>`. On Android a non-editable `TextInput` is not selectable at all, so the only copy path was the "Copy All" header button — there was no way to grab just a paragraph or a single line. This swaps it for a selectable `<Text>`, which brings up the native selection handles + copy toolbar on Android and iOS, so the markdown source is now partially selectable and copyable. The "Copy All" header button is unchanged.

This addresses the most acute part of #841 ("markdownCopyV2 only copies entire response"): once you're on the text-selection page you can now copy a portion instead of all-or-nothing. The broader ask in #841 — inline arbitrary selection directly on the rendered message and touch-visible per-block copy buttons — is left to a follow-up, since `markdownCopyV2` intentionally disables inline span selection (so the native copy menu doesn't race the long-press gesture).

### Verification
This is Android/iOS-native selection behaviour — the broken state (a non-editable `TextInput` being unselectable) only manifests on a touch device; on web the text is always selectable, so the web test harness can't exercise it. Checked: `pnpm typecheck` clean, and the change is the documented RN pattern for selectable read-only text (`<Text selectable>`). On-device repro: long-press a message → "copy markdown" screen → drag the selection handles over part of the text → native **Copy**.

### DIFF
--- a/packages/happy-app/sources/app/(app)/text-selection.tsx
+++ b/packages/happy-app/sources/app/(app)/text-selection.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Platform, View, Text, ScrollView, TextInput, Pressable } from 'react-native';
+import { Platform, View, Text, ScrollView, Pressable } from 'react-native';
 import { useRouter, useLocalSearchParams, useNavigation } from 'expo-router';
 import { StyleSheet, useUnistyles } from 'react-native-unistyles';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
@@ -97,17 +97,19 @@ export default function TextSelectionScreen() {
                     { paddingBottom: insets.bottom + 16 }
                 ]}
             >
-                <TextInput
-                    style={[styles.textInput, { 
-                        color: theme.colors.text,
-                        backgroundColor: 'transparent'
-                    }]}
-                    value={fullText}
-                    multiline={true}
-                    editable={false}
-                    selectTextOnFocus={false}
-                    scrollEnabled={false}
-                />
+                {/*
+                    Render as a selectable <Text>, not a read-only <TextInput>.
+                    On Android a TextInput with editable={false} cannot be
+                    selected at all, so the only way to copy was the "Copy All"
+                    header button. A selectable <Text> brings up the native
+                    selection handles + copy toolbar, enabling partial copy.
+                */}
+                <Text
+                    selectable={true}
+                    style={[styles.sourceText, { color: theme.colors.text }]}
+                >
+                    {fullText}
+                </Text>
             </ScrollView>
             </MobileGlassSurface>
         </View>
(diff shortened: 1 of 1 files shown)


chphch:
@bra1nDump — surfacing this one for the demand behind it. Partial text selection on the copy screen is the most-requested fix among my open PRs: seven independent reporters across #311, #532, #841, #919, #1386 and Discord, stretching back ten months (catalogued in the PR description). The fix itself is small — one file, +15/−19: the screen's non-editable TextInput can't be partially selected on Android, so it becomes a selectable Text. Re-verified today: mergeable clean, typecheck green. If you'd rather solve it differently I'm happy to close in favor of that.

chphch:

Collecting what has accumulated around this one, since it looks like the demand is wider than this PR:

- **#841** (Mar 10, 3 👍) — "no way to select, highlight, or copy any text" on mobile; the only copy path grabs the whole response.
- **#1652** (Aug 5) — the same request for Android, independently filed.
- **#750** (Feb 24) — text selection in markdown chat, among other things.
- **#1527** (Jul 14) — @davidrimshnick-pplx opened a second PR for the same screen, ~3 weeks after this one.

This PR changes one file, `sources/app/(app)/text-selection.tsx` — the screen #841 is about.

Two open PRs for one screen is wasted review either way, so I'm happy to close this in favour of #1527 if you prefer that shape. @bra1nDump — could you point at whichever one you'd rather take?

PR #1451, opened 2026-06-27
fix(app): keep user-selection UI (<options>, AskUserQuestion) visible — never fold it into collapsed work groups

chphch:
> **Rewritten 2026-09-14 against current `main`.** The original implementation hooked into `visibleForToolGrouping` and `groupToolCallsForDisplay`, both of which `main` has since removed. It is also smaller now: `main` already keeps the `AskUserQuestion` card out of the fold, so only the `<options>` half and the chronological split remain.

A turn's intermediate agent work is folded into a collapsed-by-default "Worked for…" group. That also swallows the markdown `<options>` block — an `agent-text` message the chat renders as tappable chips — so the prompt ends up hidden behind a tap-to-expand header and the user cannot see or answer it.

`main` already excludes the `AskUserQuestion` tool card from the foldable set (`collectAgentWorkGroups`, the `isInteractiveQuestionToolName` line). An `<options>` block is not a tool call, so nothing excludes it.

### Fix

`isUserSelectionMessage()` recognizes a user-selection element — an `agent-text` whose line starts with `<options>` (mirroring `parseMarkdownBlock`'s trigger), or a tool call the existing `isInteractiveQuestionToolName` already names. `collectAgentWorkGroups` then walks the foldable candidates and **splits the fold at each selection element** rather than dropping it into the group:

- the selection element itself is never a member of a group, so it renders standalone and tappable;
- work on either side of it becomes its own group, so a selection answered mid-turn stays in its true chronological position instead of being pushed to one side of a single merged group. Each group's "Worked for…" label ends at the boundary that split it.

### Behaviour

| Before | After |
| --- | --- |
| `<options>` older than the turn's final text → folded into the collapsed group, hidden until tapped | rendered standalone, always visible and tappable |
| selection element between two runs of tool work → both runs merged into one group, the element pushed to one side | fold split at the element; chronological order preserved |

### Testing

`packages/happy-app` — `pnpm typecheck` clean; `useGroupedMessages.test.ts` 13/13. Four cases are new: an `<options>` block kept visible, the interleaved split, an `AskUserQuestion` card kept visible, and a false-positive guard so an inline `` `<options>` `` mentioned in prose still folds.

### DIFF
--- a/packages/happy-app/sources/hooks/useGroupedMessages.test.ts
+++ b/packages/happy-app/sources/hooks/useGroupedMessages.test.ts
@@ -400,6 +400,191 @@ describe('useGroupedMessages', () => {
         expect(items.map((item) => item.id)).toEqual(['agent-final', 'user']);
     });
 
+    it('keeps an <options> block visible instead of folding it into the work group', () => {
+        const messages: Message[] = [
+            {
+                kind: 'agent-text',
+                id: 'agent-final',
+                localId: null,
+                createdAt: 4,
+                text: 'done',
+            },
+            {
+                kind: 'agent-text',
+                id: 'agent-options',
+                localId: null,
+                createdAt: 3,
+                text: 'Pick one:\n<options>\n    <option>A</option>\n    <option>B</option>\n</options>',
+            },
+            toolMessage('tool-earliest', 2),
+            {
+                kind: 'user-text',
+                id: 'user',
+                localId: null,
+                createdAt: 1,
+                text: 'run tools',
+            },
+        ];
+
+        const items = groupMessagesForDisplay(messages, true);
+
+        // The options message renders standalone; only the older tool folds.
+        expect(items.map((item) => ({ type: item.type, id: item.id }))).toEqual([
+            { type: 'message', id: 'agent-final' },
+            { type: 'message', id: 'agent-options' },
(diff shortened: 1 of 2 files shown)


chphch:
**Running-app proof** (self-hosted web build, before/after on the same seeded session, `groupToolCalls` enabled):

before/after — options fold

- **Before** (pre-fix `useGroupedMessages`): the agent's `<options>` choice is folded into the collapsed **"Worked 0s"** group — the user can't see or tap it.
- **After** (this PR): the `<options>` block stays visible and tappable; only the real tool call folds into "Worked 0s".

Identical seeded messages were rendered on the pre-fix bundle vs this PR's bundle (same session). `AskUserQuestion` is excluded from folding the same way. Unit coverage: `pnpm test sources/hooks/useGroupedMessages.test.ts` → 14/14.

chphch:
@bra1nDump — flagging this as an interaction bug rather than polish: when message grouping collapses a work section, any `<options>` block or AskUserQuestion card inside it is folded away too — so the agent's question is hidden at exactly the moment it's waiting on the user's answer (proof GIF above). One caveat up front: at +257/−21 this is larger than my usual fixes. If you'd prefer a minimal version that only exempts the selection cards from folding, say the word and I'll cut it down to that. Re-verified today: mergeable clean, typecheck green.

PR #1461, opened 2026-06-29
feat(app): support attachments in the new-session composer

jlixfeld:
## Summary

The new-session screen's composer only accepted **text** — there was no way to attach an image to the very first message, unlike the in-session input (`AgentInput`) which already supports the `expImageUpload` attachment pipeline. Requested in #1319 / #1270 / #919 / #70 (those track attachments broadly; none of them call out the new-session composer specifically, which is a separate component from `AgentInput`).

This wires the same attachment flow into the new-session composer, matching the **in-session input's current capabilities** on `main` (gallery pick + web paste/drag — the camera/file chooser from the still-open #1387 is intentionally out of scope here).

## Changes

- **Picker button + preview strip** in the new-session composer action row, reusing `useImagePicker` and `AgentInputAttachmentStrip` (gated on the `expImageUpload` setting).
- **Persisted attachments** — staged attachments are stored in the new-session draft store so they survive navigating away from the screen, matching the already-persisted prompt text. The picker is seeded from the draft once on mount and mirrored back on every change.
- **Send path** — on spawn, the staged attachments are passed to `sync.sendMessage`. The session already exists at that point, so the existing upload→file-event→text path handles them exactly like an in-session message. Send now fires on **text OR attachments**.
- **Refactor** — `AgentInput`'s web paste/drag effect is extracted into a shared `useWebImagePaste` hook so both composers get identical clipboard-paste and file-drop behavior on web without duplication. `AgentInput` now calls the hook (no behavior change).
- **`parsePersistedAttachments`** — a pure, RN-free validator for the persisted draft blob (drops malformed/old-version entries), in its own module so it's unit-testable in Node.

## Tests

- `attachmentDraft.test.ts` — 6 cases: JSON round-trip, valid-without-thumbhash, non-string-thumbhash rejection, non-array input, malformed-entry filtering, fresh-array identity.
- `pnpm typecheck` clean.
- Full `happy-app` suite: **689 passed** (includes the `AgentInput` web-paste refactor regression surface).

## Test plan

- [x] Unit tests above
- [ ] Manual: new session → pick a gallery image → strip shows it → spawn → image arrives in the session as the first message
- [ ] Manual: stage an image → navigate away → return → image still staged (draft persistence)
- [ ] Manual (web): paste / drag-drop an image into the new-session screen → it stages

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -42,6 +42,9 @@ import { resolveAbsolutePath } from '@/utils/pathUtils';
 import { formatPathRelativeToHome, formatLastSeen } from '@/utils/sessionUtils';
 import { useNavigateToSession } from '@/hooks/useNavigateToSession';
 import { useNewSessionDraft } from '@/hooks/useNewSessionDraft';
+import { useImagePicker } from '@/hooks/useImagePicker';
+import { useWebImagePaste } from '@/hooks/useWebImagePaste';
+import { AgentInputAttachmentStrip } from '@/components/AgentInputAttachmentStrip';
 import { useShallow } from 'zustand/react/shallow';
 import type { MultiTextInputHandle } from '@/components/MultiTextInput';
 import { Modal } from '@/modal';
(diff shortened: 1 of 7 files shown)


charliezong18:
We've been running this downstream in our fork and it fills a real gap — even with `expImageUpload` on, there's currently no way to attach an image to the very first message from the desktop new-session screen. Confirmed working in daily use (desktop web).

A few notes from adapting it onto current `main` (as of 3c0155c2), in case they help a rebase:

- Part of this PR has effectively landed already: `useNewSessionDraft` now has `attachments`/`setAttachments`, and the new-session screen's send path already forwards staged attachments to `sync.sendMessage` (came in with the Home dock flow). What's still missing on `main` is exactly the UI entry (picker button, attachment strip, web paste/drop) and the `useWebImagePaste` extraction — a rebase should shrink this diff a lot.
- The new-session composer markup has since been rebuilt (`MobileGlassSurface`, mobile action row), so the render hunks need re-grafting rather than a straight rebase.
- One design wrinkle: `useNewSessionDraft` on `main` now documents that attachments intentionally stay out of MMKV persistence ("Image picker URIs are temporary..."). The `attachmentDraft.ts`/`persistence.ts` portion runs against that decision. Downstream we dropped the persistence piece and kept everything else — typecheck and the existing draft/attachment tests pass, and the in-memory store already covers the Home → New Session bridge within an app run.

Hope this gets picked up — happy to help test a rebased version.

chphch:
Heads up — I've opened #1593 carrying this forward, since this PR has been conflicting for a while and @charliezong18 asked for it downstream above.

Your commit is preserved as-is (authorship intact), rebased onto current `main`. Two parts of it are dropped as no longer needed: `main` has since landed `attachments` on `useNewSessionDraft` plus the send-on-spawn path, and the MMKV persistence half conflicts with the store's own "picker URIs are temporary" comment — so #1593 keeps attachments in memory only. The remaining piece is what's still missing upstream: the attach button, the preview strip, and web paste on the `/new` composer.

No ownership claimed here — if you'd rather rebase this PR yourself, do that and I'll close #1593.

Issue #1464, opened 2026-07-01
Manual `happy notify` works, but session-status notifications (Claude finished / waiting for input) are never delivered


carrot0117:
# Manual `happy notify` works, but session-status notifications (Claude finished / waiting for input) are never delivered

## Environment
- Happy CLI version: `1.1.9`
- Claude Code version: `2.1.195`
- Node: `node20` (`/opt/node20/bin`)
- OS (CLI host): Linux (remote container)
- Mobile: **iOS**
- Daemon: running, connected to `wss://api.cluster-fluster.com` (heartbeat OK)

## What works
- Manual push works reliably: `happy notify -p "..." -t "..."` returns `✓ Push notification sent successfully!` and the notification **is received** on the iPhone (lock screen / background).
- App-in-foreground: conversation content syncs correctly in real time.
- iOS system settings: notifications **allowed**, Background App Refresh **enabled**, Low Power Mode off.

## What doesn't work
- **Session-status notifications are never delivered.** When a session goes from `running` → idle (Claude finished / waiting for user input), no push arrives on the phone — even with the app backgrounded / locked.
- Confirmed reproducible: end a turn so the session becomes idle → no notification, while a manual `happy notify` sent at the same time **is** received.

## Diagnostics collected
- `happy doctor`: daemon PID running, connected to server, sessions tracked normally.
- Daemon log (`*-daemon.log`) shows `Session reported` / `Session webhook` / `lifecycleState: running` events, but **no automatic notification/push events are ever emitted** for session state changes. Only manual `notify` produces a delivered push.
- `/root/.happy/settings.json` shows `"onboardingCompleted": false` (pairing still functions).

## Analysis
Because manual `notify` is delivered but session-state notifications are not, the APNs delivery path is healthy. The gap is in the **auto-trigger for session-state notifications** — the daemon never emits them. Possibly related to an in-app "notify on completion / waiting" setting not being enabled, and/or `onboardingCompleted: false` leaving notification preferences uninitialized.

## Questions for maintainers
1. Is the "notify when session finishes / waits for input" behavior controlled by a mobile-app setting, a CLI flag, or server-side session state?
2. Should the daemon log auto-notification attempts? (Currently there is no trace of any.)
3. Could `onboardingCompleted: false` prevent notification preferences from being initialized?

---

### 环境
- Happy CLI `1.1.9`，Claude Code `2.1.195`，node20，CLI 运行在 Linux 远程容器；手机端 iOS。
[...]

Issue #1383, opened 2026-06-11
Happy not sending push notifications to iPhone when Claude Code needs attention


zackwwu:
### Description

When using Happy to manage a Claude Code session remotely from my iPhone, there's no push notification when Claude finishes or needs permission approval. I only find out by manually checking the app.

happy notify works — push notifications come through to my iPhone when run manually. The missing piece is Happy automatically sending a notification when Claude Code stops or requests permission, but only during remote/mobile sessions (not when working locally at the computer).

Device: iPhone (iOS)

djv56666:
same root cause as #1464 - server-side suppression, not a delivery failure.

when a remote session runs it opens a `session-scoped` socket that never reports `app-state`, and the server's active-user check treats that as a permanently-active client via an `undefined → active` fallback. so every done/permission push is silently dropped while the session runs, while manual `happy notify` goes through because it bypasses `dispatchSessionEventPush`. fix is to narrow that check to `user-scoped` (mobile/web) sockets only.

i have a local patch, happy to open a PR. thanks for reporting!

djv56666:
fix is up in #1471 - narrows the server's active-user check to `user-scoped` sockets only, so running coding sessions no longer count as "active" and stop suppressing done/permission pushes.

PR #1471, opened 2026-07-02
fix(server): exclude coding sessions from push suppression

djv56666:
## Summary

Mobile push notifications for `done` (task finished), permission, and question events were silently suppressed whenever a CLI coding session was running. The server's presence check counted any non-machine socket as "user is looking at a client"; but a running coding session opens a `session-scoped` socket that never reports `app-state`, so the `undefined → active` fallback marked it permanently active and blocked every mobile push for that user. This narrows the check to `user-scoped` sockets (mobile app / web) only.

## Proof it works

Self-hosted server, same repro: a `happy` coding session running in a terminal, phone backgrounded, trigger a task completion.

**Before** — server log:
```
Suppressed session-event push for user cmr... session cmr...: user active
```
Phone receives nothing. Daemon log confirms it was sent: `[PUSH] sendSessionNotification dispatched via server (kind=done)`.

**After** — server log (same repro):
```
Push sent for user cmr...: 1 token(s)
```
Phone receives the push.

## Root cause

`hasActiveNonMachineSocket` only excluded `machine-scoped` (daemon) sockets. A `session-scoped` coding-session socket combined with the `appState === undefined → active` fallback meant any running coding session read as a permanently-active UI client, suppressing all session-event pushes. Manual `happy notify` worked because it bypasses `dispatchSessionEventPush`.

## Changes

- `packages/happy-server/sources/app/events/eventRouter.ts`: `hasActiveNonMachineSocket` → `hasActiveUiClient`; only `user-scoped` sockets now count toward active.
- `packages/happy-server/sources/app/push/focusTracker.ts`: updated caller + doc comment.

## Notes

- This is a server change, which `contributing.md` asks to discuss first — that discussion happened in #1464 and #1383 (root-cause analysis posted there before opening this PR).
- Refs #1464, #1383. Not using `Fixes` so the reporters (on the hosted server) can still verify once this is deployed.

### DIFF
--- a/packages/happy-server/sources/app/events/eventRouter.ts
+++ b/packages/happy-server/sources/app/events/eventRouter.ts
@@ -281,16 +281,26 @@ class EventRouter {
     // === PRESENCE QUERIES ===
 
     /**
-     * Returns true if the user has any non-machine socket that hasn't
-     * reported `app-state: background`.  Old clients that never send
-     * `app-state` are treated as active (connected = present).
+     * Returns true if the user has any *UI* client (mobile app / web) that
+     * hasn't reported `app-state: background` — i.e. someone is looking at a
+     * Happy client right now, so a push would be redundant.
+     *
+     * Only `user-scoped` sockets count. `machine-scoped` (CLI daemon) and
+     * `session-scoped` (CLI coding sessions) are background processes that
+     * never surface notifications, so they must NOT suppress push — otherwise
+     * simply having a coding session running silently blocks every mobile
+     * push for the user. Old UI clients that never send `app-state` are
+     * treated as active (connected = present) for backwards compatibility.
      *
      * Uses fetchSockets() which works cross-replica via Redis streams adapter.
      */
-    async hasActiveNonMachineSocket(userId: string): Promise<boolean> {
+    async hasActiveUiClient(userId: string): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


zzlatev-openclaw-neo:
Confirmed this exact bug on a self-hosted Happy deployment with the stock Android app **v1.7.0**.

Server-log analysis showed:

- 20 `POST /v1/sessions/:id/push-event` requests received successfully
- 20/20 logged as `Suppressed session-event push ...: user active`
- 0 `Push sent`, `Push partial`, or dispatch-failure entries
- Push-token registration succeeded repeatedly via `POST /v1/push-tokens`

The clearest proof was this sequence:

1. Android `user-scoped` socket disconnected.
2. A session push event arrived while Android remained disconnected.
3. The server still suppressed it as `user active`.
4. Android reconnected afterward.

This rules out Android foreground state as the cause. The remaining `session-scoped` coding-session socket was being treated as an active UI client because `hasActiveNonMachineSocket()` excluded only `machine-scoped` sockets and interpreted missing `appState` as active.

This matches the root cause and proposed fix in #1471: only `user-scoped` UI sockets should participate in push-suppression decisions.

Android notification permission was enabled, and the device Expo token was successfully registered. The server never reached the Expo send path because every event was suppressed first.

Thanks for the fix—this is reproducible and currently makes session-event notifications effectively nonfunctional on this self-hosted deployment.

PR #1472, opened 2026-07-02
fix(server): don't count CLI coding sessions as user presence for push

StefanGilligan:
## Problem

Session-event push notifications (agent finished, permission request, question) are suppressed whenever a coding session is connected — which is essentially always — so mobile devices receive no push notifications.

`EventRouter.hasActiveNonMachineSocket()` decides whether to skip a push because the user is already looking at Happy. It treated **every non-`machine-scoped` socket** as presence, which includes `session-scoped` sockets. But a `session-scoped` socket is a CLI coding session (the agent, not the user), and it stays connected for the entire life of the session. So the very session dispatching a "done" push is itself counted as active presence and suppresses its own notification.

## Fix

Count only `user-scoped` connections (the user's mobile / web / desktop clients) as presence. `session-scoped` (agent) and `machine-scoped` (daemon) sockets are excluded. A `user-scoped` client is still treated as active unless it has reported `app-state: background`.

## How it was found / verified

On a live server, with the old logic the phone reported `app-state: background` yet the push was still `Suppressed … user active`, because ~10 `session-scoped` (`cli-coding-session`) sockets with no `app-state` counted as presence. With the fix, the same push is delivered.

Added `eventRouter.spec.ts` covering: session-scoped and machine-scoped sockets are ignored; a user-scoped client counts when foreground or before it reports state; and — the exact regression — a backgrounded user-scoped client is not active even with session sockets present.

### DIFF
--- a/packages/happy-server/sources/app/events/eventRouter.spec.ts
+++ b/packages/happy-server/sources/app/events/eventRouter.spec.ts
@@ -1,0 +1,46 @@
+import { describe, it, expect } from 'vitest';
+import type { Server } from 'socket.io';
+import { eventRouter } from './eventRouter';
+
+// hasActiveNonMachineSocket only touches `io.in(room).fetchSockets()`, so a tiny
+// stub is enough — it ignores the room and returns the provided sockets.
+function stubIo(sockets: Array<{ data: Record<string, unknown> }>): Server {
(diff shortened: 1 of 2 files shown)


seliver:
This works for me.

PR #1476, opened 2026-07-03
fix(app): stop session list jumping during multi-session streaming

charliezong18:
## Problem

Follow-up to #1440. That PR added an opt-in "sort sessions by activity" mode, but it sorts on `session.updatedAt` — which bumps on **every** background agent update. With several sessions streaming at once, the list reorders continuously and jumps under the user (the most active background session keeps stealing the top spot). It makes the sidebar unusable while multiple agents are running.

## Fix

Sort on **the last user-sent message time**, not `updatedAt`. Background output no longer reorders the list — a session only moves to the top when *you* act on it (send a message).

- New local-only `Session.lastMessageSentAt`, stamped in `sync.sendMessage` via a new `markSessionMessageSent` action.
- Both sort sites use `lastMessageSentAt ?? createdAt` when `sortSessionsByActivity` is on (falls back to creation order for sessions you haven't messaged yet).
- Persisted to MMKV (like drafts / model modes) so ordering survives an app restart.
- Not synced to the server; behavior stays behind the existing default-off `sortSessionsByActivity` setting.
- `deleteSession` now also clears the session's `lastMessageSentAt` from MMKV (matching how it already clears drafts / model modes / effort levels), and passes `unreadSessionIds` when rebuilding the list so remaining sessions keep their unread badges.

## Why `lastMessageSentAt` and not `updatedAt`

`updatedAt` reflects *any* change including streamed agent tokens, so it's inherently high-frequency and shared across all live sessions. "Most recently used" — which is what the sort is trying to express — is really "most recently *acted on* by the user", and that only changes on send. This keeps the list stable during heavy multi-session streaming.

## Testing

`pnpm --filter happy-app typecheck` passes; `pnpm --filter happy-app test` → 682 pass (1 unrelated pre-existing `settings.spec` snapshot failure on the base branch). Behavioral repro (list stops jumping while multiple sessions stream) is best seen live — happy to record a clip if useful.

### DIFF
--- a/packages/happy-app/sources/sync/persistence.ts
+++ b/packages/happy-app/sources/sync/persistence.ts
@@ -196,6 +196,23 @@ export function clearRegisteredPushToken() {
     mmkv.delete(REGISTERED_PUSH_TOKEN_KEY);
 }
 
+export function loadSessionLastMessageSentAt(): Record<string, number> {
+    const timestamps = mmkv.getString('session-last-message-sent-at');
+    if (timestamps) {
+        try {
+            return JSON.parse(timestamps);
+        } catch (e) {
+            console.error('Failed to parse session last message sent timestamps', e);
+            return {};
+        }
+    }
+    return {};
+}
+
+export function saveSessionLastMessageSentAt(timestamps: Record<string, number>) {
+    mmkv.set('session-last-message-sent-at', JSON.stringify(timestamps));
+}
+
 export function loadProfile(): Profile {
     const profile = mmkv.getString('profile');
     if (profile) {
--- a/packages/happy-app/sources/sync/storage.ts
+++ b/packages/happy-app/sources/sync/storage.ts
@@ -1067,10 +1067,12 @@ export const storage = create<StorageState>()((set, get) => {
             saveSessionLastMessageSentAt(allTimestamps);
 
             // Rebuild list view data — this timestamp drives activity-based sort.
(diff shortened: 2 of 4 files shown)


charliezong18:
@bra1nDump following up on your Discord question — no, this isn't a revert of #1440. The setting and feature stay; only the sort key changes. #1440 sorted on `updatedAt`, which bumps on every streamed agent token, so backgrounded sessions kept stealing the top spot. This PR adds a local-only `lastMessageSentAt` (stamped only when the user sends a message, persisted to MMKV) and sorts on that, falling back to `createdAt`. So "recently active" now means "recently acted on by me", not "most chatty agent".

Ready for review whenever you have a sec 🙏

charliezong18:
**Known limitation worth flagging upfront: this ordering is per-device.**

`lastMessageSentAt` is local-only (MMKV), so each device keeps its own map. Send a message from your phone and the session won't bubble up on web — and on web the map lives in localStorage, so clearing browser data resets the ordering. Same root cause: sends that don't go through this app's `sync.sendMessage` (another device, terminal) never stamp.

For a jump-fix that stays behind a default-off setting I think that's an acceptable trade, but the *proper* long-term shape is probably to derive "last user message time" from the message stream in the reducer — messages already sync, so ordering would be consistent across devices with no new persisted state. That touches the reducer/sync path though, so per the contributing guidelines I'd want to discuss before attempting it. Happy to file a follow-up issue if you think it's the right direction.

PR #1491, opened 2026-07-06
feat(cli): render agent tool-result images inline in the app

### LINKED ISSUE #128: Give Claude a tool to show me a image
Sometimes Claude makes data visualizations for me. If it had a tool that it could use to send a image file to my phone and the app embedded that into the ui that would be sick

ashleywr:
## Author note

I have been wanting this feature myself for a while and wondered why it wasn't working. So I was browsing with Claude through the open issues and tried to work backwards to find out what was missing. That's how we got to the solution presented in this PR. I admit I don't actually know TypeScript, but I did run a code-review, tested locally, and am happy to discuss any changes or concerns!

Tool results returned by the agent can contain base64 `image` content blocks — for example screenshots captured by MCP tools (Godot capture, browser automation, data-viz renderers). Previously these images never appeared in the app: the session protocol mapper emitted only a `tool-call-end` event for `tool_result` blocks and dropped any nested image content, so no `file` event was ever produced.

This adds support for surfacing those images inline in the conversation, which is exactly what folks have been asking for in #128 (and the "I want to see UI-test screenshots on mobile" comment there).

## Changes

All in `packages/happy-cli/src/api/apiSession.ts`:

- `extractToolResultImages()` — walks `tool_result` content and pulls out base64 `image` blocks (mirrors the existing `extractLocalTranscriptImageAttachments` helper).
- `uploadAndEnqueueToolResultImages()` — uploads each image as an encrypted attachment blob via the existing `uploadLocalImageAttachmentEnvelope` path and enqueues a `file` event.
- `sendClaudeSessionMessage()` — fires the upload as fire-and-forget when a message carries tool-result images, then proceeds through the unchanged synchronous mapper path. No changes to the mapper or to any callers.

The images ride the same end-to-end-encrypted blob pipeline already used for user→agent uploads, so no new server or crypto surface is introduced. Because the upload is async, a file event lands just after its `tool-call-end` rather than nested inside the tool result.

## Testing

- `tsc --noEmit` passes.
- Added a unit test in `apiSession.test.ts` that drives `sendClaudeSessionMessage` with a `tool_result` containing a base64 PNG and asserts the encrypted blob is uploaded and a `file` envelope with the right ref/name/size/mimeType is enqueued.
- Verified live: ran a Godot screenshot through a locally-linked build and confirmed the image renders inline in the app on device.

## Minor Notes

Display is not gated by `expImageUpload` — that flag only covers user→agent uploads. If gating agent→user images is desired, happy to add a flag check.

Closes #128

### DIFF
--- a/packages/happy-cli/src/api/apiSession.test.ts
+++ b/packages/happy-cli/src/api/apiSession.test.ts
@@ -542,6 +542,99 @@ describe('ApiSessionClient v3 messages API migration', () => {
         expect(decryptBlob(new Uint8Array(uploadBody), blobKey)).toEqual(pngBytes);
     });
 
+    it('uploads image blocks nested inside tool results as file events', async () => {
+        const client = new ApiSessionClient('fake-token', session);
+        const pngBytes = new Uint8Array([0x89, 0x50, 0x4E, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 4, 5, 6]);
+
+        mockAxiosPost.mockImplementation(async (url: string, payload: any) => {
+            if (url.endsWith('/attachments/request-upload')) {
+                expect(payload).toMatchObject({
+                    filename: 'tool-image-1.png',
+                });
+                expect(payload.size).toBeGreaterThan(pngBytes.length);
+                return {
+                    data: {
+                        ref: 'sessions/test-session-id/attachments/tool-image.enc',
+                        uploadUrl: 'https://server.test/v1/sessions/test-session-id/attachments/tool-image.enc',
+                        method: 'PUT',
+                    },
+                };
+            }
+
+            return {
+                data: {
+                    messages: payload.messages.map((_message: unknown, index: number) => ({
+                        id: `msg-${index + 1}`,
+                        seq: index + 1,
+                        localId: `local-${index + 1}`,
+                        createdAt: 1,
+                        updatedAt: 1,
+                    })),
+                },
+            };
+        });
+        mockAxiosPut.mockResolvedValueOnce({ data: { ok: true } });
+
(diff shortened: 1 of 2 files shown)


Jtome46:
+1 from a production user — I run Happy on iOS as my daily driver for Claude Code sessions on a headless machine, and this is the one gap I keep hitting: the mobile app is my only window into the session, and today every image the agent produces (screenshots from MCP tools, `Read` on an image file, data-viz renderers) is invisible on mobile.

For whoever reviews this, I dug into why it hasn't worked until now, and the gap really is exactly the one this PR fills:

- The wire protocol already has a `file` event (`sessionFileEventSchema` in happy-wire, including optional `image: {width, height, thumbhash}` metadata)
- The app already renders `file` events as inline agent-side attachment bubbles regardless of which side uploaded them — `typesRaw.ts` maps them to a `file` tool-call + paired tool-result
- The only missing piece is CLI-side extraction: `extractLocalTranscriptImageAttachments` deliberately skips messages containing `tool_result` blocks, so agent-produced images never became `file` events

Two small notes on the diff (not blockers):

1. `isSidechain` is filtered out, so images produced inside subagent transcripts won't surface. Looks intentional (matches the human-image path), just worth being aware of.
2. The upload path here doesn't set `ev.image` metadata — same as the existing user-image CLI path, so no thumbhash placeholder. Fine to leave, but if dimensions were computed at upload time the chat bubble would get a placeholder for free.

Linking #128, which is the oldest ask for this ("Give Claude a tool to show me an image"). That one's complementary: this PR covers agent-produced images automatically, while a file-path-based tool would let the agent deliberately push an arbitrary image to the phone.

Happy to test this against a real session with tool-heavy image workflows if that helps.

PR #1502, opened 2026-07-10
fix(cli): forward resolved permission mode to local Claude spawn

mshykhov:
## Summary

Local interactive sessions never entered the bypass permission mode that Happy resolved for them. The `claudeLocal` spawn builder forwarded `claudeArgs` verbatim but, unlike the remote/SDK path (`claudeRemote.ts`, which passes `permissionMode` to the SDK and gets `--permission-mode` emitted), it never translated the resolved mode into a Claude CLI flag. Since the default mode is `yolo` (bypass), a plain `happy` invocation started the local `claude` process in `default` - approval prompts still gated tool calls even though the session metadata (and the app UI) reported `dangerouslySkipPermissions: true`. This PR threads the resolved `permissionMode` down to the local spawn and injects the matching flag, so local sessions reach the same mode the remote path already does.

## Root cause

- `runClaude.ts` resolves `initialPermissionMode` (default `yolo` → bypass) and records `dangerouslySkipPermissions` in metadata.
- **Remote** path applies it first-class: `claudeRemote.ts` sets `permissionMode: mapToClaudeMode(...)` on the SDK, which emits `--permission-mode <mode>`.
- **Local** path (`claudeLocal.ts`) only did `args.push(...opts.claudeArgs)` and never emitted a permission flag from the resolved mode.
- The **sandbox** branch was unaffected only because it force-injects `--dangerously-skip-permissions`; plain non-sandbox local sessions hit the bug.

An explicit `--dangerously-skip-permissions` / `--permission-mode` typed on the CLI *did* survive (it rides `claudeArgs`), so the gap is every mode that comes from the default or `options.permissionMode` rather than an explicit CLI flag.

## Fix

Thread the resolved `permissionMode` through `Session → loop → claudeLocalLauncher → claudeLocal` (and the offline path in `runClaude`), then inject the matching flag in the spawn builder, next to the existing sandbox injection:

- `bypassPermissions` → `--dangerously-skip-permissions` (matches the sandbox branch and `claude --dangerously-skip-permissions` directly)
- `plan` / `acceptEdits` → `--permission-mode <mode>`
- `default` → nothing
- Skipped entirely when `claudeArgs` already carries an explicit permission flag (no duplicates, user override wins).

## Proof it works (real spawn)

Drove the real `claudeLocal` → real `cross-spawn` → real `claude_local_launcher.cjs` → a stub `claude` (via `HAPPY_CLAUDE_PATH`) that prints the argv it actually received - same evidence format as a `ps` on the live process.

**Before** (`happy@1.1.10`, plain local session) - no permission flag reached claude:
```
claude --append-system-prompt <happy title prompt> --mcp-config {...} --allowedTools mcp__happy__change_title --settings <hook>
```

**After** (this branch), real argv delivered to claude:

| Invocation | claude argv tail |
|---|---|
| plain `happy` (default `yolo`) | `… --dangerously-skip-permissions` |
| `permissionMode=bypassPermissions` | `… --dangerously-skip-permissions` |
| `permissionMode=plan` | `… --permission-mode plan` |
| `permissionMode=default` | *(no permission flag)* |
| `yolo` + explicit `--permission-mode plan` in args | `… --permission-mode plan` *(bypass not added - user override wins)* |

```
==== default happy (yolo) ====
STUB_CLAUDE_ARGV ["--session-id","…","--append-system-prompt","…","--dangerously-skip-permissions"]
==== plan mode ====
STUB_CLAUDE_ARGV ["--session-id","…","--append-system-prompt","…","--permission-mode","plan"]
==== default mode (no flag expected) ====
STUB_CLAUDE_ARGV ["--session-id","…","--append-system-prompt","…"]
==== yolo + explicit --permission-mode plan ====
STUB_CLAUDE_ARGV ["--session-id","…","--append-system-prompt","…","--permission-mode","plan"]
```

To reproduce interactively: `happy --yolo` (or plain `happy`), then in another terminal `ps -o args= -p <claude pid>` - the argv now carries `--dangerously-skip-permissions`.

## Testing

- `pnpm --filter happy test` - added 6 unit tests in `claudeLocal.test.ts` (inject bypass for yolo/bypass, `--permission-mode` for plan, no flag for default, no duplicate when a flag is already present, explicit-flag override). Full suite green.
- `tsc --noEmit` - clean.
- `pnpm --filter happy build` - passes.

## Scope

CLI-only, one focused bug fix. No sync engine / RPC / encryption / server changes.

### DIFF
--- a/packages/happy-cli/src/claude/claudeLocal.test.ts
+++ b/packages/happy-cli/src/claude/claudeLocal.test.ts
@@ -316,3 +316,135 @@ describe('claudeLocal --continue handling', () => {
         expect(spawnedArgs).not.toContain('--dangerously-skip-permissions');
     });
 });
+
+describe('claudeLocal permission mode forwarding', () => {
+    let onSessionFound: any;
+
+    beforeEach(() => {
+        mockSpawn.mockReturnValue({
+            stdio: [null, null, null, null],
+            on: vi.fn((event, callback) => {
+                if (event === 'exit') {
+                    process.nextTick(() => callback(0));
+                }
+            }),
+            addListener: vi.fn(),
+            removeListener: vi.fn(),
+            kill: vi.fn(),
+            stdout: { on: vi.fn() },
+            stderr: { on: vi.fn() },
+            stdin: { on: vi.fn(), end: vi.fn() },
+        });
+        onSessionFound = vi.fn();
+        vi.clearAllMocks();
+    });
+
+    const spawnedArgs = () => mockSpawn.mock.calls[0][1] as string[];
+
+    it('injects --dangerously-skip-permissions for the default yolo mode', async () => {
+        await claudeLocal({
+            abort: new AbortController().signal,
+            sessionId: null,
+            path: '/tmp',
+            onSessionFound,
+            claudeArgs: [],
+            permissionMode: 'yolo',
+        });
+        expect(spawnedArgs()).toContain('--dangerously-skip-permissions');
+    });
+
+    it('injects --dangerously-skip-permissions for bypassPermissions', async () => {
+        await claudeLocal({
+            abort: new AbortController().signal,
+            sessionId: null,
+            path: '/tmp',
+            onSessionFound,
+            claudeArgs: [],
+            permissionMode: 'bypassPermissions',
+        });
+        expect(spawnedArgs()).toContain('--dangerously-skip-permissions');
+    });
+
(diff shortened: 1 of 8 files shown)


mshykhov:
@bra1ndump small CLI fix for bypass permissions being dropped in local sessions - CI needs first-time contributor approval, mind taking a look? Thanks!

Issue #247, opened 2025-11-28
Android app hasn't been updated to see new models for codex


cruzanstx:
Android app hasn't been updated since September. Since then a few more models have been added. Also seems to fail MCP calls from codex.

PR #1504, opened 2026-07-10
feat(cli): publish Codex model list to session metadata

firesmasher231:
## Problem

The Codex model picker is a hardcoded array in `packages/happy-app/sources/components/modelModeOptions.ts`:

```ts
export function getCodexModelModes(): ModelMode[] {
    return [
        { key: 'default', name: 'default model', description: null },
        { key: 'gpt-5.5', name: 'gpt-5.5', description: null },
        { key: 'gpt-5.4', name: 'gpt-5.4', description: null },
        // ...
    ];
}
```

It tops out at `gpt-5.5`. Codex CLI 0.144.1 ships `gpt-5.6-sol`, `gpt-5.6-terra` and `gpt-5.6-luna`, and none of them are selectable from the app. Today the only workaround is to pick `default` and set `model` in `~/.codex/config.toml`.

Extending the array would fix it until the next model ships, and would need a client release to reach mobile users.

## What this does

`getAvailableModels()` already prefers backend-supplied `metadata.models[]` over the hardcoded array, and already re-adds a `default` option for Codex sessions:

```ts
const metadataModels = mapMetadataOptions(metadata?.models);
if (metadataModels.length > 0) {
    if (flavor === 'codex' && !metadataModels.some((model) => model.key === 'default')) {
        return [{ key: 'default', name: 'default model', description: null }, ...metadataModels];
    }
    return metadataModels;
}
return getHardcodedModelModes(flavor, translate);
```

Nothing was populating `metadata.models` on the Codex path — only the ACP path (`runAcp.ts`) does it, via `models_update`.

The Codex app-server protocol exposes a `model/list` RPC. This PR calls it once after `client.connect()` and publishes the result to session metadata. The picker then reflects whatever the installed `codex` actually supports — including models released after any given Happy version.

**This is CLI-only. No app change, no store release.**

## Changes

| File | Change |
|---|---|
| `src/codex/codexAppServerTypes.ts` | `CodexModel`, `ListModelsParams`, `ListModelsResponse` |
| `src/codex/codexAppServerClient.ts` | paginated `listModels()` (bounded at 10 pages) |
| `src/codex/utils/modelMetadata.ts` | `toMetadataModels()` — `{ id, displayName, description }` → `{ code, value, description }` |
| `src/codex/runCodex.ts` | `publishCodexModelMetadata()` after connect |
| `src/codex/utils/modelMetadata.test.ts` | 6 unit tests |

Design notes:

- **`default` is not synthesised.** The app adds it, and in Happy's Codex semantics it means "send no model override", which lets `~/.codex/config.toml` decide. Emitting a `default` entry here would change that.
- **`currentModelCode` is not set**, for the same reason — preselecting the backend's default model would turn a config-driven session into an explicit per-message override.
- **`hidden` models are filtered.** Those are the ones Codex itself keeps out of its picker.
- **Best-effort.** An older `codex` without `model/list` throws, we log at debug, `metadata.models` stays unset, and `getAvailableModels()` falls back to the hardcoded list. No regression for anyone.
- `DEFAULT_CODEX_MODEL` is left alone; it's still the fallback when the app sends no override at all.

## Verification

Against `codex-cli 0.144.1`, `model/list` returns 7 visible models:

```
id='gpt-5.6-sol'    displayName='GPT-5.6-Sol'    isDefault=True  hidden=False
id='gpt-5.6-terra'  displayName='GPT-5.6-Terra'  isDefault=False hidden=False
id='gpt-5.6-luna'   displayName='GPT-5.6-Luna'   isDefault=False hidden=False
id='gpt-5.5'        displayName='GPT-5.5'        isDefault=False hidden=False
id='gpt-5.4'        displayName='GPT-5.4'        isDefault=False hidden=False
id='gpt-5.4-mini'   displayName='GPT-5.4-Mini'   isDefault=False hidden=False
...
nextCursor: None
```

`id` and `model` are identical (both the slug); `id` is used as `code`.

- `tsc --noEmit` on `happy-cli`: clean (exit 0)
- `vitest run --project unit src/codex/`: **101 passed**, 17 files (6 new)

## Notes for reviewers

- The protocol types were derived from `codex app-server generate-json-schema` (`v2/ModelListResponse.json`), not hand-guessed.
- Pagination is implemented (`cursor` / `nextCursor`) and bounded at 10 pages so a misbehaving backend can't hang session startup.
- Related: #247 ("Android app hasn't been updated to see new models for codex"). This removes the need for an app release to surface *new* Codex models, so it addresses the model half of that report — but only for app builds that already contain the `getAvailableModels()` metadata path, and the MCP failures mentioned there are a separate problem.

### DIFF
--- a/packages/happy-cli/src/codex/codexAppServerClient.ts
+++ b/packages/happy-cli/src/codex/codexAppServerClient.ts
@@ -31,6 +31,9 @@ import type {
     RollbackConversationResponse,
     InjectItemsParams,
     InjectItemsResponse,
+    CodexModel,
+    ListModelsParams,
+    ListModelsResponse,
     ThreadGoalSetParams,
     ThreadGoalSetResponse,
     ThreadGoalClearParams,
@@ -192,6 +195,9 @@ function normalizeRawFileChangeList(changes: unknown): LegacyPatchChanges | unde
 
     return Object.keys(normalized).length > 0 ? normalized : undefined;
 }
+
+const MODEL_LIST_PAGE_SIZE = 100;
+const MODEL_LIST_MAX_PAGES = 10;
 
 export class CodexAppServerClient {
     private process: ChildProcess | null = null;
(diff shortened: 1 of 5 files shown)


firesmasher231:
hey @bra1nDump, just a small update to ensure models are available as soon as people have access to them on the CLI.

PR #1505, opened 2026-07-11
feat(cli): add a project-scoped Claude model pin

bogdanbaciu21:
Happy currently lets app-supplied model changes override a session's startup model, so a project that must stay on one Claude model can drift when controlled remotely. This adds an optional .happy-force-model file in the working directory; its first non-comment line pins that session's model at startup and remains authoritative over later controller updates. With no marker, behavior is unchanged.

## Running proof

I started a real Happy CLI session from a directory containing claude-sonnet-5 while passing the conflicting startup choice claude-opus-4-8. The session answered HFORK_FORCE_OK, and the runtime log emitted:

~~~text
.happy-force-model active: forcing model=claude-sonnet-5 (app model choices ignored)
~~~

## Current-base validation

The single contribution commit is rebased onto upstream 5e2dd1ae. Its range-diff and stable patch ID are unchanged, while upstream effort handling, Agy support, and GPT-5.6 remain ancestors.

- frozen workspace install passed
- focused forceModel suite: 5/5
- full Happy CLI unit suite: 746/746
- happy-wire build and full Happy CLI build passed
- diff contains only runClaude.ts plus the forceModel helper and test
- git diff --check passed

## Overlap

This is related to #1500 but has different semantics: #1500 supplies machine-wide startup defaults, while this marker is project-scoped and deliberately pins the model for the whole session. I can adapt it onto #1500 if maintainers prefer one configuration surface.

Refs bogdanbaciu21/dans-brain#10806 and bogdanbaciu21/dans-brain#15726.

### DIFF
--- a/packages/happy-cli/src/claude/runClaude.ts
+++ b/packages/happy-cli/src/claude/runClaude.ts
@@ -4,6 +4,7 @@ import { randomUUID } from 'node:crypto';
 import { ApiClient } from '@/api/api';
 import { logger } from '@/ui/logger';
 import { loop } from '@/claude/loop';
+import { readForcedModel, resolveForcedModel } from '@/claude/utils/forceModel';
 import { AgentGoalStatus, AgentState, Metadata } from '@/api/types';
 import packageJson from '../../package.json';
 import { Credentials, readSettings } from '@/persistence';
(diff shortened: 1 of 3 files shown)


bogdanbaciu21:
@bra1ndump Refreshed on the released current main `9a57219f` (Happy 1.2.0) at `152a1e82`; range-diff preserves the contribution byte-for-byte. The new CLI Smoke Test is again blocked before job creation by GitHub’s first-time-contributor approval gate: https://github.com/slopus/happy/actions/runs/29148186602. Could you approve it when convenient? Fresh local proof is green (746/746 CLI unit tests, focused 5/5, frozen install, and CLI build), with the live forced-model proof retained in the PR body. Thanks.

PR #1506, opened 2026-07-11
fix(cli): transcode HEIC attachments before Claude upload

bogdanbaciu21:
Happy's Claude CLI path currently accepts only JPEG, PNG, GIF, and WebP magic bytes, so HEIC/HEIF blobs from iPhone/iPad or older clients are omitted. This detects ISO-BMFF HEIF brands and converts the decrypted bytes to JPEG before creating the Claude image block; conversion failure stays fail-soft and preserves the existing skip behavior.

## Live running proof

From this rebased contribution commit, I started a real Happy remote session against the production relay. I uploaded the pinned 718,114-byte libheif HEIC fixture through the same encrypted file envelope and attachment endpoints used by the app, followed by a real text turn. The running CLI log showed the full relay-to-agent path:

~~~text
[API] Received file event { size: 718114, hasMimeType: true }
[loop] File event received: hfork011-proof.heic (718114 bytes)
[loop] Attachment decrypted: hfork011-proof.heic (718114 bytes)
[remote] Transcoded HEIC/HEIF attachment hfork011-proof.heic to JPEG (558966 bytes)
[remote] Combined 1 image(s) with text message
result: HFORK_HEIC_E2E_OK
~~~

This exercises the production relay, encrypted blob upload/download, CLI decryption, HEIC conversion, Claude image-block construction, and a live GLM-4.6 response - not only the helper in isolation.

## Checks

- rebased onto upstream 5e2dd1ae; range-diff is identical to the original patch
- frozen workspace install passed
- focused HEIC suite: 7/7
- full Happy CLI unit suite: 748/748
- happy-wire build and full Happy CLI build passed
- standalone real fixture conversion: 718,114-byte HEIC to complete 558,966-byte JPEG
- lockfile delta contains only heic-convert, its types, and transitive dependencies

## Overlap

#1387 normalizes HEIC in the current app before upload; this is CLI-side defense in depth for older clients and already-uploaded HEIC blobs. #1387 also moves attachment routing into attachmentContentBlocks.ts. If it lands first, I will rebase this helper into that conversion boundary instead of retaining a competing launcher edit.

Refs bogdanbaciu21/dans-brain#10806 and bogdanbaciu21/dans-brain#15726.

### DIFF
--- a/packages/happy-cli/package.json
+++ b/packages/happy-cli/package.json
@@ -93,6 +93,7 @@
     "expo-server-sdk": "^3.15.0",
     "fastify": "^5.6.2",
     "fastify-type-provider-zod": "^6.1.0",
+    "heic-convert": "2.1.0",
     "http-proxy": "^1.18.1",
     "http-proxy-middleware": "^3.0.5",
     "ink": "^6.5.1",
@@ -110,6 +111,7 @@
   },
   "devDependencies": {
     "@eslint/compat": "^1",
+    "@types/heic-convert": "^2.1.1",
     "@types/inquirer": "^9.0.9",
     "@types/node": ">=20",
     "@types/ws": "^8.18.1",
--- a/packages/happy-cli/src/claude/claudeRemoteLauncher.ts
+++ b/packages/happy-cli/src/claude/claudeRemoteLauncher.ts
@@ -1,6 +1,7 @@
 import { render } from "ink";
 import { Session } from "./session";
 import { MessageBuffer } from "@/ui/ink/messageBuffer";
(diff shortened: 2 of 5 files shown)


bogdanbaciu21:
@bra1ndump Refreshed on the released current main `9a57219f` (Happy 1.2.0) at `5231efd2`; range-diff preserves the contribution byte-for-byte. The new CLI Smoke Test is again blocked before job creation by GitHub’s first-time-contributor approval gate: https://github.com/slopus/happy/actions/runs/29148186539. Could you approve it when convenient? Fresh local proof is green (748/748 CLI unit tests, focused 7/7, frozen install, and CLI build), with the live encrypted relay-to-CLI HEIC proof retained in the PR body. Thanks.

Issue #1540, opened 2026-07-17
chore: plan to remove the deprecated gemini backend


charliezong18:
## Context

In #1430 the gemini backend was marked deprecated in favor of `happy agy` (Antigravity CLI), with the warning "may be removed in a future release". This issue proposes what that removal should actually look like, so it can be discussed before anyone writes a large PR.

## Why remove it

- agy supersedes the gemini backend: same Google models, actively maintained CLI, and the ACP integration the gemini backend relies on has been a recurring source of bugs (#1289, #1093, #1094 are all open fixes against it).
- Keeping two Google backends doubles the maintenance surface for permission handling, transports, and display code.

## Why not remove it yet

- The deprecation shipped about a week ago; users on `happy gemini` have had little time to migrate.
- There are open issues asking for better Gemini CLI support (#896, #469, #971), so there is still real usage.
- Open PRs (#1289, #1093, #1094, #708) touch gemini code and would conflict with a removal.

## Proposed staged plan

1. **Now — stop steering new users to it.** Remove gemini from the app's new-session agent picker. `happy gemini` keeps working from the terminal with the existing deprecation warning, and existing gemini sessions keep rendering. (Small PR, ~1 line: I'll open it alongside this issue.)
2. **Next release cycle — remove the spawn path.** Drop the `happy gemini` subcommand and the gemini entry from CLI detection/spawn (`detectCLI`, `AgentRegistry`, daemon spawn options), after the in-flight gemini PRs above are merged or closed.
3. **Later — remove the backend code.** Delete `packages/happy-cli/src/gemini/`, `GeminiTransport`, `GeminiDisplay`, and related tests. Keep the app-side rendering for the `gemini` session flavor so historical sessions still display correctly, and keep `gemini` in the session-metadata types for backward compatibility.

Non-goals: anything named "gemini" that refers to *models* rather than the CLI backend (e.g. voice models, model pickers) is untouched.

Happy to do the work if maintainers agree with the direction — or adjust the staging if you'd rather keep the backend longer.

PR #1541, opened 2026-07-17
chore(app): remove gemini from the new-session agent picker

charliezong18:
## What

Removes the `gemini (deprecated)` entry from the new-session agent picker. One-line change.

## Why

The gemini backend was deprecated in #1430 in favor of `happy agy`. This is step 1 of the staged removal proposed in #1540: stop steering new sessions to a deprecated backend, without breaking anyone who still uses it.

## What still works

- `happy gemini` from the terminal (with the existing deprecation warning)
- Rendering of existing gemini sessions (flavor handling in `AgentInput`, `ToolView`, session info etc. is untouched)
- A persisted new-session draft with `agentType: 'gemini'` is rewritten to the first available agent by the existing corrective effect in `new/index.tsx` (gemini no longer appears in `availableAgents`), so the spawn path never sees `'gemini'`

## Known inconsistency (intentional)

`settings/agents.tsx` still shows a Gemini defaults section (it iterates `agentKeys`, which drives the stored-defaults schema), and the dev-only session-composer still lists gemini. Both are deliberately left for stage 2/3 of #1540 to keep back-compat with stored defaults.

## Testing

- `tsc --noEmit` in `packages/happy-app`: clean
- `vitest`: 706/706 passing
- Built the app with Metro (`expo start --web`, full non-lazy bundle) and verified the served bundle contains claude/codex/openclaw/agy picker labels and zero occurrences of `gemini (deprecated)`

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -74,7 +74,6 @@ const ALL_AGENTS: { key: AgentKey; label: string }[] = [
     { key: 'claude', label: 'claude code' },
     { key: 'codex', label: 'codex' },
     { key: 'openclaw', label: 'openclaw' },
-    { key: 'gemini', label: 'gemini (deprecated)' },
     { key: 'agy', label: 'agy' },
 ];
 


bra1nDump:
thank you for your work! :D

Issue #1251, opened 2026-05-11
[Bug]: chat title not created anymore


ilteoood:
Since the last update, chat's titles are not updated anymore once interacting with Claude Code.
I can see only the generic "New chat" title for every interaction.
<img width="515" height="671" alt="Image" src="https://github.com/user-attachments/assets/3926ada1-3294-46f3-8a93-a25c21348a64" />

Puppo:
Same here! On iOS.

grxphxnxx:
Same here on iOS

ashleywr:
I dug into the CLI/app code to understand why titles stop appearing (and specifically why it correlates with the model in use). Sharing in case it helps triage.

**Titles are entirely agent-driven — there is no server-side or automatic title generation.** The whole pipeline depends on the coding agent voluntarily calling the `mcp__happy__change_title` MCP tool:

1. The CLI appends an instruction to the system prompt telling the model to set a title:
   > "ALWAYS when you start a new chat - you must call a tool `mcp__happy__change_title` to set a chat title..."
   (`packages/happy-cli/src/claude/utils/systemPrompt.ts:7-9`)
2. When (and only when) the model calls that tool, the handler emits a `type: 'summary'` message:
   (`packages/happy-cli/src/claude/utils/startHappyServer.ts:63-70`)
3. The CLI stores that as session metadata `summary`, and the app displays it — falling back to "New chat" when it is absent:
   (`packages/happy-app/sources/utils/sessionUtils.ts:80-85`)

```ts
export function getSessionName(session: Session): string {
    if (session.metadata?.summary) {
        return session.metadata.summary.text;
    }
    return t('session.newChat');
}
```

**Why it correlates with weaker/faster models (e.g. Haiku):** setting a title requires the model to proactively obey a meta-instruction and fire an MCP tool call at the very start of the conversation, alongside doing the user's actual task. Smaller models follow that "also call this tool" instruction much less reliably and often just skip it. There is **no fallback anywhere** — no retry, no forced first-turn call, no separate title model — so if the model never calls `change_title`, the title stays "New chat" permanently. Opus/Sonnet sessions title consistently; Haiku sessions frequently do not.

So this looks like a **design gap rather than a regression in the title code itself**: the title pipeline is intact, but it has a single point of failure (model cooperation) with no backstop. If the default/available model shifted toward Haiku in a recent build, that would explain the "since the last update" timing without any change to the title code.

**Possible fixes (CLI-side, model-independent):**
[...]

PR #1544, opened 2026-07-19
fix(app): reflect spawn-time model/permission/effort pick in session config

chphch:
When a new session is spawned with a non-default model (e.g. picking a model other than the agent default on the New Session screen), the per-session agent-mode override written right after `refreshSessions()` in `handleSend` can be silently dropped: if the freshly-spawned session has not yet landed in local storage at that instant, `updateSessionAgentModes()` returns early (`if (!session) return`) and the parallel metadata push throws into a swallowed `.catch`. For a Claude session the `modelMode` mirror is the only carrier of the pick (`currentModelCode` is written only on the ACP path), so the in-session model config then falls back to the agent default — e.g. a session actually running `fable` shows `opus` in its config until the user manually re-picks in the session. The neighboring `update-session` handler already guards this "session not ready yet" window (#1251), but the new-session override write did not.

Fix: bounded-wait for the spawned session to appear in local storage (reusing the public `refreshSessions()`) before writing the picks. This applies to all three overrides — permission / model / effort — which shared the same silent-drop.

## Proof

Verified locally end-to-end via a standalone `happy-server` (PGlite) + Expo web with a paired `happy-cli` daemon, driven headlessly with Playwright: spawn a new session picking the `fable` model, then open the in-session model config — it correctly shows **fable 5** selected (before the fix it fell back to the default). Screenshot in the follow-up comment.

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -1493,6 +1493,20 @@ function NewSessionScreen() {
                     // later settings/default change must not silently rewrite
                     // an existing session's permission, model, or effort.
                     if (!spawnRigCreation) {
+                        // The just-spawned session may not be in local storage the
+                        // instant refreshSessions() resolves (server read-after-write
+                        // + broadcast timing). sessionSetAgentModes would then silently
+                        // no-op (updateSessionAgentModes returns early when the session
+                        // is missing), permanently dropping the picked model/permission/
+                        // effort so the in-session config falls back to the agent
+                        // default (e.g. shows 'opus' for a session spawned as 'fable').
+                        // Wait (bounded) for the session to land, then write the picks.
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — spawn-time model pick reflected in session config

Verified locally end-to-end: standalone `happy-server` (PGlite) + Expo web with a paired `happy-cli` daemon, driven headlessly with Playwright. A new session is spawned picking the **fable** model on the New Session screen; opening the in-session model config then shows **fable 5** selected — not the agent default.

spawn-time model reflected in config

Frame 1: New Session screen with `fable 5` picked. Frame 2: the spawned session's model config sheet — `fable 5` is the selected radio. Before this change, `updateSessionAgentModes()` no-op'd when the just-spawned session hadn't landed in storage yet, so the config fell back to the default (`opus`) until a manual re-pick.

PR #1568, opened 2026-07-25
fix(app): keep new-session mode selections per agent

charliezong18:
## Context

Switching agents in HomeDock silently resets the user's model/permission/effort choices. `selectAgent()` overwrites the shared draft fields with the new agent's defaults, so e.g. picking a non-default Claude model, toggling to Codex to run something, and coming back leaves the Claude picker back on its default — every time. The new-session screen (`new/index.tsx`) doesn't clobber, so the two entry points also disagree about what an agent switch does.

## What changed

- The new-session draft now stores mode selections per agent (`agentModes` map in `persistence.ts`), instead of one shared `permissionMode`/`modelMode`/`effortLevel` triple.
- `useNewSessionDraft` still exposes the flat fields, but they are a view of the currently selected agent; `setAgentType` swaps the view. Consumers (`new/index.tsx`, `useStartSessionFromDraft`) needed no changes.
- `HomeDock.selectAgent` no longer writes the target agent's defaults into the draft — saved selections restore, and agents without saved selections fall back to their defaults as before (via the existing `resolveOption`/`resolveAgentDefaultConfig` chain).
- Legacy drafts (single shared triple) migrate on load: the stored values are attributed to the draft's agent. Same storage key, no data loss.

## Tests

- `useNewSessionDraft.test.ts`: updated for the new shape; new case covering claude → codex → claude round-trip keeping each agent's model choice.
- `newSessionDraft.spec.ts` (new): draft parsing — per-agent shape, legacy migration, empty/garbage payloads.
- Full happy-app suite 784/784 green, `tsc --noEmit` clean.

(Screenshots of the agent-switch round-trip coming shortly.)

### DIFF
--- a/packages/happy-app/sources/components/HomeDock.tsx
+++ b/packages/happy-app/sources/components/HomeDock.tsx
@@ -760,12 +760,11 @@ export const HomeDock = React.memo(({
     }, [finishCloseFocusMode, focusPresentation]);
 
     const selectAgent = React.useCallback((agent: NewSessionAgentType) => {
-        const nextDefaults = resolveAgentDefaultConfig(defaultOverrides, agent);
+        // Mode selections live per agent in the draft, so switching agents
+        // restores the target agent's saved modes (or falls back to its
+        // defaults) instead of overwriting them.
         setAgentType(agent);
-        setPermissionMode(nextDefaults.permissionMode);
-        setModelMode(nextDefaults.modelMode);
-        if (nextDefaults.effortLevel) setEffortLevel(nextDefaults.effortLevel);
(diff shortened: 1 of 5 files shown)


charliezong18:
Sequencing note: this and #1569 should land together, in order. Alone, this PR would regress settings-default application — the old `selectAgent` incidentally re-stamped configured defaults into the draft on every agent toggle, and with per-agent memory alone a remembered pick would shadow a newly configured default with no clearing path. #1569 adds exactly that clearing path; I'll rebase it as soon as this lands.

charliezong18:
Closing — @bra1nDump is taking the new-session defaults behaviour in his own direction (defaults configured in settings become sticky), which addresses the same underlying complaint this PR was aimed at. No sense leaving a competing implementation in the queue.

PR #1569, opened 2026-07-25
fix(app): clear stale draft selections when agent defaults change

charliezong18:
## Context

Configuring an agent default in Settings → Agents can silently do nothing. The new-session draft remembers the user's last manual mode selection and always wins over configured defaults in the resolution chain (draft → override → code default), so any older remembered pick permanently shadows a newly configured default. Concretely: set Opus 5 as the Claude default, and new sessions keep spawning with a remembered `opus` pick — with no indication why.

## What changed

- `useNewSessionDraft`: new `clearAgentMode(agent, field)` (drops one remembered selection; removes the agent's entry entirely when all fields are empty; keeps the current-agent view in sync) and `clearAllAgentModes()`.
- Settings → Agents: changing any default (including selecting "Use code default") clears the matching remembered draft selection for that agent; "Clear Overrides" clears all remembered selections.
- Net effect: the moment you configure a default, it actually takes effect for the next session.

## Notes for review

**Stacked on #1568** (per-agent draft modes) — please review only the last commit (`0302bf29`); the first commit is #1568's. I'll rebase onto main once #1568 lands.

## Tests

3 new cases in `useNewSessionDraft.test.ts`: clearing one field leaves the agent's other fields intact; clearing another agent's selection doesn't touch the current agent's view; reset clears everything. Full happy-app suite 787/787 green, `tsc --noEmit` clean.

### DIFF
--- a/packages/happy-app/sources/app/(app)/settings/agents.tsx
+++ b/packages/happy-app/sources/app/(app)/settings/agents.tsx
@@ -11,6 +11,7 @@ import {
     type ModeOption,
 } from '@/components/modelModeOptions';
 import { useSettingMutable } from '@/sync/storage';
+import { useNewSessionDraft } from '@/hooks/useNewSessionDraft';
 import {
     agentKeys,
     getCodeAgentDefaults,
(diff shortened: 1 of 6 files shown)


charliezong18:
Known limitation from self-review: `agentDefaultOverrides` is synced across devices, but the new-session draft is per-device. Changing a default on device A clears the remembered pick on device A only — device B's older pick still shadows the newly synced default until it's touched there. A follow-up could clear matching picks where the settings-sync apply path ingests override changes; happy to file that separately if you want it.

charliezong18:
Closing — @bra1nDump is taking the new-session defaults behaviour in his own direction (defaults configured in settings become sticky), which addresses the same underlying complaint this PR was aimed at. No sense leaving a competing implementation in the queue.

PR #1570, opened 2026-07-25
feat(cli,app): support image attachments in agy sessions

### LINKED ISSUE #1559: bug(provider): agy provider drops multimodal/image payloads from UI drag-and-drop
# Draft Issue: agy Provider Multimodal Support

**Title:** `bug(provider): agy provider drops multimodal/image payloads from UI drag-and-drop`

## Context
Happy's UI supports dragging and dropping images directly into the chat canvas, which works seamlessly with the `claude` provider. However, when using the newly integrated `agy` (Antigravity/Gemini) provider, images dropped into the UI are silently discarded before the payload reaches the `agy` backend. Only the text portion of the prompt is transmitted.

## Expected Behavior
Gemini (via `agy`) is natively multimodal. When an image is attached in the Happy UI, the `agy` provider adapter should properly serialize the image (e.g., base64 encoding or file path extraction, depending on what the `agy` protocol expects) and include it in the payload sent to the backend.

## Current Behavior
The `agy` provider implementation currently appears to only parse and forward text, stripping out any multimodal attachments attached from the Happy Desktop UI. As a result, the agent replies that it cannot see the image.

## Proposed Scope
1. Inspect the payload conversion logic in `packages/happy-cli/src/agent/factories/agy.ts` (or the respective request formatter).
2. Map Happy's internal image attachment structure into the multimodal structure expected by the `agy` API/CLI protocol.
3. Verify that `agy` correctly receives and processes the image payload alongside the text.

*(Would be happy to submit a PR for this if the protocol shape for `agy` image attachments is known/documented!)*

charliezong18:
Fixes #1559

## Problem

The agy backend silently drops image attachments: the app never shows the
attach button for agy sessions, and the CLI ignores any attachment payload
on incoming user messages — even though agy itself can read local image
files just fine.

## How it works

agy has no image-input flag, so the only viable path is the filesystem:

- **app**: allow the `agy` flavor in `supportsImageAttachmentsForFlavor`,
  so the existing image picker (behind the `expImageUpload` feature flag)
  appears in agy sessions.
- **cli**: decode wire attachments into a per-session cache dir under
  `~/.happy/agy-image-cache/<sha256(sessionId)>`, expose that dir for the
  turn via `--add-dir`, and prefix the prompt with the exact file paths.
  The dir is removed at session end. Decoding and cache layout are reused
  from the Codex image-input helpers (same magic-byte sniffing, same 0700
  permissions), only the cache root differs.

Session dirs are keyed by a hash of the raw session id rather than the
sanitized id: the sanitizer is not injective (`a/b` and `a_b` collide),
and since the dir is both handed to `--add-dir` and `rm -rf`'d at session
end, two sessions must never share one — even for hostile server-issued
ids.

Isolation is structural, not prompt-based: given a nonexistent path agy
has been observed to scan the surrounding directory and describe
unrelated files, so each session dir contains only that session's
attachments. The "read ONLY these exact paths" instruction in the prompt
is belt on top of those suspenders.

Also fixes a latent crash: `runAgy`/`runCodex` register attachment
handlers unconditionally at startup, but the offline session stub didn't
implement them, so offline startup died with a TypeError before the
reconnection loop began. The stub now implements the attachment surface,
with a regression test.

## Before

Dragging an image into an agy session: the attach button never appears
(flavor not whitelisted); if a message with attachments reaches the CLI,
they are silently discarded.

## After

Google Chrome 2026-07-24 23 03 45

## Testing

- `happy-cli`: 792/792 unit tests (84 files), including new suites for
  `imageInput` (11 tests, incl. the stale-dir sweep), `AgyBackend` add-dir passthrough, and the
  offline stub regression; `tsc --noEmit` clean.
- `happy-app`: 779/779 tests (65 files) incl. the updated
  `attachmentSupport` whitelist test; `tsc --noEmit` clean.
- e2e: verified locally end to end — image attached in the web app, agy
  (`--sandbox`) reads the decoded file and describes it.

### DIFF
--- a/packages/happy-app/sources/sync/attachmentSupport.test.ts
+++ b/packages/happy-app/sources/sync/attachmentSupport.test.ts
@@ -7,11 +7,12 @@ import {
 } from './attachmentSupport';
 
 describe('supportsImageAttachmentsForFlavor', () => {
-    it('supports legacy sessions, Claude, and Codex', () => {
+    it('supports legacy sessions, Claude, Codex, and agy', () => {
         expect(supportsImageAttachmentsForFlavor(undefined)).toBe(true);
         expect(supportsImageAttachmentsForFlavor(null)).toBe(true);
         expect(supportsImageAttachmentsForFlavor('claude')).toBe(true);
         expect(supportsImageAttachmentsForFlavor('codex')).toBe(true);
+        expect(supportsImageAttachmentsForFlavor('agy')).toBe(true);
     });
 
     it('rejects Gemini, OpenClaw, and unknown explicit flavors', () => {
--- a/packages/happy-app/sources/sync/attachmentSupport.ts
+++ b/packages/happy-app/sources/sync/attachmentSupport.ts
@@ -8,7 +8,7 @@ export type ImageAttachmentSendPlan = {
 };
 
 export function supportsImageAttachmentsForFlavor(flavor: ImageAttachmentFlavor): boolean {
-    return !flavor || flavor === 'claude' || flavor === 'codex';
+    return !flavor || flavor === 'claude' || flavor === 'codex' || flavor === 'agy';
 }
 
 export function getImageAttachmentSendPlan(opts: {
--- a/packages/happy-cli/src/agy/AgyBackend.test.ts
+++ b/packages/happy-cli/src/agy/AgyBackend.test.ts
@@ -61,6 +61,27 @@ describe('AgyBackend', () => {
     expect(messages.at(-1)).toMatchObject({ type: 'status', status: 'idle' });
   });
 
+  it('passes per-turn extra add-dirs through to the agy argv', async () => {
+    const { child } = makeFakeChild();
+    const spawnFn = vi.fn(() => child) as unknown as SpawnFn;
+
+    const backend = new AgyBackend({
+      cwd: '/work',
+      permissionMode: 'default',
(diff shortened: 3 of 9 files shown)


charliezong18:
Pushed a follow-up from self-review tightening the cleanup story:

1. **Cleanup now survives the daemon's stop path.** The daemon stops sessions with a bare SIGTERM, which skips the `finally` cleanup — every daemon-stopped session with images leaked its cache dir permanently (dirs are keyed by hashed session id; nothing revisits them). Cleanup now also runs on SIGTERM/SIGINT, and a startup sweep removes leftover dirs older than 7 days — age-gated so a concurrent live session's dir is never touched.
2. **The `--add-dir` grant is latched for the session** once any image lands: agy is spawned fresh per `--print` turn, so a later text-only turn ("look at that screenshot again") previously lost sandbox access to earlier images. The files live until session end regardless, so this widens nothing.
3. The offline session stub gained the missing `sendSessionProtocolMessage` no-op (reached via `sendEnvelopes` on every turn; previously hidden by the cast).

Known limitation at the intersection with #1574: a conversation resumed after a restart may reference image paths deleted at the previous session's end. Structurally safe — the dir is absent and no longer granted — but degraded; the real fix is conversation-lifetime caching, which belongs with the resume work as a follow-up.

charliezong18:

PR #1571, opened 2026-07-25
feat(cli,app): surface Antigravity (agy) quota in the session status bar

charliezong18:
Follow-up to #1430 (agy backend) and #1556 (usage limits in the status bar).

## Problem

agy sessions show no usage information: agy reports no quota over its CLI
stream, so the status bar chip and usage popover added in #1556 stay empty
for agy sessions.

## How it works

- **cli — quota client** (`agyQuotaClient.ts`): talks to the same private
  endpoint the agy CLI itself uses
  (`cloudcode-pa.googleapis.com/v1internal:retrieveUserQuota`),
  authenticated with a short-lived access token minted from agy's stored
  refresh token. The OAuth client id/secret aren't published; they are
  recovered by scanning the agy binary (self-healing across agy upgrades
  that rotate the client) and cached in a dedicated file. All external
  effects (fs, `strings`, fetch) are injectable, so the token/quota logic
  is unit-testable without a real binary or network.
- **cli — adapter** (`agyUsageAdapter.ts`): agy quota is per-model — each
  bucket is one model's remaining fraction with its own reset time; there
  is no 5h/7d concept. It maps onto the backend-neutral `UsageLimits`
  windows from #1556: one synthetic `agy` headline window (the binding
  model, i.e. lowest remaining) so the status bar renders a single chip,
  plus one `agy:<modelId>` window per model for the detail popover.
- **cli — collector** (`agyUsageCollector.ts`): turn-driven, not polled —
  one pull when the session opens and one after each turn (including
  failed turns: a quota rejection is exactly when the chip matters), with
  a 60s minimum interval so a burst of short turns doesn't become a burst
  of token exchanges. Pulls are serialized and never awaited by the turn;
  a failure just hides the chip instead of breaking anything.
- **cli — wiring** (`runAgy.ts`): patches merge into
  `agentState.usageLimits` exactly like the Claude path.
- **app**: one label entry (`agy`) in the chip whitelist; per-model
  `agy:<modelId>` windows stay popover-only like other unknown window ids.

## Before

agy sessions: no usage chip, empty usage popover.

## After

agy quota chip in a live agy session

The `agy 100%` chip in the status bar (bottom right) of a live agy
session — same session as the screenshot in #1570.

## Testing

- 19 new unit tests across the quota client (token minting, client
  recovery from the binary, caching, error paths), the adapter (bucket
  mapping, binding-model selection, reset times), and the collector
  (turn-driven cadence, min interval, serialization).
- `happy-cli`: 800/800 tests (85 files), `tsc --noEmit` clean.
- `happy-app`: 780/780 tests (65 files) incl. the new chip-label
  coverage, `tsc --noEmit` clean.
- Verified live: chip renders in a real agy session (screenshot above).

### DIFF
--- a/packages/happy-app/sources/utils/sessionStatusBar.test.ts
+++ b/packages/happy-app/sources/utils/sessionStatusBar.test.ts
@@ -70,6 +70,22 @@ describe('usage limit helpers', () => {
         expect(chips[0].id).toBe('seven_day');
     });
 
+    it('chips the agy headline window and leaves per-model windows to the popover', () => {
+        const agyLimits = {
+            capturedAt: 1,
+            windows: [
+                { id: 'agy', label: 'agy', utilization: 50, resetsAt: 100 },
+                { id: 'agy:gemini-3.1-pro-high', label: 'Gemini 3.1 Pro', utilization: 50, resetsAt: 100 },
+                { id: 'agy:claude-sonnet-4-6', label: 'Claude Sonnet 4.6', utilization: 0, resetsAt: 200 },
+            ],
+        };
+        const chips = getUsageLimitChips(agyLimits, false);
+        expect(chips.map(c => c.id)).toEqual(['agy']);
+        expect(chips[0].shortLabel).toBe('agy');
+        // The per-model detail still reaches the popover.
+        expect(getUsageLimitRows(agyLimits).map(r => r.label)).toEqual(['agy', 'Gemini 3.1 Pro', 'Claude Sonnet 4.6']);
+    });
+
     it('hides chips for windows without numeric utilization and for absent data', () => {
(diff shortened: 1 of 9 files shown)


charliezong18:
Pushed a hardening commit from self-review:

- the creds cache is written `0600` — it holds the client secret, and agy's own token file next to it is `0600`
- a failed cache write no longer discards a successfully minted token (it previously fell through to the remaining candidate pairs and ended in a misleading "no working OAuth client found")
- both fetches carry a 10s timeout so a black-holed connection can't hold the process open past session end
- the bare-name binary lookup uses `execFile` with the platform's own resolver (`which`/`where`) instead of a Unix-only interpolated shell string, so Windows degrades gracefully

charliezong18:

Issue #1558, opened 2026-07-23
Support session resume for agy mode


charliezong18:
Currently, `happy-app` throws a `resumeSessionMissingBackendId` error when trying to resume an `agy` session because it only checks for `claudeSessionId` and `codexThreadId`. Furthermore, `happy-cli` does not accept a `--resume` flag for `agy` mode.\n\nTo fix:\n1. Update `MetadataSchema` in `storageTypes.ts` to include `agyConversationId` (or similar).\n2. Update `AgyBackend.ts` to populate this ID.\n3. Update `useSessionQuickActions.ts` to recognize it and allow resuming.\n4. Update `runAgy.ts` to accept and handle a `--resume` flag, linking it to the Antigravity Conversation ID.

charliezong18:
Closing this — the implementation (#1574) was withdrawn on July 28 after talking it through with @bra1nDump: the agy backend is a niche surface, and this line was largely scratching my own itch. Leaving the issue open would imply someone is still driving it upstream, which isn't the case.

The work itself is done and running on my fork, so if agy resume ever comes back, the cross-restart resume-id handling is there to lift.

PR #1574, opened 2026-07-25
feat(cli): resume agy conversations across session restarts

charliezong18:
Part of #1558.

## Problem

The agy backend pins its conversation id only in process memory: `agy --print`
doesn't echo the id it creates, so the backend recovers it from agy's per-cwd
cache after the first turn and passes it per turn — but nothing persists it.
Restart happy-cli (or resume via the daemon) and the session starts a
brand-new agy conversation.

## How it works

Same pattern as `claudeSessionId` / `codexThreadId`:

- **Persist:** `AgyBackend` gains an `onConversationId` callback that fires
  once when a fresh conversation is pinned; `runAgy` writes it to session
  metadata as `agyConversationId`. On a resumed run the seeded id is persisted
  immediately — a resumed run is a fresh Happy session and the callback
  doesn't fire for seeded ids, so without this the id would be dropped after
  one hop and the resumed session would itself be unresumable.
- **Restore:** `buildResumeLaunch` recognizes the agy flavor and emits
  `happy agy --resume <id>`. The seeded id comes from the session's own
  persisted metadata, so — unlike the cwd cache — adopting it cannot
  cross-resume another session's conversation (that guard is untouched).
- Plumbing mirrors the siblings: `Metadata`, `ResumableMetadataSchema`,
  `needsFreshMetadata`, and the daemon's stale-metadata check.

## Scope

CLI-side only. Two follow-ups deliberately not in this PR:
- app-side wiring so the UI offers resume for agy sessions
  (`MetadataSchema`, `hasBackendResumeId`, `buildResumeInvocation`);
- reconnect-in-place: `runAgy` doesn't read `HAPPY_RECONNECT_SESSION_ID` yet,
  so a daemon-driven resume continues the old agy conversation in a new Happy
  session. Previously daemon resume hard-errored ("unsupported flavor"), so
  this is a strict improvement.

## Testing

`happy-cli`: 785/785 tests (82 files) incl. new coverage for seeded-vs-fresh
pinning, agy resume launch construction, and the cross-resume guard;
`tsc --noEmit` clean. `--resume` validates its value (rejects missing or
flag-like values).

### DIFF
--- a/packages/happy-cli/src/agy/AgyBackend.test.ts
+++ b/packages/happy-cli/src/agy/AgyBackend.test.ts
@@ -202,6 +202,50 @@ describe('AgyBackend', () => {
     expect(pinned).toEqual([]);
   });
 
+  it('adopts the replacement id when a stale seeded conversation was not resumed (self-heal)', async () => {
+    const spawnCalls: string[][] = [];
+    let current = makeFakeChild();
+    const spawnFn = vi.fn((_bin: string, args: string[]) => {
+      spawnCalls.push(args);
+      return current.child;
+    }) as unknown as SpawnFn;
+
+    // The persisted seed points at a conversation agy no longer knows:
+    // `--conversation <dead-id>` exits 0, silently creates a fresh
+    // conversation, and writes the fresh id to the cwd cache after the turn.
+    // Without adoption the dead id would be re-persisted forever and every
+    // resumed turn would start amnesiac.
+    const pinned: string[] = [];
+    let recorded: string | null = null;
+    const backend = new AgyBackend({
+      cwd: '/work',
+      permissionMode: 'default',
+      spawnFn,
+      resolveConversationId: () => recorded,
+      initialConversationId: 'dead-cid',
+      onConversationId: (id) => pinned.push(id),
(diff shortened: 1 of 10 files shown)


charliezong18:
Pushed a follow-up from self-review that closes a silent failure mode: agy never validates a pinned id — `--conversation <dead-uuid>` exits 0 and silently starts a fresh conversation (verified against a live agy), so a stale persisted id (state cleanup, reinstall, `~/.gemini` wipe) meant every resumed turn succeeded amnesiac while re-persisting the dead id forever.

agy writes the id it actually used back to the per-cwd cache after each turn, so the backend now adopts a cache id that changed during our turn to something other than the pin, firing `onConversationId` so the persisted metadata self-heals. The changed-during-turn guard keeps the existing cross-resume protection intact: an unchanged foreign cache entry is still never adopted. Covered by a new test.

Also fixed: the daemon's resume path appends `--model`/`--permission-mode` for every flavor, and the agy arg loop silently dropped both — they're now parsed and applied as the initial model/permission mode.

charliezong18:

PR #1576, opened 2026-07-25
feat(cli,app): per-model effort levels and current model slugs for agy

charliezong18:
## Problem

The agy integration hardcodes the pre-1.1.5 model list — display names with the effort baked into the name ("Gemini 3.1 Pro (High)") — so the model picker is missing Gemini 3.6 Flash entirely and agy sessions have no Effort control, even though the CLI has grown one.

agy 1.1.5 added stable model slugs accepted by `--model` and an `--effort` flag that selects a model's reasoning variant.

## How it works

- **app**: `getAgyModelModes()` now lists base slugs (incl. `gemini-3.6-flash`); `getEffortLevelsForModel` gains an agy branch with per-model level sets (3.1 pro: low/high — it has no medium; 3.5/3.6 flash: low/medium/high; fixed-variant models: none). Every picker and the agent-defaults screen already resolve effort per (flavor, model), so the Effort control lights up everywhere from this one function.
- **cli**: `buildAgyArgs` emits `--effort`, gated on a variant map — agy *requires* `--effort` for base slugs with variants (`--model gemini-3.1-pro` alone errors) and *rejects* it for everything else, so the flag is emitted only for known variant models, and a level the model doesn't offer is replaced by the default (high) instead of erroring the turn. `runAgy` picks up `meta.effort` per turn, following the `runClaude` pattern.
- **machine page**: adds an Agy CLI-availability row, shown only when the daemon reports the (optional) field — older CLIs don't.

Backward compatibility: agy still accepts the old display names, so stored drafts/overrides/session selections keep working with no migration; they just get no Effort control (their effort is baked into the name).

Deliberately NOT wired: agy's `--mode accept-edits|plan`. In `--print` mode the flag is accepted but not enforced — plan mode still writes files (verified on agy 1.1.7) — so offering those permission modes would be a false promise. A comment in `getAgyPermissionModes` records this.

## Test plan

- happy-cli: agy suite 22/22 (4 new cases for the `--effort` gating, 1 for backend passthrough)
- happy-app: tsc + full suite 780/780 (new per-model effort assertions)

## Screenshots

The AGY model picker with the current base slugs:

AGY model picker expanded

The per-model Effort control (gemini 3.1 pro exposes low/high only):

AGY effort control expanded

### DIFF
--- a/packages/happy-app/sources/app/(app)/machine/[id].tsx
+++ b/packages/happy-app/sources/app/(app)/machine/[id].tsx
@@ -562,6 +562,19 @@ export default function MachineDetailScreen() {
                                 </Text>
                             }
                         />
+                        {/* agy is optional in the metadata schema (older CLIs don't
+                            report it) — undefined means unknown, not missing. */}
+                        {metadata.cliAvailability.agy !== undefined && (
+                            <Item
+                                title="Agy"
+                                showChevron={false}
+                                rightElement={
+                                    <Text style={{ color: metadata.cliAvailability.agy ? '#34C759' : theme.colors.textSecondary, fontSize: 14 }}>
+                                        {metadata.cliAvailability.agy ? t('machine.cliInstalled') : t('machine.cliNotFound')}
+                                    </Text>
+                                }
+                            />
+                        )}
                         <Item
                             title={t('machine.lastDetected')}
                             subtitle={new Date(metadata.cliAvailability.detectedAt).toLocaleString()}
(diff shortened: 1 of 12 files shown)


charliezong18:
Pushed a follow-up fixing a gap self-review caught: `MessageMetaSchema` on the CLI side had no `effort` field, and zod strips unknown keys at the `UserMessageSchema` boundary — so the effort the app attached to a user message never reached the backend, and agy always ran with the default `--effort` regardless of the pick. The unit tests missed it by calling `setEffort` below the schema boundary; there's now a regression test at the boundary itself.

Side effect worth flagging: codex reads `meta.effort` the same way, so this also enables mid-session effort changes for codex, which were previously silently dropped (codex only received effort at spawn time via argv).

charliezong18:

PR #1577, opened 2026-07-25
feat(app): optional purple unread indicator + fix unread masking running state

charliezong18:
## What

Two related changes to the session list unread indicator:

1. **Fix: unread no longer masks a running session.** A session marked unread that later resumed working kept showing the solid unread dot until opened — the unread override took precedence over the live state. The override now only applies while the session is settled; `thinking` (pulsing blue) and `permission_required` (pulsing orange) take precedence. Unread is not lost — the solid dot returns once the session settles again.

2. **Feature: optional purple unread indicator.** New synced setting `unreadIndicatorPurple` (default **off**, no visual change unless opted in). The solid blue unread dot and the pulsing blue thinking dot differ only by animation, which is hard to tell apart at a glance when scanning many concurrent sessions. With the toggle on, unread renders purple (#AF52DE): pulsing blue = still running, purple = finished with results to check. The toggle sits in Appearance below "Show AI Provider Icons"; translations added for all supported languages.

## Testing

- `tsc --noEmit` passes
- `settings.spec.ts` passes (34/34), new default covered by the defaults assertion
- Dogfooded on my fork with the toggle on: unread shows purple, and a session that resumes thinking correctly pulses blue instead of staying solid

Google Chrome 2026-07-25 00 24 08
Google Chrome 2026-07-25 00 24 23

### DIFF
--- a/packages/happy-app/sources/app/(app)/settings/appearance.tsx
+++ b/packages/happy-app/sources/app/(app)/settings/appearance.tsx
@@ -214,6 +214,7 @@ export default function AppearanceSettingsScreen() {
     const [alwaysShowContextSize, setAlwaysShowContextSize] = useSettingMutable('alwaysShowContextSize');
     const [avatarStyle, setAvatarStyle] = useSettingMutable('avatarStyle');
     const [showFlavorIcons, setShowFlavorIcons] = useSettingMutable('showFlavorIcons');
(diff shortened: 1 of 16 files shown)


charliezong18:
Closing — the purple indicator isn't a UI change the project wants, which is fair. The unread-masking fix this PR bundled alongside it is also largely subsumed by #1579 (now merged), which fixed the root cause rather than the symptom.

PR #1578, opened 2026-07-25
fix(cli): surface the real launch error instead of a bare notice

charliezong18:
## What

Both Claude launchers swallow the thrown Error on a failed launch and report a fixed string:

```ts
} catch (e) {
    logger.debug('[remote]: launch error', e);
    if (!exitReason) {
        session.client.closeClaudeSessionTurn('failed');
        session.client.sendSessionEvent({ type: 'message', message: 'Process exited unexpectedly' });
```

The Error message is the only clue about what actually went wrong, and it never reaches the app.

## Why

Hit this in practice. Switching the model mid-session triggers a relaunch; the relaunch failed and the app showed only `Process exited unexpectedly`. Switching back to the previous model failed identically, so the session was effectively dead with no indication why.

The real error was sitting in the CLI log:

```
[remote]: launch error Error: Native CLI binary for darwin-arm64 not found.
Reinstall @anthropic-ai/claude-agent-sdk without --omit=optional, or set
options.pathToClaudeCodeExecutable.
```

The root cause turned out to be unrelated to the model switch: a `pnpm install` in the repo had rebuilt `node_modules/@anthropic-ai/`, invalidating the resolve path the already-running process had cached. The SDK's own lookup swallows every failure in a `catch {}` and reports it as "binary not found", so even that message is imprecise — but it still names the package and a fix, which is far more than the app showed.

Recovering that one line took a long dig through the logs. A user without CLI log access would have had nothing to go on.

## What changed

`launchFailureMessage(error)` appends the Error message to the existing notice:

- non-Error throwables and empty messages fall back to the unchanged string
- whitespace is collapsed so the client's single-line status rendering stays intact
- the detail is truncated at 300 characters

Both launchers call it. Behaviour is unchanged when there is no usable message. Result:

```
Process exited unexpectedly: Native CLI binary for darwin-arm64 not found. Reinstall @anthropic-ai/claude-agent-sdk without --omit=optional, or set options.pathToClaudeCodeExecutable.
```

## Testing

- 7 unit tests on the helper: fallbacks for non-Errors and empty messages, whitespace collapsing, and the truncation boundary — including the exact SDK message above as a regression case
- 1 launcher test exercising the catch path end to end, asserting the detail reaches `sendSessionEvent`. Confirmed it fails when the launcher is reverted to the hardcoded string, and that the failure diff is exactly the dropped detail
- `pnpm vitest run src/claude` — 223 passed, 6 skipped (skips are integration tests requiring live API access)
- `tsc --noEmit` clean

## Not included

- `claudeRemoteLauncher` has no existing test harness (it renders through ink), so the new launcher test covers the local launcher only. The two catch paths are structurally identical and both call the same helper.
- `runCodex.ts` has the same hardcoded string in two places and could use the same helper. Left out to keep this focused — happy to fold it in if preferred.

### DIFF
--- a/packages/happy-cli/src/claude/claudeLocalLauncher.test.ts
+++ b/packages/happy-cli/src/claude/claudeLocalLauncher.test.ts
@@ -206,4 +206,54 @@ describe('claudeLocalLauncher', () => {
         localRun.resolve();
         await launcher;
     });
+
+    it('reports the underlying error when a launch throws, instead of a bare notice', async () => {
+        // The SDK surfaces an unresolvable native binary as a plain Error; the
+        // launcher used to drop it and report only "Process exited unexpectedly".
+        const sdkFailure = new Error(
+            'Native CLI binary for darwin-arm64 not found. Reinstall @anthropic-ai/claude-agent-sdk without --omit=optional, or set options.pathToClaudeCodeExecutable.'
+        );
+        mockClaudeLocal
+            .mockRejectedValueOnce(sdkFailure)
+            .mockResolvedValueOnce(undefined);
+
+        const session = {
+            sessionId: 'claude-session-3',
+            path: '/tmp/project',
+            client: {
+                sendClaudeSessionMessage: vi.fn(),
+                sendClaudeSessionMessageFromLocalTranscript: vi.fn(async () => {}),
+                closeClaudeSessionTurn: vi.fn(),
+                sendSessionEvent: vi.fn(),
+                rpcHandlerManager: {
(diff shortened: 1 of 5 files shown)


charliezong18:
Pushed a small hardening commit from self-review: the detail string is now scrubbed of ANSI escape sequences and stray C0 controls before whitespace collapsing — child-process errors can carry colored output, and the string renders verbatim in the app. Tests cover a colored multi-line message and an ANSI-only message (falls back to the bare notice).

Issue #1492, opened 2026-07-06
Per-session model/effort selection doesn't sync across devices (Android vs web)


ashleywr:
### Summary

The per-session **model** and **effort** selections don't sync across devices. Pick Sonnet (or a different effort) on the Android app, open the same session in the Chrome web app, and it shows a different model/effort — same account, same session. This is confusing UX, and the git history suggests it's an unintended side effect of copy-paste rather than a deliberate design choice.

### Repro

1. Open a session on device A (e.g. Android), change the model or effort in the session UI.
2. Open the same session on device B (e.g. Chrome web) for the same account.
3. Device B shows a different model/effort than device A.

Effort diverges even more visibly than model, because effort has no synced fallback layer (see below).

### Root cause

The session UI resolves model/effort/permission from a priority chain (`packages/happy-app/sources/sync/SessionView.tsx`, ~L460-486). The **highest-priority** layer is a per-session override stored only in that device's MMKV and explicitly **not synced**:

```ts
// packages/happy-app/sources/sync/storageTypes.ts:163-165
permissionMode?: string | null; // Local permission mode key, not synced to server
modelMode?: string | null;      // Local model key, not synced to server
effortLevel?: string | null;    // Local effort level key, not synced to server
```

So when you pick a model on one device it writes to that device's `session-model-modes` MMKV key (`persistence.ts`, ~L209-245) and never leaves the device. The other client sees `session.modelMode === null` and falls through to the synced `agentDefaultOverrides` / code default (e.g. opus) or the CLI-reported `metadata.currentModelCode`. Effort has no metadata fallback at all, so a per-device effort pick simply doesn't appear on the other client.

### Why this looks unintentional (git archaeology)

The "not synced to server" behavior appears to have been inherited by copy-paste, not decided per field:

- **`35527e93` (Aug 2025) "session draft persistence and permission mode storage"** introduced device-local MMKV persistence for *draft messages* — which correctly should be device-local — and piggybacked `permissionMode` onto the same machinery. The `// not synced to server` comment was copied straight from the adjacent `draft` field.
- **`ca6a0562` → `bb7a1173` → `8baf25b0`** then added `modelMode` and `effortLevel`, each duplicating the permission-mode line **including the comment verbatim**. No separate "should this sync?" decision appears in the history.
- **`bb7a1173`**'s own title — *"metadata-driven model/mode selection with sync mode hacks"* — describes the synced `currentModelCode` metadata fallback as "hacks" layered under the local override.

[...]

PR #1580, opened 2026-07-26
feat(app): sync the unread read-position across clients via session metadata

chphch:
Implements #1630.

> **Depends on #1374.** GitHub can't base a PR on a branch outside this repo, so this one carries #1374's three commits underneath. **Only the last commit (`feat(app): sync the unread read-position across clients via session metadata`) is new here.** Merge #1374 first and this collapses to that single commit; happy to rebase whenever it lands.

## Problem

Unread is local-only: an in-memory `unreadSessionIds` set, filled when a session goes thinking → idle and cleared on view. Two consequences:

- Reading a session on the phone leaves it unread on the desktop, and vice versa.
- Every unread mark disappears on app restart — the set is never persisted.

## Approach

Derive unread from a synced read position instead of tracking a local set.

| Value | Meaning | Lives in |
|---|---|---|
| `metadata.lastReadSeq` | highest message seq the user has read, on any device | session metadata (synced) |
| `Session.lastMessageSeq` | highest message seq that has arrived | local, mirrored from the sync engine's cursor |

`hasUnread = lastMessageSeq > lastReadSeq`, minus the session currently open.

It rides the existing session-metadata channel — the same one the permission/model/effort picks use (#1492) — so **no server change is needed**. Writes go through `sync/ops` (`sessionMarkRead`, `sessionMarkUnread`) with the same optimistic-push + inbound-reconcile handling: the local value applies immediately, and while our push is in flight an echoed metadata update can't roll it back (`isAgentModePushPending`).

Two details worth calling out, both of which caused real bugs while building this:

- **`Session.seq` can't stand in for `lastMessageSeq`.** Metadata writes bump `Session.seq` too, so storing a read position would itself look like a new message and immediately re-mark the session unread. Hence the separate message-only mirror.
- **Absent `lastReadSeq` counts as read.** Otherwise the first run after upgrading flags all of history unread. This also matches the previous behavior, where a restart cleared unread.

## Also in this commit

Both fall out of the model change rather than being separate features:

- **"Mark as unread"** in the session long-press menu. It only becomes meaningful once the read position is synced — as a local flag it wouldn't survive a restart or reach another device. Menu action ids get their own `SessionActionId` type, since this action carries no keyboard chord and `SESSION_ACTION_SHORTCUTS` stays keyed only by the ids that do (looked up through `getSessionActionShortcut`).
- **`buildSessionListViewData`'s second parameter is now required.** It used to be an optional unread set, and three reducers (`applyMachines`, `deleteMachine`, and the draft update) omitted it — so any rebuild through those paths silently dropped every unread mark until the next `applySessions`. Making it required turns that whole class of omission into a compile error.

## Verification

`pnpm typecheck` clean; app suite 779/779. Running on a self-hosted build with multiple clients (phone + desktop web): reading on one device clears the unread mark on the other, and "Mark as unread" brings it back on both.


chphch:
Closing — this is the wrong shape to be asking you to review, even though I think the problem is real.

It lands in the "core refactors — sync engine … discuss first" bucket from CONTRIBUTING (`storage.ts` +86/−68, a new `ops.ts`), it's stacked on #1374 which is itself conflicting, and it's conflicting now too. That's three things you'd have to untangle before you could even judge the idea.

The underlying problem stands: unread is an in-memory set, so reading a session on the phone leaves it unread on the desktop, and every mark disappears on restart. I've left #1630 open with the design written up — if the read-position approach is one you'd want, I'll come back with a single clean PR against current main rather than this stack.

Running it on my fork meanwhile.

PR #1078, opened 2026-04-15
feat(app): star/bookmark conversations with pin-to-top

chphch:
Closes #913.

**What this adds:** Star/bookmark for conversations so important sessions are easy to find. Long-press (mobile) or right-click (web) on any conversation surfaces a Star/Unstar action; starred conversations float to the top of the session list with a gold ★ badge on the avatar; state persists locally via MMKV (not synced to server, same pattern as `draft` / `permissionMode`). Pure feature — no test infrastructure changes; e2e tests for star/unstar can land in a follow-up once #1069 (E2E smoke infra) is merged.

## Proof

Long-press (mobile) / right-click (web) → Star → gold ★ badge appears and the session floats to the top:

star/bookmark conversation

## Scope

6 files, +100 / -13. No new dependencies. No test or CI changes.

## Behaviour

| Location | Behaviour |
|---|---|
| Active sessions | Starred sessions sorted first within the active group |
| Inactive sessions | Dedicated **Starred** section at the top, above all date groups |
| Avatar badge | Gold star icon (hidden when a draft badge is already shown) |

## Implementation

- **`persistence.ts`** — `loadStarredSessions` / `saveStarredSessions` (MMKV key `starred-sessions-v1`)
- **`storageTypes.ts`** — `Session.starred?: boolean` (local-only, same pattern as `draft` / `permissionMode`)
- **`storage.ts`** — `SessionRowData.starred`, `toggleSessionStarred()` action, sort logic, dedicated "Starred" header section
- **`useVisibleSessionListViewData.ts`** — keeps starred active sessions visible in the visible-list view
- **`useSessionQuickActions.ts`** — Star / Unstar action item in the quick-action menu
- **`SessionsList.tsx`** — Gold ★ badge on avatar

Issue #913, opened 2026-03-24
Feature Request: Star/Bookmark conversations for easy retrieval


DoloresSusu:
### Problem
When a conversation goes idle and turns gray, there's no way to mark/bookmark it for easy retrieval later. In a long list of conversations, it's hard to find the one you want to come back to.

### Proposed Solution
Add a **Star / Bookmark / Pin** feature for conversations, so users can:

1. Star/bookmark important conversations they want to revisit
2. Filter or view a "Starred" list in the sidebar
3. Quickly find high-value conversations without scrolling through the full history

### Use Case
I had a great session where Claude built me a personal GitHub Pages site. After some time away, the conversation grayed out and I wanted to mark it to come back later — but had no way to do so other than remembering the title.

### Additional Context
This is a common pattern in chat/messaging apps (Slack starred messages, Telegram saved messages, etc.) and would significantly improve the UX for power users who have many conversations.

---
Submitted via Happy + Claude Code

PR #1581, opened 2026-07-26
feat(app): star projects to pin them to the top of the session list

chphch:
## Problem

The session list groups sessions into project cards. Past a handful of projects, the two or three you actually work on daily sit wherever the ordering puts them — so finding them means scanning the whole list, every time.

## What this adds

A star toggle on each project card header. Starred projects rise to the top of their machine group.

The starred set lives in `settings.starredProjects`, so it rides the ordinary settings sync: star a project on the laptop and it is starred on the phone.

**Starring is per repo, not per worktree.** `projectStarKey` maps a worktree path back to its repo, so starring a repo lifts all of its worktrees together, and a worktree can't drift into a different position from the repo it belongs to.

**Rig projects show no star.** They carry a durable project identity rather than one working directory, so there is no machine-and-path key to store — `ProjectGroupData.path` is null for them and the control is hidden.

## Where the ordering runs

Inside `buildSessionProjectDisplayGroups`, the pass that assembles the machine → project hierarchy just before render.

That placement is forced rather than chosen. The pass ends by sorting each machine group by project name, so an order applied anywhere upstream of it is simply overwritten — the card would keep its star and stay exactly where the alphabet put it. Sorting by the star inside that pass, with name and id demoted to tiebreakers, is what actually moves it.

A star lifts a project **within its own machine group**, never above another machine's header, because the grouping happens before the sort.

`getSessionShortcutIdsInDisplayOrder` walks the same pass and takes the same starred set, so the ⌘1–9 badges count in the order that is on screen rather than one nobody can see.

The starred set is read in `SessionsList` via `useSessionListStarredProjects`, so a settings write re-runs the grouping. It also rides along in `extraData`: web re-renders eagerly, but native `FlatList` memoizes cells and would otherwise keep showing the old order until something else forced a repaint.

## Supporting change: `utils/projectPath.ts` (new)

The pure path logic — worktree markers, `isWorktreePath`, `getRepoPath`, the project key, and the star key — moves out of `utils/worktree.ts`. That module imports `sync/ops` and so drags React Native into anything importing it, including a unit test, which is why these rules had no coverage. The new module is dependency-free; `utils/worktree.ts` re-exports the moved helpers, so **no import site changes**.

It also collapses a duplication that was already there: `${machineId}:${path}` was being assembled by hand in `gitStatusSync` and in `storage.getSessionPathKey`. The latter now uses `projectKey`. (I left `gitStatusSync`'s private copy alone to keep this diff to the feature — happy to fold it in if you'd prefer.)

## Supersedes #1078

#1078 was the earlier, session-level take on the same itch (star individual conversations). Having dogfooded both, project-level is the one that stuck — sessions are transient, so their stars accumulate as clutter, while the project you work in is stable and is the thing you actually navigate to. Session-level starring turned out to be dead UI on my own build: weeks of running it and I never reached for it.

**#1078 is now closed**, so there is no choice to weigh here — this is the version that replaced it, not an alternative to it.

## Verification

`pnpm typecheck` clean; app suite 1125/1125.

Eleven tests cover the rules:

- `utils/projectPath.spec.ts` (3) — the star key: plain checkout, worktree inherits its repo, cross-machine isolation.
- `utils/sessionDisplayOrder.test.ts` (7) — the ordering: a star lifts a project above the alphabetical order of its machine; projects sharing a starred state keep alphabetical order; nothing starred leaves the order alone; a worktree card rides up on its repo's star; a card with no path can't be starred; a star stays inside its own machine group; and the session shortcut numbers follow the starred order.
- `hooks/useVisibleSessionListViewData.test.ts` (1) — the hook hands the ordering pass the set it reads.

### DIFF
--- a/packages/happy-app/sources/components/CommandPalette/CommandPaletteProvider.tsx
+++ b/packages/happy-app/sources/components/CommandPalette/CommandPaletteProvider.tsx
@@ -15,7 +15,7 @@ import {
     getPreferredShortcutModifier,
 } from '@/keyboard/shortcuts';
 import { isTauri } from '@/utils/isTauri';
-import { useVisibleSessionListViewData } from '@/hooks/useVisibleSessionListViewData';
+import { useSessionListStarredProjects, useVisibleSessionListViewData } from '@/hooks/useVisibleSessionListViewData';
 import { getSessionShortcutIdsInDisplayOrder } from '@/utils/sessionDisplayOrder';
 import { t } from '@/text';
 
@@ -28,6 +28,9 @@ export function CommandPaletteProvider({ children }: { children: React.ReactNode
     const commandPaletteEnabled = storage(useShallow((state) => state.localSettings.commandPaletteEnabled));
     const sessionListViewData = useVisibleSessionListViewData();
     const machines = useAllMachines();
+    // The number badges follow what is on screen, so they have to be built from
+    // the same starred set the list orders its project cards by.
+    const starredProjects = useSessionListStarredProjects();
     const navigateToSession = useNavigateToSession();
     const preferredModifier = useMemo(() => getPreferredShortcutModifier(
         typeof navigator === 'undefined' ? undefined : navigator
@@ -37,7 +40,8 @@ export function CommandPaletteProvider({ children }: { children: React.ReactNode
         sessionListViewData,
         machines,
         t('status.unknown'),
-    ), [machines, sessionListViewData]);
+        starredProjects,
+    ), [machines, sessionListViewData, starredProjects]);
 
     // Define available commands
     const commands = useMemo((): Command[] => {
--- a/packages/happy-app/sources/components/ProjectGroup.tsx
+++ b/packages/happy-app/sources/components/ProjectGroup.tsx
@@ -6,7 +6,7 @@ import { StyleSheet, useUnistyles } from 'react-native-unistyles';
 import { Text } from '@/components/StyledText';
 import { Typography } from '@/constants/Typography';
 import { t } from '@/text';
-import { ProjectGroupData, ProjectWorkspaceGroup, useSessionGitStatus } from '@/sync/storage';
+import { ProjectGroupData, ProjectWorkspaceGroup, useIsProjectStarred, useSessionGitStatus, storage } from '@/sync/storage';
 import { CompactSessionRow } from './ActiveSessionsGroupCompact';
 import { Avatar } from './Avatar';
 import { requestHomeDockFocus } from './homeDockFocus';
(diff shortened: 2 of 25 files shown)


chphch:
Force-pushed two corrections into the commit (no new commits — the PR is still one commit).

**1. Starring did not reorder the list until a remount.**

`useStarredProjects` built the `Set` inside the store selector and wrapped it in `useDeepEqual`. That helper uses `fast-deep-equal`, whose default build has no `Set` support — it falls through to comparing `Object.keys()`, which is always `[]` for a `Set`:

```js
const equal = require("fast-deep-equal");
equal(new Set(["x"]), new Set(["x", "y"]));  // => true
```

So every `Set` compared equal, the zustand subscription never fired, and the sort kept the `Set` captured at mount. The star icon still flipped (that reads `useIsProjectStarred`, which returns a boolean), which is why it looks like only the ordering is broken.

Fixed by subscribing to the underlying array with `useShallow` and deriving the `Set` in a `useMemo`.

Verified in a browser against a local server with two sessions in different cwds: before, the order only changed after a reload; after, it changes on the click. Same script on both builds, so the difference is the code, not the harness.

**2. The filled star rendered grey.**

`theme.colors.warning` is `#8E8E93` in both the light and dark palettes, so a starred project was indistinguishable from an unstarred one by colour — only the glyph fill differed. Changed to `#f5a623`.

`pnpm typecheck` clean, 787 tests pass.

chphch:
Rebased onto current `main` (`3cf9ce70f`), and the UI anchor had to move — the home rework deleted the row this PR hung the star on.

**What changed and why**

- **The star lives in the worktree header now.** `c63c80ff3` / `6dc878dc3` removed `ProjectGroup`'s own header (the collapsible row with the chevron, machine subtitle and session count) and gave each worktree its own header carrying the project name. The star had no row left to sit in, so it now renders in that header, immediately left of the new-session `+`.
- **Once per project, not once per worktree.** Every worktree of a repo shares one star — upstream's grouping already collapses a worktree path to its repo path (`buildPathProjectGroups` keys on `getRepoPath(...)`), so repeating the control per section would show several stars that are really one. `ProjectGroup` computes the star state once and passes `showStar` to the first section only.
- **`ProjectGroupData.path` is optional** (`path?: string | null`) rather than required, so upstream's own `ProjectGroupData` fixtures keep type-checking without being touched.
- **Dropped the `utils/worktree.ts` re-export churn and the duplicated path helpers.** `utils/worktreePaths.ts` (added by `c63c80ff3`) already is the dependency-free module this PR was creating; `utils/projectPath.ts` now re-exports from it and defines only `projectKey` / `projectStarKey`.

The feature itself is untouched upstream — nothing lets a user promote the projects they actually work in, and the list is now sorted per-machine alphabetically, so recency no longer surfaces daily work either.

`pnpm typecheck` clean; app suite 929/930 (the one failure is `encryption/blob.test.ts`'s 1 MB case timing out under load — it passes standalone on plain `main` at ~1.85 s, and this PR touches no encryption code).

chphch:

Two independent requests for this are open:

- **#1707** (Aug 20) — pin projects / manual ordering, because recency-ordering makes the list jumpy once there are many folders and muscle memory never forms. That is what this PR does.
- **#913** (Mar 24, 1 👍) — the same idea one level down, for conversations rather than projects.

This PR covers the project level (#1707); #913's session-level starring would sit naturally on top of the same stored set.

@bra1nDump — is the project level the right place to start, or would you rather see it at the session level first?

PR #1329, opened 2026-05-25
feat(app): surface Hermes agent in client (follow-up to #1322)

chphch:
## Summary

- Follow-up to #1322 (CLI-only Hermes Agent integration). PR #1322 mirrors the Kimi/#1276 "self-hosted-CLI" shape, which lands Hermes in `happy-cli` but does not surface it in the app — the agent picker, icons, flavor labels, defaults and option pickers are all hardcoded to the prior agent set, so a CLI advertising `hermes: true` is invisible to the user.
- This PR threads `'hermes'` through every app-side type/list so the existing CLI bridge becomes user-reachable. **Stacked on #1322** — it has to be merged after (or with) #1322. The diff vs `main` of this PR alone is app-only; combined with #1322 it gives a complete end-to-end Hermes flow.

## What changes

- **Agent picker + icon** (`app/(app)/new/index.tsx`, `app/(app)/dev/session-composer.tsx`) — new `icon-hermes{,@2x,@3x}.png` purple "H" monogram placeholder until Nous Research ships an official asset.
- **Agent Defaults screen** (`app/(app)/settings/agents.tsx`) — Hermes section with permission/model rows.
- **Flavor → label** (`app/(app)/session/[id]/info.tsx`) → `'Hermes'`.
- **Type unions**: `NewSessionAgentType`, `ops.startSession` agent option, `AgentInput.agentType`, `modelModeOptions.AgentFlavor`, raw `provider` enum in `sync/typesRaw.ts`.
- **`sync/agentDefaults`**:
  - `agentKeys` includes `'hermes'`.
  - `AgentDefaultOverridesSchema.hermes` slot added (optional, like other agents).
  - `codeAgentDefaults.hermes` set to neutral defaults — model picker is populated at runtime from ACP `available_models`, so the static fallback intentionally stays empty.
- **`components/modelModeOptions`**:
  - `getHermesPermissionModes` exposes only `default` (Hermes ACP currently doesn't advertise operating modes — `hooks_auto_accept` is CLI-side).
  - `getHermesModelModes` ships a sentinel `default`; the real list comes from `metadata.models` populated by the ACP `available_models` event at session start.
- **`sync/storageTypes.cliAvailability.hermes`**: `z.boolean().optional()` for backward-compat with daemons that predate Hermes.
- **i18n**: 11 locale tables (`_default.ts` + 10 translations: en/zh-Hans/es/ru/pl/zh-Hant/pt/it/ja/ca) gain the `agents.hermes = 'Hermes'` row.

[...]

PR #1322, opened 2026-05-23
feat(cli): add Hermes Agent CLI integration via ACP

chphch:
Adds first-class support for [NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent) via the Agent Client Protocol, mirroring the pattern in #1276 (Kimi CLI). Hermes ships ACP support out of the box through `hermes acp` (installed via `pip install 'hermes-agent[acp]'`), so this PR is a small factory + runner + subcommand wiring — no new ACP infrastructure needed.

Closes #1129.

## What this adds

- **Factory** `packages/happy-cli/src/agent/factories/hermes.ts` (+ unit tests) — builds an `AcpBackend` with `command='hermes'`, `args=['acp']`, optional `HERMES_API_KEY` / `HERMES_MODEL` env passthrough.
- **Entry wrapper** `packages/happy-cli/src/hermes/runHermes.ts` — thin wrapper around the generic ACP runner, plus a `hermesCliAvailable()` pre-flight that prints a `pip install` hint and exits with code 1 when the `hermes` binary is missing from PATH.
- **Subcommand** `happy hermes` runs the pre-check **before** `authAndSetupMachineIfNeeded()` and `ensureDaemonRunning()` so first-time users without Hermes installed are not dragged through the QR flow only to hit an install-hint error.
- **Wiring** `hermes` is added to `KNOWN_ACP_AGENTS`, `resolveSessionFlavor`, the `AgentId` union, the `BackendFlavor` union, the factories barrel, and `initializeAgents()`.
- **Docs** README's agents list and commands table, plus `happy --help`, now list `happy hermes` alongside the existing agents.

## Why mirror #1276 (Kimi) rather than #1317 (Devin)?

Hermes Agent is a self-hosted Python CLI with its own auth/config — no extra app- or daemon-side UI surface is needed beyond the existing ACP plumbing. The Kimi PR's factory + runner + one-subcommand shape is the smallest correct surface for that case. (Happy to expand if maintainers prefer the Devin-style integration.)

## Proof

CLI smoke test of the missing-binary path (default for anyone who hasn't installed Hermes yet):

\`\`\`
\$ PATH=/usr/bin:/bin ./bin/happy.mjs hermes
Error: \`hermes\` CLI not found on PATH.

Hermes Agent is a Python package from Nous Research.
Install with ACP support:

  pip install 'hermes-agent[acp]'

Or use uv:

  uvx --from 'hermes-agent[acp]' hermes-acp

[...]

PR #1584, opened 2026-07-26
fix(app): stop seeding ACP session defaults from Claude's table

### LINKED ISSUE #1583: Generic ACP sessions get Claude Code's agent defaults: agentKeys omits 'acp' and 'opencode'
## Summary

`agentKeys` in `packages/happy-app/sources/sync/agentDefaults.ts` omits two flavors that `happy-cli` actively produces: **`'opencode'`** and **`'acp'`**. `normalizeAgentKey()` silently collapses both to `'claude'`, so sessions for those agents are seeded from **Claude Code's** defaults table.

`'opencode'` is a shipped, documented agent (`happy acp opencode`), so this affects a released integration today, not a hypothetical one.

All line references are against `3c0155c254f3518a9d73d6e6f2fd9cfd86e0b1e3`.

## The mismatch

The CLI produces three ACP-side flavor values:

```ts
// packages/happy-cli/src/agent/acp/runAcp.ts:439
function resolveSessionFlavor(agentName: string): 'gemini' | 'opencode' | 'acp' {
  if (agentName === 'gemini')   return 'gemini';
  if (agentName === 'opencode') return 'opencode';
  return 'acp';
}
```

`'acp'` and `'opencode'` are both first-class members of `BackendFlavor` (`packages/happy-cli/src/utils/createSessionMetadata.ts:23`).

The app's defaults table recognises neither:

```ts
// packages/happy-app/sources/sync/agentDefaults.ts:3
export const agentKeys = ['claude', 'codex', 'gemini', 'openclaw', 'agy'] as const;

// :40-45
export function normalizeAgentKey(flavor: string | null | undefined): AgentKey {
    if (flavor === 'codex' || flavor === 'gemini' || flavor === 'openclaw' || flavor === 'agy') {
        return flavor;
    }
    return 'claude';
}
```

So `normalizeAgentKey('acp') === 'claude'` and `normalizeAgentKey('opencode') === 'claude'`, and the lookup lands on:

```ts
// :33
claude: { permissionMode: 'bypassPermissions', modelMode: 'opus', effortLevel: 'medium' },
```

## What this reaches

Confirmed reachable read paths:

- `packages/happy-app/sources/-session/SessionView.tsx:667-669` — `resolveAgentDefaultConfig(agentDefaultOverrides, flavor)` with the live session's flavor. An `acp`/`opencode` session's effective defaults resolve to `bypassPermissions` / `opus` / `medium`.
- `packages/happy-app/sources/components/modelModeOptions.ts:323-329` — `getDefaultModelKey(flavor)` and `getDefaultPermissionModeKey(flavor)` both route through `getCodeAgentDefaults`, same collapse.

`opus` and `bypassPermissions` are Claude-specific keys. For an ACP agent they cannot match anything the agent advertises, so the seeded default is not merely cosmetically wrong, it is unmatchable.

## Why this looks unintentional rather than by design

The same file tree already handles unknown flavors **correctly** one function over:

```ts
// packages/happy-app/sources/components/modelModeOptions.ts:291-300
if (flavor === 'claude' || flavor === 'codex' || flavor === 'openclaw' || flavor === 'agy') {
    return hackModes(getHardcodedPermissionModes(flavor, translate));
}
const metadataModes = mapMetadataOptions(metadata?.operatingModes);
if (metadataModes.length > 0) {
    return hackModes(metadataModes);
}
return hackModes(getHardcodedPermissionModes(flavor, translate));
```

`getAvailablePermissionModes` deliberately falls through for non-hardcoded flavors and reads the agent's advertised `operatingModes` — which is exactly right for ACP, where the agent sends its own modes over the wire. The *available options* path respects unknown flavors; the *defaults* path collapses them to Claude. That asymmetry inside one component is what makes this read as an oversight.

It also gets worse rather than better over time: #1575 is deprecating the `gemini` surfaces, and `gemini` is currently the only ACP-derived flavor that `agentKeys` covers. Once it goes, every ACP agent is on the collapsed path.

## Reproduction

```bash
happy acp -- hermes acp     # any ACP binary reproduces this; hermes-agent ships one
```

The session registers with `flavor: 'acp'` (visible in `~/.happy/sessions.json` under `metadata.flavor`), and its defaults resolve through the `'claude'` branch.

Verified on `happy` 1.1.9-beta.0, macOS 15 arm64. `~/.happy/sessions.json` excerpt:

```
cms1xodbwz4j7wc0unvvcolaa   flavor='acp'   path=~/hermes-mobile   startedBy=terminal
```

## Suggested shape

Add `'opencode'` and `'acp'` to `agentKeys` with neutral defaults, plus their `AgentDefaultOverridesSchema` slots. The precedent for "neutral" here is #1329, which sets Hermes's static fallback empty on the grounds that the model picker is populated at runtime from ACP `available_models` — the same reasoning applies to every generic ACP flavor.

I have not sent a PR because the actual default values are a maintainer taste call, and because I did not want to collide with #1322 / #1329, which are already in flight for the *named* `hermes` flavor. This issue is complementary to those: even after both land, `happy acp -- <any other agent>` still resolves through the `'claude'` branch. Happy to send the patch if you tell me what defaults you want.

## Scope notes, so this is not over-read

- The **write** path looks unreachable for these flavors: `setAgentDefaultOverride` is only called from `app/(app)/settings/agents.tsx`, which iterates `agentKeys`, so there is no UI surface that writes an `'acp'` override into the `'claude'` slot. I checked for that specifically and could not find a path to it. Read-side seeding is the confirmed defect.
- I have **not** traced whether the seeded `bypassPermissions` survives `resolveCurrentOption()` into an applied permission mode, since that depends on what the agent advertises in `operatingModes`. I am reporting the defaults lookup as wrong, not claiming a permission bypass.
- `getHardcodedEffortLevels` (`modelModeOptions.ts:352-355`) already returns `[]` for unknown flavors, so effort levels are unaffected.

Related: #1322, #1329, #1129, #1575.

hansai-art:
Fixes #1583.

## The bug

`agentKeys` in `sources/sync/agentDefaults.ts` omits `'acp'` and `'opencode'`, two flavors `happy-cli` actively reports (`agent/acp/runAcp.ts:439` `resolveSessionFlavor`, both first-class in `BackendFlavor`). `normalizeAgentKey()` collapsed anything unrecognised to `'claude'`, so those sessions were seeded from Claude Code's row:

```ts
claude: { permissionMode: 'bypassPermissions', modelMode: 'opus', effortLevel: 'medium' },
```

`opus` and `bypassPermissions` are Claude-specific keys. An ACP agent advertises its own models and modes over the wire, so for an ACP session the seeded default is not merely the wrong choice, it is unmatchable. `'opencode'` ships today via `happy acp opencode`, so this is a released integration, not a hypothetical one.

Confirmed read paths in #1583: `-session/SessionView.tsx:667-669` (`resolveAgentDefaultConfig` with the live flavor) and `components/modelModeOptions.ts:323-329` (`getDefaultModelKey` / `getDefaultPermissionModeKey`).

## The fix

`matchAgentKey()` does exact membership against `agentKeys` and returns `null` for anything else. Unrecognised flavors then get neutral defaults instead of falling through to claude:

```ts
const NEUTRAL_AGENT_DEFAULTS: AgentDefaultConfig = {
    permissionMode: 'default', modelMode: 'default', effortLevel: null,
};
```

`getAgentDefaultOverride` and `setAgentDefaultOverride` stop touching the `claude` slot for a flavor that has no row of its own. `normalizeAgentKey()` keeps its claude fallback, so nothing that depends on getting an `AgentKey` back changes behaviour.

**`agentKeys` is deliberately unchanged.** That array drives `settings/agents.tsx:143`, which renders a Permission field for every entry from `getHardcodedPermissionModes(agent, t)` — and that function *also* falls through to `getClaudePermissionModes()` (`modelModeOptions.ts:146-160`). Adding `'acp'` there would put Claude's permission modes under an ACP heading in the settings screen, with no session metadata available to correct them. That is the fix I started with and abandoned; this narrower one has no UI surface.

On the values: #1583 said I'd wait for a maintainer to name the defaults, and that is still your call. I'm sending the patch anyway because the two halves separate cleanly — the wrong lookup isn't a taste question, and the values are one object literal to change. `'default'` / `'default'` / `null` is the most conservative option available and matches the existing `openclaw` row rather than inventing a convention; #1329's reasoning (static fallback stays empty because the model picker is populated at runtime from ACP `available_models`) is the precedent. Say the word and I'll change them.

## Relation to #1322 / #1329

Complementary, no overlap. Those add a *named* `hermes` flavor; this fixes the *generic* path, which survives them — after both land, `happy acp -- <any other agent>` still resolves through the `'claude'` branch. No file touched here is touched there.

## Tests

8 new tests in `sources/sync/agentDefaults.test.ts`, following the conventions of the sibling `modelModeOptions.test.ts`.

They are not vacuous. With the test file present and `agentDefaults.ts` reverted to `3c0155c`, **5 of the 8 fail** with exactly the bug's signature:

```
✓ keeps agents that have a static profile resolving to their own row
× does not hand Claude Code defaults to flavors with no static profile
  → expected { …(3) } to not deeply equal { …(3) }
× gives flavors with no static profile neutral defaults
  → expected 'bypassPermissions' to be 'default'
× covers every flavor an ACP session can report
  → expected { …(3) } to not deeply equal { …(3) }
× does not read the claude override slot for flavors with no static profile
  → expected { modelMode: 'sonnet' } to deeply equal {}
× does not write into the claude override slot for flavors with no static profile
  → expected { claude: { modelMode: 'something' } } to deeply equal {}
✓ still writes and clears overrides for agents that have a profile
✓ leaves normalizeAgentKey falling back to claude for unknown input

Tests  5 failed | 3 passed (8)
```

With the fix: 8 passed.

- `pnpm --filter happy-app typecheck` → **0 errors** (this is what `typecheck.yml` runs on the PR).
- Full `happy-app` suite: 786 passed, 1 failed. The failure is `sources/encryption/blob.test.ts > should encrypt and decrypt a large blob (1MB)`, `Test timed out in 5000ms` — it also fails on unmodified `3c0155c`, and passes in 2.3s when that file is run alone. Pre-existing flakiness under parallel load, unrelated to this change. Flagging it rather than quietly reporting green.

`happy-app` tests aren't run by CI (`typecheck.yml` only runs `tsc`), so the above are local runs: macOS 15 arm64, Node 20, vitest via the workspace binary.

## Scope

Two files, +117/−7, no dependencies, no UI change, no behaviour change for `claude` / `codex` / `gemini` / `openclaw` / `agy` (first and seventh tests pin that). I did not touch `AgentDefaultOverridesSchema`; adding slots there is only needed if you also want these flavors editable in settings, which is the UI question above and a separate decision.

### DIFF
--- a/packages/happy-app/sources/sync/agentDefaults.test.ts
+++ b/packages/happy-app/sources/sync/agentDefaults.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, it } from 'vitest';
+import {
+    agentKeys,
+    getAgentDefaultOverride,
+    getCodeAgentDefaults,
+    normalizeAgentKey,
+    resolveAgentDefaultConfig,
+    setAgentDefaultOverride,
+} from './agentDefaults';
+
+// `resolveSessionFlavor()` in happy-cli (src/agent/acp/runAcp.ts) can emit exactly these three
+// values for an ACP-launched session. 'gemini' has a row in `agentKeys`; the other two do not.
+const ACP_SESSION_FLAVORS = ['gemini', 'opencode', 'acp'] as const;
+const FLAVORS_WITHOUT_PROFILE = ['opencode', 'acp'] as const;
+
+describe('agentDefaults', () => {
+    it('keeps agents that have a static profile resolving to their own row', () => {
+        for (const key of agentKeys) {
+            expect(getCodeAgentDefaults(key)).toBe(getCodeAgentDefaults(key));
+            expect(normalizeAgentKey(key)).toBe(key);
+        }
+        // Distinct agents must not share a defaults row, otherwise a collapse would go unnoticed.
+        expect(getCodeAgentDefaults('codex')).not.toEqual(getCodeAgentDefaults('claude'));
+    });
+
+    it('does not hand Claude Code defaults to flavors with no static profile', () => {
+        for (const flavor of FLAVORS_WITHOUT_PROFILE) {
+            expect(getCodeAgentDefaults(flavor)).not.toEqual(getCodeAgentDefaults('claude'));
+        }
+    });
+
+    it('gives flavors with no static profile neutral defaults', () => {
+        for (const flavor of FLAVORS_WITHOUT_PROFILE) {
+            const defaults = getCodeAgentDefaults(flavor);
+            expect(defaults.permissionMode).toBe('default');
+            expect(defaults.modelMode).toBe('default');
+            expect(defaults.effortLevel).toBeNull();
+        }
+    });
+
+    it('covers every flavor an ACP session can report', () => {
+        for (const flavor of ACP_SESSION_FLAVORS) {
+            const defaults = getCodeAgentDefaults(flavor);
+            const hasOwnProfile = (agentKeys as readonly string[]).includes(flavor);
+            if (hasOwnProfile) {
+                continue;
+            }
+            expect(defaults).not.toEqual(getCodeAgentDefaults('claude'));
+        }
+    });
+
+    it('does not read the claude override slot for flavors with no static profile', () => {
+        const overrides = { claude: { modelMode: 'sonnet' } };
+        expect(getAgentDefaultOverride(overrides, 'claude')).toEqual({ modelMode: 'sonnet' });
+        for (const flavor of FLAVORS_WITHOUT_PROFILE) {
+            expect(getAgentDefaultOverride(overrides, flavor)).toEqual({});
+            expect(resolveAgentDefaultConfig(overrides, flavor).modelMode).not.toBe('sonnet');
+        }
+    });
+
+    it('does not write into the claude override slot for flavors with no static profile', () => {
+        for (const flavor of FLAVORS_WITHOUT_PROFILE) {
+            const next = setAgentDefaultOverride({}, flavor, 'modelMode', 'something');
+            expect(next).toEqual({});
+        }
+        // An existing override must survive an unmatched-flavor write untouched.
+        const existing = { claude: { modelMode: 'sonnet' } };
+        expect(setAgentDefaultOverride(existing, 'acp', 'modelMode', 'something')).toEqual(existing);
+    });
+
+    it('still writes and clears overrides for agents that have a profile', () => {
+        const written = setAgentDefaultOverride({}, 'codex', 'modelMode', 'something');
+        expect(written.codex).toEqual({ modelMode: 'something' });
+        expect(setAgentDefaultOverride(written, 'codex', 'modelMode', null)).toEqual({});
+    });
+
+    it('leaves normalizeAgentKey falling back to claude for unknown input', () => {
+        // Kept for callers that need a concrete key. The defaults lookups deliberately do not
+        // use this fallback any more.
+        expect(normalizeAgentKey('acp')).toBe('claude');
+        expect(normalizeAgentKey(null)).toBe('claude');
+        expect(normalizeAgentKey(undefined)).toBe('claude');
+        expect(normalizeAgentKey('')).toBe('claude');
+    });
+});
--- a/packages/happy-app/sources/sync/agentDefaults.ts
+++ b/packages/happy-app/sources/sync/agentDefaults.ts
@@ -37,22 +37,41 @@ const codeAgentDefaults: Record<AgentKey, AgentDefaultConfig> = {
     agy: { permissionMode: 'default', modelMode: 'Gemini 3.1 Pro (High)', effortLevel: null },
 };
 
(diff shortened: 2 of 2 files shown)


hansai-art:
CI note: this PR shows 0 check-runs and `unstable` because the `Expo App TypeScript typecheck` run is sitting at `action_required` (contributor approval gate), so it has not started rather than failed.

Two other contributors are in the same state on this repo right now — `arozanov` and `sandersjdd` — listable as `conclusion=action_required` in the Actions run list. One approval each clears it, permanently per contributor.

I ran the gated check locally in the meantime: `pnpm --filter happy-app typecheck` → **0 errors**, which is exactly what that workflow runs.

Issue #1516, opened 2026-07-13
Happy launches a stale npm claude install instead of native installer binary


SebastianAigner:
I have Claude Code installed via the native installer (`~/.local/bin/claude`, currently v2.1.207), and it's configured to authenticate through our org's internal proxy via a `settings.json` `apiKeyHelper`. This works fine when I run `claude` directly.

When launching through Happy, the same setup failed with:
```
Invalid API key · Please run /login
```
even though credentials, `ANTHROPIC_BASE_URL`, and `apiKeyHelper` were all correctly configured.

Via `happy --version`: Turns out happy wasn't launching my native-installer binary at all, it was using a much older `@anthropic-ai/claude-code` install from a global NPM install which I had forgotten about:
```
Using Claude Code v2.1.17 from npm
```

As a **workaround**, setting `HAPPY_CLAUDE_PATH` to my native-installer binary explicitly fixed it:
```sh
export HAPPY_CLAUDE_PATH="$HOME/.local/bin/claude"
```
After this, `happy --version` correctly reports:
```
Using Claude Code v2.1.207 from HAPPY_CLAUDE_PATH
```
(and authentication against our proxy works as expected)

I've had Claude try to locate the offending bits, here's what it had to say:

> Looking at `findClaudeInPath()` in `packages/happy-cli/scripts/claude_version_utils.cjs`, it resolves `claude` from `PATH`, follows the symlink to the real file, and only treats it as directly executable if the resolved path ends in `.js`, `.cjs`, or `.exe`. If not, it assumes the file is a Windows-style npm shim and tries to resolve past it via an adjacent `node_modules/@anthropic-ai/claude-code`.
> 
> The native-installer binary resolves to an extension-less file (a plain Mach-O binary), so it fails that check. There's no `node_modules` next to it either, so the lookup comes up empty and Happy silently falls through to a global npm install instead.
> 
> The comment above that logic suggests it was only meant to apply on Windows, but the check isn't actually gated by platform, so it seems to misfire for native-installer and Homebrew installs on macOS/Linux, silently downgrading to whatever else happens to be sitting in a global npm prefix.

hansai-art:
This has a fix in flight: #1591 (`mergeable`, opened 2026-07-26). Neither it nor the earlier #1394 links this issue, so it would close silently.

[...]

Issue #1543, opened 2026-07-19
happy CLI 0.12.0: `happy` falsely reports 'Claude Code is not installed' when @anthropic-ai/claude-code is 2.1.x (package no longer ships cli.js)


LiNan1984:
## Bug

Running bare `happy` on a machine where `@anthropic-ai/claude-code` is installed globally (currently `v2.1.202+`) prints:

```
Claude Code is not installed

Please install Claude Code using one of these methods:

Option 1 - npm (recommended, highest priority):
  npm install -g @anthropic-ai/claude-code
...
```

…but `claude --version` works fine in the same shell, and a recent Claude Code is genuinely installed under `npm root -g`:
`/Users/linan/.nvm/versions/node/v22.18.0/lib/node_modules/@anthropic-ai/claude-code/`.

`happy codex` works correctly; only the default Claude Code entry point is broken.

## Reproduction

```bash
npm install -g happy-coder                 # 0.12.0
npm install -g @anthropic-ai/claude-code   # 2.1.202 or newer
happy                                      # ❌ "Claude Code is not installed"
claude --version                           # ✅ works
happy doctor                               # ✅ shows Claude Code as expected
happy codex                                # ✅ works
```

Tested on macOS arm64, Node v22.18.0, shell `zsh`. `which claude` and `which happy` both resolve inside the same `npm` bin dir.

## Expected

`happy` should detect the globally installed Claude Code and start a Claude Code session, just like it does in the bundled fallback / SDK path.

## Actual

`happy` exits with code 1 and the installer prompt above. The Claude Code session is never started.

## Root cause

`scripts/claude_version_utils.cjs` (called from `bin/happy.mjs`) looks for Claude Code exclusively via:

```js
function findNpmGlobalCliPath() {
  const globalRoot = execSync('npm root -g', { encoding: 'utf8' }).trim();
  const globalCliPath = path.join(globalRoot, '@anthropic-ai', 'claude-code', 'cli.js');
  if (fs.existsSync(globalCliPath)) {
    return globalCliPath;
  }
  return null;
}
```

…but `@anthropic-ai/claude-code` starting in the 2.1.x line no longer ships `cli.js` at the package root. The current top level of `node_modules/@anthropic-ai/claude-code/` is:

```
LICENSE.md
README.md
bin/                    # ← contains a Mach-O native binary `bin/claude.exe`
cli-wrapper.cjs         # ← Node fallback for --ignore-scripts installs
install.cjs
node_modules/
package.json
[...]

PR #1591, opened 2026-07-26
fix: detect native-installer claude binary in PATH resolution instead of treating it as an npm shim

sandersjdd:
## Problem

Claude Code's native installer (>= 2.1.113) installs `~/.local/bin/claude` as a symlink to an extensionless Mach-O/ELF binary at `~/.local/share/claude/versions/<version>`.

In `packages/happy-cli/scripts/claude_version_utils.cjs`, `findClaudeInPath()` treats any resolved PATH hit that doesn't end in `.js`/`.cjs`/`.exe` as an npm text shim (the `!isExecutable` branch), tries to resolve a `@anthropic-ai/claude-code` package next to it, and returns `null` when none exists. A native-installer binary therefore gets misclassified as a shim and the launch fails.

The existing `isPlainTextWithoutShebang` helper is not a usable inverse here — it returns `false` for both shebang scripts and binaries — so a dedicated check was needed.

## Fix

Add an `isNativeBinaryFile()` helper that sniffs the leading bytes of an extensionless file: shebang scripts and all-printable-text files (npm shims) are not native binaries; anything with non-text bytes (Mach-O `0xcf 0xfa`, ELF `0x7f 'ELF'`, etc.) is. In `findClaudeInPath()`, this check runs before the npm text-shim fallback, and a detected native binary is returned directly with the source derived via `detectSourceFromPath()`.

## Testing

- Added unit tests in `claude_version_utils.test.ts` covering Mach-O-like and ELF leading bytes (true), shebang scripts, plain text, and empty files (false).
- Full `claude_version_utils.test.ts` suite passes (55 tests).
- Verified locally on macOS against a native-installer setup (`~/.local/bin/claude` symlinked to the versioned native binary), where the previous behavior failed the launch.

### DIFF
--- a/packages/happy-cli/scripts/claude_version_utils.cjs
+++ b/packages/happy-cli/scripts/claude_version_utils.cjs
@@ -105,6 +105,44 @@ function isPlainTextWithoutShebang(filePath) {
 }
 
 /**
+ * Detect a native executable: an extensionless file whose leading bytes are
+ * not plain text (Mach-O/ELF), e.g. the native-installer binaries at
+ * ~/.local/share/claude/versions/<version>. These must not be treated as
+ * npm text shims.
+ */
+function isNativeBinaryFile(filePath) {
+    try {
+        const fd = fs.openSync(filePath, 'r');
+        try {
+            const buffer = Buffer.alloc(128);
+            const bytesRead = fs.readSync(fd, buffer, 0, buffer.length, 0);
+            if (bytesRead === 0) {
+                return false;
+            }
+            // Shebang scripts are text executables, not native binaries
+            if (bytesRead >= 2 && buffer[0] === 0x23 && buffer[1] === 0x21) {
+                return false;
+            }
+            for (let i = 0; i < bytesRead; i++) {
+                const byte = buffer[i];
+                const isPrintableAscii = byte >= 0x20 && byte <= 0x7e;
+                const isCommonWhitespace = byte === 0x09 || byte === 0x0a || byte === 0x0d;
+                if (!isPrintableAscii && !isCommonWhitespace) {
+                    // Non-text byte found (e.g. Mach-O 0xcf 0xfa, ELF 0x7f 'ELF')
+                    return true;
+                }
+            }
+            // All bytes are plain text — likely an npm shim script
+            return false;
+        } finally {
+            fs.closeSync(fd);
+        }
+    } catch (e) {
+        return false;
+    }
+}
+
+/**
  * Find path to npm globally installed Claude Code CLI
  * @returns {string|null} Path to cli.js or native binary, or null if not found
  */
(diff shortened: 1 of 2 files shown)


hansai-art:
Independent third-party verification of both the bug and this patch, on real hardware. I have no competing PR and no stake in the diff.

**Environment**: macOS 15 arm64, Claude Code native installer 2.1.220, `happy` at `3c0155c`.

```
~/.local/bin/claude -> ~/.local/share/claude/versions/2.1.220
file:  Mach-O 64-bit executable arm64      (extensionless, exactly as you describe)
```

**Before your patch**, with `PATH=~/.local/bin:/usr/bin:/bin`:

```
which claude              = /Users/…/.local/bin/claude          ← found
findClaudeInPath()        = null                                 ← discarded
detectSourceFromPath(…)   = 'native installer'                   ← same file knows the layout
```

**After applying this PR** (`git apply --3way`), same probe:

```
findClaudeInPath()        = { path: '…/.local/share/claude/versions/2.1.220',
                              source: 'native installer' }
```

Your 55 tests pass locally.

### This PR also fixes #1516, which is open and unlinked

#1516 is "Happy launches a stale npm claude install instead of native installer binary". That is this bug's downstream effect, via the `PATH → npm → Bun → Homebrew → Native` priority in `findGlobalClaudeCliPath()`: once `findClaudeInPath()` returns `null`, the npm finder is next in line and wins.

Demonstrated rather than argued. I planted a fake stale global npm install (`@anthropic-ai/claude-code` 1.0.0, `bin/claude`) via `npm_config_prefix`, with the real native 2.1.220 first in `PATH`:

```
                          before this PR                     after this PR
which claude              ~/.local/bin/claude (2.1.220)      (same)
findClaudeInPath()        null                               native installer 2.1.220
findNpmGlobalCliPath()    …/npmroot/…/claude  (1.0.0)        (same, still found)
findGlobalClaudeCliPath() → npm 1.0.0            ✗            → native installer 2.1.220   ✓
```

So the user gets silently downgraded to whatever ancient npm copy is lying around, which is precisely #1516's report.

**#1543** ("`happy` falsely reports 'Claude Code is not installed'") may be the same root cause, in the case where the native install sits outside the standard prefixes that `findNativeInstallerCliPath()` probes, so nothing rescues the `null`. I have **not** verified that one, since I cannot reproduce that reporter's layout, and I am not claiming it.

Worth adding `Fixes #1516` to the description so it closes on merge. Also flagging for triage that #1394 (2026-06-13) is an earlier PR against the same root cause; it currently reports `CONFLICTING` while this one is `mergeable`, so this is likely the one to take, but the two authors deserve to know about each other.

One process note, unrelated to the code: this PR shows 0 check-runs because its workflow run is sitting at `action_required` (contributor approval gate). Same state as mine and `arozanov`'s. It has not failed, it has not started.

PR #1592, opened 2026-07-26
fix(app): keep Stop reachable while a selection card is pending

chphch:
While an `AskUserQuestion` card or a permission request is pending, `useSessionStatus` reports `permission_required` — a state the `showAbortButton` gate in `SessionView` covers on neither platform (web checks `thinking || waiting`, native checks `thinking`). So the Stop control disappears at exactly the moment the agent is parked inside the tool call and aborting is the only way to ignore the offered choices and keep typing: on mobile the composer's primary button falls back to `idle` (`resolveAgentInputPrimaryAction` needs `showAbortButton` to return `'stop'`), and on web the Escape-to-abort handler — gated on the same prop — stops responding. This adds `permission_required` to the gate on both platforms; native still hides Stop on an idle `waiting` session, so the behaviour introduced in "refresh mobile UI and improve session handling" is otherwise unchanged.

## Proof

Before/after on an Android emulator (Pixel 5, API 30, arm64), same session parked in `permission_required` on a pending `AskUserQuestion` card: the pre-fix release build offers only the voice button, the build with this commit shows Stop. GIF and capture details in the comment below.

`pnpm typecheck` and `pnpm exec vitest run` (65 files / 779 tests) pass in `packages/happy-app`.

### DIFF
--- a/packages/happy-app/sources/-session/SessionView.tsx
+++ b/packages/happy-app/sources/-session/SessionView.tsx
@@ -976,9 +976,15 @@ export function SessionViewLoaded({
             onMicPress={(embedded || isDisconnected) ? undefined : micButtonState.onMicPress}
             isMicActive={(embedded || isDisconnected) ? false : micButtonState.isMicActive}
             onAbort={isDisconnected || !rigCanAbort(session.metadata) ? undefined : handleAbort}
-            showAbortButton={rigCanAbort(session.metadata) && (Platform.OS === 'web'
-                ? sessionStatus.state === 'thinking' || sessionStatus.state === 'waiting'
-                : sessionStatus.state === 'thinking')}
+            showAbortButton={rigCanAbort(session.metadata) && (
+                sessionStatus.state === 'thinking'
+                // A pending selection (AskUserQuestion / permission request) parks the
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — Stop is reachable while an `AskUserQuestion` card is pending

Captured on an Android emulator (Pixel 5, API 30, arm64) against a self-hosted server, with a real `happy-cli` daemon session parked inside an `AskUserQuestion` tool call (`permission_required`, composer empty, so `resolveAgentInputPrimaryAction` decides the primary control).

before/after

- **BEFORE** — release build from before this commit: the session sits in `permission_required`, and the composer offers only the voice button. There is no way to interrupt the pending turn, so the offered choices cannot be ignored.
- **AFTER** — release build with this commit: identical session and identical pending card, and the primary control is Stop.

Same emulator, same session, same un-answered question card — only the build differs. (The two builds also differ in unrelated composer icons that landed on my fork in between; the control this PR changes is the primary button on the right.)

`pnpm typecheck` and `pnpm exec vitest run` (65 files / 779 tests) pass in `packages/happy-app`.

PR #1593, opened 2026-07-26
feat(app): attach images from the new-session composer

chphch:
The new-session screen already sends whatever attachments the draft store holds, but nothing on that screen could put them there — only `HomeDock` stages an image before routing here, so starting a session from `/new` had no way to include a picture with the first message. This adds the missing staging surface: an attach button in both composer rows, the shared `AgentInputAttachmentStrip` for staged previews, and web paste support. The clipboard listener moves out of `AgentInput` into a reusable `useWebImagePaste` hook so both composers share one implementation. Attachments stay in memory only, matching the existing comment in `useNewSessionDraft` — picker URIs are temporary, so nothing new is persisted. On native mobile the button mirrors the session composer's `+` bubble so the attach affordance is identical before and after the first message; the desktop row keeps the outline glyph that matches its own action row.

### Continues #1461

This is @jlixfeld's PR #1461 rebased onto current `main`, with authorship preserved on the commit. That PR has been CONFLICTING since late June and another fork user [asked for it downstream](https://github.com/slopus/happy/pull/1461#issuecomment-5077466875), so this carries it forward rather than duplicating it. **If @jlixfeld rebases #1461, close this one** — no ownership claimed.

Two things changed while rebasing:

- `main` has since added `attachments` to `useNewSessionDraft` and the send-on-spawn path, so those parts of #1461 are dropped as already-landed.
- #1461 also persisted attachments to MMKV via a new `attachmentDraft` parser. `main`'s store comments say picker URIs are intentionally *not* persisted, so that half is dropped too and this PR keeps the in-memory behaviour.

### Proof

`pnpm typecheck` clean; `pnpm exec vitest run` 778 passed (the one failure is `blob.test.ts`'s 1MB case timing out under parallel load — passes when run alone, untouched by this diff).

Running on the self-hosted fork build: the attach control appears on the `/new` composer, picking a file stages it in the strip, and sending spawns the session with the image as the first message.

### Note on overlap

#1582 touches the same composer's send button in `app/(app)/new/index.tsx`. The hunks are adjacent but independent — the two PRs merge cleanly against main in either order, and whichever lands second I'll rebase and resolve the trivial conflict the same day. No sequencing decision needed on your side.

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -43,6 +43,9 @@ import { resolveAbsolutePath } from '@/utils/pathUtils';
 import { formatPathRelativeToHome, formatLastSeen } from '@/utils/sessionUtils';
 import { useNavigateToSession } from '@/hooks/useNavigateToSession';
 import { useNewSessionDraft } from '@/hooks/useNewSessionDraft';
+import { useImagePicker } from '@/hooks/useImagePicker';
+import { useWebImagePaste } from '@/hooks/useWebImagePaste';
+import { AgentInputAttachmentStrip } from '@/components/AgentInputAttachmentStrip';
 import { useShallow } from 'zustand/react/shallow';
 import type { MultiTextInputHandle } from '@/components/MultiTextInput';
 import { Modal } from '@/modal';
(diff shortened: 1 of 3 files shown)


chphch:
## Proof — attaching an image to the first message

Captured locally: standalone `happy-server` (PGlite) + Expo web, authenticated via the `getDevWebQueryCredentials()` dev-token URL bypass, driven headlessly with Playwright against this branch.

new-session composer attach

Frame 1 — `Image Upload` toggled on in Settings → Features, so the attach control renders in the `/new` composer action row (it is absent with the flag off). Frame 2 — clicking it opens the file chooser; the picked file stages in `AgentInputAttachmentStrip` above the input, with the remove `×` affordance, and the attach glyph switches to its active tint.

0 console errors across the run. `pnpm typecheck` clean; `pnpm exec vitest run` 778 passed (the single failure is `blob.test.ts`'s 1MB case timing out under parallel load — it passes when run alone and this diff doesn't touch encryption).

The native-mobile branch of the same control renders the session composer's `+` bubble instead of the outline glyph; that path is `Platform.OS !== 'web'`-gated so it isn't exercised by this web capture.

charliezong18:
+1 on this. We've been running an equivalent adaptation downstream since Jul 25 with the same trade-offs — memory-only attachments (no MMKV persistence), and the already-landed draft-store half dropped — see my notes on #1461. Confirmed working in daily use on desktop web: attach button, preview strip, and paste all behave as expected with `expImageUpload` on. Happy to see this land instead of my carrying the patch; thanks @chphch for preserving @jlixfeld's authorship.

chphch:
@bra1nDump — surfacing this one, since it picked up independent validation while it sat.

@charliezong18 has been running an equivalent adaptation downstream since Jul 25, landing on the same trade-offs this PR makes — memory-only attachments (no MMKV persistence), and the draft-store half dropped because it already landed separately. They report the attach button, preview strip and paste all behaving in daily desktop-web use with `expImageUpload` on, and (their words, in the comment above) would rather this land than keep carrying their own patch.

Status: rebased onto current `main` today — mergeable, typecheck green. Authorship stays with @jlixfeld, who wrote the original commit; I'm a co-author on it.

No urgency from my side. Flagging it only because a second user independently arriving at the same implementation seemed worth putting in front of you, and I'm happy to rescope or split the diff if it's larger than you'd want to take in one go.

PR #1598, opened 2026-07-27
fix(app): restore the invisible settings/model/effort controls on Android

chphch:
## Problem

On Android the session composer's permission, model and effort controls do not paint: the action row shows the photo, mic and send buttons with an empty gap where the three controls belong, so the model and effort pickers are unreachable on a phone. The controls are still laid out and tappable, which makes it look like the row is simply missing options.

## Root cause

The Compose → React Native boundary. Each trigger is a React Native subtree handed to a Jetpack Compose `DropdownMenu.Trigger`, and expo-modules-core wraps such a child in `ExpoComposeAndroidView`, which renders it with `Modifier.size(view.width, view.height)` sampled once at composition:

```kotlin
// expo-modules-core/android/src/compose/.../ExpoComposeAndroidView.kt
AndroidView(
  factory = { view },
  modifier = Modifier.size(
    view.width.toFloat().pxToDp().dp,
    view.height.toFloat().pxToDp().dp
  )
)
```

There is no layout listener, so a trigger composed before React Native has measured it stays pinned at 0x0 for the rest of its life.

## Fix

`AgentInput` already ships popup pickers for the non-native path that render the same three controls, so Android now uses those and iOS keeps its native menus.

Two smaller fixes in the same area:

- `DropdownMenuItem.Text` is a native slot view (`SlotView`), so the bare string child threw `Text strings must be rendered within a <Text> component` once per option on every menu render — nine per render on the rig preview. Compose's own `Text` turns children into the native `text` prop instead. Same fix in `NativeOptionsPicker.android`, which has the identical slot usage.
- The composer passed `NativeSettingsMenu` the full button style for both the wrapper and the child inside it, so padding applied twice once the controls became visible. The wrapper now carries only the slot's size, matching how `HomeDock` already splits it.

## Proof

Verified on an Android emulator (Pixel 5, API 30, arm64) against `happy://dev/rig-preview`: before the change the row renders only the send control; after it, the gear and the `GPT Shared · high` model/effort control render and tapping the latter opens the picker. Before/after screenshots are in the comments below.

### DIFF
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -696,8 +696,14 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
     // existing composer affordances rather than inheriting it.
     const runningOnMac = isRunningOnMac();
     const compactMobileComposer = Platform.OS !== 'web' && !runningOnMac && screenWidth <= 700;
-    const useNativeSettingsMenus = compactMobileComposer
-        || shouldUseExpoNativeSettingsMenu(Platform.OS, runningOnMac);
+    // iOS only. On Android the settings/model/effort triggers are React Native
+    // subtrees hosted inside a Jetpack Compose DropdownMenu, and expo-modules-core
+    // pins such a child to `Modifier.size(view.width, view.height)` sampled once at
+    // composition with no layout listener (ExpoComposeAndroidView) — composed before
+    // React Native measures it, the trigger stays 0x0 and the control is invisible
+    // while still occupying its slot. The composer's own popup pickers below render
+    // identically and work, so Android uses those instead of the native menu.
+    const useNativeSettingsMenus = shouldUseExpoNativeSettingsMenu(Platform.OS, runningOnMac);
     const activeSendIconColor = compactMobileComposer ? theme.colors.text : theme.colors.button.primary.tint;
     const isSendBlocked = props.blockSend ?? false;
 
--- a/packages/happy-app/sources/components/NativeOptionsPicker.android.tsx
+++ b/packages/happy-app/sources/components/NativeOptionsPicker.android.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { DropdownMenu, DropdownMenuItem } from '@expo/ui/jetpack-compose';
+import { DropdownMenu, DropdownMenuItem, Text as ComposeText } from '@expo/ui/jetpack-compose';
 import type { NativeOptionsPickerProps } from './NativeOptionsPicker';
 
 export function NativeOptionsPicker({
@@ -13,8 +13,12 @@ export function NativeOptionsPicker({
             <DropdownMenu.Items>
                 {options.map((option) => (
                     <DropdownMenuItem key={option.key} onClick={() => onSelect(option.key)}>
+                        {/* Native slot view: a bare string child throws "Text strings
+                            must be rendered within a <Text> component" mid-render. */}
                         <DropdownMenuItem.Text>
-                            {option.key === selectedKey ? `✓ ${option.label}` : option.label}
+                            <ComposeText>
+                                {option.key === selectedKey ? `✓ ${option.label}` : option.label}
+                            </ComposeText>
                         </DropdownMenuItem.Text>
                     </DropdownMenuItem>
                 ))}
--- a/packages/happy-app/sources/components/NativeSettingsMenu.android.tsx
+++ b/packages/happy-app/sources/components/NativeSettingsMenu.android.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { DropdownMenu, DropdownMenuItem } from '@expo/ui/jetpack-compose';
+import { DropdownMenu, DropdownMenuItem, Text as ComposeText } from '@expo/ui/jetpack-compose';
 import { View } from 'react-native';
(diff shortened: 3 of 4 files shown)


chphch:
Device capture — Android emulator (Pixel 5, API 30, arm64, 393dp wide so `compactMobileComposer` is on), driven against `happy://dev/rig-preview`, which mounts a real `AgentInput` with permission modes, models and effort levels wired up.

**Before** — the action row paints only the send/stop control; the gear and the model/effort control occupy their slots but render nothing:

before

**After** — the gear and the `GPT Shared · high` model/effort control render:

after

**After, tapped** — the model picker opens:

after, picker open

One caveat on the "before" frame so it is not overstated: it was captured on the same debug build with an earlier, ineffective attempt at the fix in place (wrapping the trigger in `RNHostView`), not on a separately built pristine `main` APK. The trigger still rendered nothing, so the frame shows the same behaviour `main` produces, but it is not a matched-pair capture from two builds. The "after" frames are the shipped patch.

For what it is worth, `RNHostView` — the expo-ui component whose whole purpose is this Compose → RN boundary — did not rescue it either: with `Host` + `RNHostView` the controls painted once and then the entire row, send button included, collapsed after the first tap. That is why this patch routes Android to the composer's existing popup pickers rather than trying to keep the Compose menu.

chphch:
### Re-captured on top of #1604

The frames in the PR body were taken 07-27, before #1604 reworked this composer, so here is a fresh device pair against `main` as it stands today.

session composer controls, before/after on post-#1604 main

Android emulator (Pixel 5 profile, API 30 arm64 — 393dp wide, so the compact mobile composer is active), light theme, `dev/rig-preview` (one of the two places that render `<AgentInput>`). Red box marks the action row that holds the gear and the model/effort control.

`BEFORE` and `AFTER` are two **release APKs** differing by exactly this PR's diff: `AFTER` is the shipped `2059` build, `BEFORE` is that same commit with only this fix reverted (throwaway `versionCode` 90005, same release keystore, never published to the update channel).

In `BEFORE` the row is empty — the gear, model and effort triggers are laid out and tappable but paint nothing, which is the `ExpoComposeAndroidView` 0×0 behaviour this PR describes. In `AFTER` the gear and `GPT Shared · high` render. #1604's rework does not touch this path, so the bug is unchanged on current `main`.

<sub>Methodology note: each APK was captured after `pm clear`, because the emulator *does* fetch and apply the self-host OTA channel — without clearing, both builds silently run the same OTA bundle and the comparison is meaningless.</sub>

mysteriousHerb:
the bug remains still on the android device.

chphch:
Rebased onto current `main` — mergeable again, typecheck green, `happy-app` suite 909/909.

Worth flagging what changed, since it is not a plain replay. Upstream reworked the mobile composer's native-menu sizing while this sat (`resolveMobileComposerMenuGeometry`, landing in 2c335c48 and refined through 301bce52 / 1072d782). The patch is now expressed on top of that geometry rather than the ad-hoc styles it originally replaced, so the diff is **smaller than before** — part of what this PR used to carry is now upstream's own.

What the rework does not change is the root cause described above: `ExpoComposeAndroidView` still samples the child's size once at composition with no layout listener, so a trigger composed before React Native has measured it stays pinned at 0x0. This PR's fix — route Android to the composer's existing popup pickers, keep native menus on iOS — is unaffected by the new geometry.

Also folded in, and worth calling out because it belongs here rather than anywhere else: the `@expo/ui/jetpack-compose` test mock now exports `Text`. This change moves `DropdownMenuItem`'s label to Compose's `Text` so the string stops being a bare slot child, which means `nativeSettingsMenu.android.test.ts` needs the mock to provide `Text`. I had been carrying that one line separately in my fork, where it kept getting stranded across rebases; it lives in this commit now so the two travel together.

The before/after screenshots further down were captured against the pre-rework composer, so they show the older layout. The behaviour they demonstrate — an action row that lays out and accepts taps while painting nothing — is what this still addresses.

chphch:
@bra1nDump — this is the other half of #1582, which you merged on Aug 3.

Same origin: 1677fadc ("feat: refresh mobile UI and improve session handling") reworked `AgentInput.tsx`, `HomeDock.tsx` and `NativeSettingsMenu.android.tsx` in one pass. #1582 restored the send button on that row; this restores the gear and the model/effort control sitting beside it, which are still invisible on Android — laid out and tappable, painting nothing, so a phone user has no way to change model or effort.

Three things that were not true when this was opened:

- **It is mergeable and green again.** Rebased onto current `main`, expressed on top of the `resolveMobileComposerMenuGeometry` rework rather than the styles it originally replaced. The diff is smaller than it was, because part of what it used to carry is now upstream's own.
- **It has real device use behind it now, not just the emulator captures in the body.** I have been running this on my fork's Android build since late July; the gear and the model/effort control render and the pickers open.
- **The bug is still being reported on stock builds.** See the Aug 7 comment above — that is after 2c335c48 and 301bce52 landed, so the native-menu rework did not close it. The root cause is upstream of the styling: `ExpoComposeAndroidView` samples the child's size once at composition with no layout listener, so a trigger composed before React Native measures it stays pinned at 0x0.

#1601 is the home-dock instance of the same root cause, also mergeable and green.

Happy to split this, narrow the scope, or re-capture before/after proof against the current composer if any of that makes it easier to take.

bra1nDump:
hey @chphch thank you for the detailed prs!

and ouch we shipped some very broken code to android it seems.

taking a closer look and will merge within the next day

PR #1600, opened 2026-07-27
fix(app): follow the selected agent in the home dock placeholder

chphch:
## Problem

The session list's dock composer hardcodes `Ask Codex` as the placeholder of its focused state:

https://github.com/slopus/happy/blob/3c0155c2/packages/happy-app/sources/components/HomeDock.tsx#L1049

It names Codex regardless of which agent the new-session draft has selected, so anyone whose default is Claude Code (or Gemini, OpenClaw, Agy) is told they are about to talk to a different agent than the one that will actually start. The dock right below it shows the real agent, so the two disagree on the same screen.

## Fix

Derive the placeholder from the selected agent, the same way the first-message composer already does:

https://github.com/slopus/happy/blob/3c0155c2/packages/happy-app/sources/app/(app)/new/index.tsx#L1879

```diff
-                            placeholder="Ask Codex"
+                            placeholder={`Ask ${currentAgent.name}`}
```

`currentAgent` is already computed in this component for the model button's fallback label, so there is no new state or lookup.

Left the string hardcoded rather than routing it through `t(...)`, to match its neighbours in this component (`Plan, ask, build…`, `Machine`, `Project`, `Worktree`, `No worktree`) — happy to switch it to a translation key if you would rather fix all of them at once.

## Verification

`pnpm typecheck` clean on this branch. Rendered the dock at phone width and switched the agent between Claude Code and Codex — the placeholder follows the selection (`Ask Claude Code` ↔ `Ask Codex`).

Before/after emulator screenshots are in the comments below.

### DIFF
--- a/packages/happy-app/sources/components/HomeDock.tsx
+++ b/packages/happy-app/sources/components/HomeDock.tsx
@@ -1046,7 +1046,7 @@ export const HomeDock = React.memo(({
                             value={prompt}
                             onChangeText={onPromptChange}
                             onFocus={() => setIsFocused(true)}
-                            placeholder="Ask Codex"
+                            placeholder={`Ask ${currentAgent.name}`}
                             placeholderTextColor={theme.colors.textSecondary}
                             selectionColor={theme.colors.text}
                             autoCorrect


chphch:
Native before/after, captured on an Android emulator (Pixel-5-class, arm64 debug build). One debug APK built from upstream/main; the JS bundle comes from Metro pointed at each branch's worktree — the diff is JS-only, so only Metro's worktree differs between the two frames. App runs against a throwaway standalone `happy-server`.

**Before** (upstream/main): the focused home-dock composer's placeholder reads **"Ask Codex"** while the actually-selected agent is the default **Claude Code** (`useNewSessionDraft.ts` defaults to `'claude'`) — the placeholder doesn't follow the selection:

before

**After** (this branch): identical pristine state — the placeholder reads **"Ask Claude Code"**, following the selected agent:

after

One honest caveat: I couldn't demonstrate a *non-default* selection in this build — the `@expo/ui/jetpack-compose` dropdown triggers don't paint or respond in the Android debug client (the same Compose-boundary breakage #1598/#1601 address), so the pair shows the default-agent mismatch rather than a picker round-trip.

chphch:
Closing — fixed upstream by ce636c5a9.

The placeholder is no longer the hardcoded `Ask Codex`: `HomeDock.tsx:708` now derives it through `resolveHomeDockPromptPlaceholder(currentAgent.key, currentAgent.name)`, and grepping `main` for `Ask Codex` returns nothing.

Upstream's version is better than the one-liner here — it special-cases the legacy agent keys, so `claude` reads "Ask Claude Code" rather than whatever `AGENTS` happens to name it, and it ships with a test. No residual gap.

PR #1601, opened 2026-07-27
fix(app): restore the home dock's agent and config pickers on Android

chphch:
## Problem

On Android, tapping the session list's dock composer opens focus mode with **nothing but a send button** — no machine / project / worktree rows, no gear, no model or effort control. Every one of the dock's six pickers is unreachable, so a new session can only be started with whatever config the draft already had.

All six are React Native subtrees handed straight to a Jetpack Compose `DropdownMenu.Trigger`:

- `NativeOptionsPicker.android` for the machine / project / worktree rows
- `NativeSettingsMenu.android` for the gear, model and effort controls

`expo-modules-core` wraps such a child in `ExpoComposeAndroidView`, which pins it to `Modifier.size(view.width, view.height)` — read **once at composition** with no layout listener. A child composed before React Native has measured it therefore stays `0x0` forever: the control still occupies its slot and still accepts taps, but it paints nothing. That is why the row of controls looks empty rather than missing.

Same root cause as #1598 (the session composer's gear/model/effort), which this repo's `AgentInput` now sidesteps by gating the native menus to iOS and falling back to its own popup picker. `HomeDock` has no such fallback, so it stayed broken — and unlike `AgentInput` it also loses its three environment rows.

## Fix

Gate the native menus to iOS exactly as `AgentInput` does, and drive the same pickers from a sheet everywhere else:

- `useNativeMenus = Platform.OS === 'ios'` — iOS keeps its real native menus, untouched.
- Every trigger becomes a plain `Pressable` on other platforms, so it is measured and painted by React Native.
- The sheet reuses the option-list styles **this component already carried unused** (`settingsSurface`, `settingsHeader`, `backButton`, `settingsTitle`, `optionList`, `option*`) — they appear to be left over from a pre-native-menu picker, so no new design was invented.
- It renders **inside the focus `Modal`**, sharing that window. An app-level modal would risk landing behind the RN `Modal` on Android.
- The gear opens the agent + permission list (same grouping as the iOS gear menu) and drills into either; model, effort and each environment row open their option list directly. Selecting applies and returns to the dock; the backdrop closes the sheet before it closes focus mode; leaving focus mode resets it.

Web is also affected, though invisibly today: `NativeSettingsMenu.web` is a no-op passthrough that renders the trigger with no menu at all, so those controls were inert there. The same sheet now serves web.

## Verification

`pnpm typecheck` clean, `pnpm vitest run` 779 passed.

Rendered the non-iOS path at 393dp and exercised every control: the three environment rows are visible, the gear opens `AGENT` / `PERMISSION MODE`, the agent list shows all agents with the selected one checked, `MODEL` and `EFFORT` open their own lists (with descriptions), and picking an agent applies it and updates both the placeholder and the model label.

Since the dock is native-only in `MainView`, that render was driven by temporarily mounting `HomeDock` in the dev rig screen — so this exercises the **same JS code path Android takes** (`Platform.OS !== 'ios'`), not the Android view layer itself. The Compose boundary that caused the bug is no longer on that path, which is the point of the fix, but I have not captured it on a device.

### DIFF
--- a/packages/happy-app/sources/components/HomeDock.tsx
+++ b/packages/happy-app/sources/components/HomeDock.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ActivityIndicator, Keyboard, LayoutChangeEvent, Modal as RNModal, Platform, Pressable, Text, TextInput, View } from 'react-native';
+import { ActivityIndicator, Keyboard, LayoutChangeEvent, Modal as RNModal, Platform, Pressable, ScrollView, Text, TextInput, View } from 'react-native';
 import { Ionicons } from '@expo/vector-icons';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { StyleSheet, useUnistyles } from 'react-native-unistyles';
@@ -17,7 +17,7 @@ import Animated, {
 import { MobileGlassSurface } from './MobileGlass';
 import { BubblePressable } from './BubblePressable';
 import { NativeOptionsPicker } from './NativeOptionsPicker';
-import { NativeSettingsMenu, type NativeSettingsMenuGroup } from './NativeSettingsMenu';
+import { NativeSettingsMenu, type NativeSettingsMenuGroup, type NativeSettingsMenuProps } from './NativeSettingsMenu';
 import { AgentInputAttachmentStrip } from './AgentInputAttachmentStrip';
 import { Typography } from '@/constants/Typography';
 import { layout } from './layout';
@@ -62,6 +62,7 @@ export const MOBILE_HOME_DOCK_CONTENT_INSET = 108;
 
 type EnvironmentSetting = 'machine' | 'project' | 'worktree';
 type AgentSetting = 'agent' | 'model' | 'permission' | 'effort';
+type PickerPage = EnvironmentSetting | AgentSetting;
 
 const CUSTOM_PROJECT_PATH_KEY = '__custom_project_path__';
 
@@ -456,6 +457,14 @@ export const HomeDock = React.memo(({
     const [isFocused, setIsFocused] = React.useState(false);
     const [focusModeVisible, setFocusModeVisible] = React.useState(false);
     const [focusedInputContentHeight, setFocusedInputContentHeight] = React.useState(0);
+    // Jetpack Compose measures a DropdownMenu.Trigger's React Native child once,
+    // at composition time, and never re-reads it — so on Android every native
+    // menu here composes before React Native has measured it and stays 0x0, i.e.
+    // an invisible control. Android drives the same pickers from the in-modal
+    // sheet below instead; iOS keeps its real native menus.
+    const useNativeMenus = Platform.OS === 'ios';
+    const [sheetPage, setSheetPage] = React.useState<PickerPage | null>(null);
+    const [sheetRootVisible, setSheetRootVisible] = React.useState(false);
     const expImageUpload = useSetting('expImageUpload');
     const { selectedImages, pickImages, removeImage, clearImages } = useImagePicker();
     const agentType = useNewSessionDraft((state) => state.agentType);
@@ -783,6 +792,9 @@ export const HomeDock = React.memo(({
     const finishCloseFocusMode = React.useCallback(() => {
         setIsFocused(false);
         setFocusModeVisible(false);
+        // The sheet lives inside the focus modal, so it must not survive it.
+        setSheetPage(null);
+        setSheetRootVisible(false);
     }, []);
 
     const closeFocusMode = React.useCallback(() => {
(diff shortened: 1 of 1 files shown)


chphch:
## Device proof — Android emulator, matched pair

Captured on a Pixel 5 AVD (arm64, 1080×2340 @440dpi = **393dp**), API 30, `google_apis`. Both frames are the **same emulator, same account, same screen**; the only difference is the installed APK.

Two **release** APKs were built, one per commit, and swapped on the one device (same release keystore, so `adb install -r -d` preserves app data — the account and composer state carry across the swap). Neither was published: throwaway `versionCode` 90003/90004, no GH release, no OTA channel touched.

| | commit | versionCode |
|---|---|---|
| BEFORE | `5798705b` (this PR's parent) | 90003 |
| AFTER | `03a49651` (this PR) | 90004 |

### Tapping the dock composer

before/after

BEFORE: the config area above the composer is empty and the action row holds only `+` and send — the machine/project/worktree rows, the gear, and the model/effort control are all present in the tree and tappable, but paint nothing.

AFTER: `Select machine` / `~` / `No worktree` render with their icons, and the action row has the gear and `opus 4.8 · auto`.

### The sheet the fix adds

sheet

The gear opens agent + permission, drilling in shows the agent list with the current one checked, and the back chevron returns to the list. Model, effort and each environment row open their own option list the same way.

Note the placeholder reads `Ask Claude Code` in **both** panels — that is #1600, which is already in the parent commit here, so it stays constant across the pair and this comparison isolates exactly this PR's change.

The throwaway account created for this run has been deleted from the server.

chphch:
### Now captured on a device

The PR body says the non-iOS path was exercised by mounting `HomeDock` in the dev rig, and that I had **not** captured it on a device. That gap is now closed — this is the real Android view layer, on `main` as it stands after #1604.

home dock focus mode, before/after on a device

Android emulator (Pixel 5 profile, API 30 arm64 — 393dp wide), light theme, tapping the session list's dock composer to open focus mode. Upper red box = the three environment rows, lower red box = the dock's control row.

`BEFORE` and `AFTER` are two **release APKs** differing by exactly this PR's diff: `AFTER` is the shipped `2059` build, `BEFORE` is that same commit with only this fix reverted (throwaway `versionCode` 90006, same release keystore, never published to the update channel).

`BEFORE` reproduces the report exactly — focus mode opens with **nothing but a send button**: no machine / project / worktree rows, no gear, no model or effort control, just `+` and the arrow. `AFTER` shows all six: `Select machine`, `~`, `No worktree`, plus the gear and `opus 4.8 · auto`.

So the Compose-boundary failure is confirmed on the actual Android view layer, not just inferred from the shared JS path — and the fix removes it there.

<sub>Methodology note: each APK was captured after `pm clear`, because the emulator *does* fetch and apply the self-host OTA channel — without clearing, both builds silently run the same OTA bundle and render identically. That is exactly what happened on the first attempt here.</sub>

chphch:
Rebased onto current `main` — mergeable again, typecheck green, `happy-app` suite 909/909.

Same note as on #1598, since this is the home-dock half of the identical root cause: upstream reworked the native-menu sizing while this sat (`resolveMobileComposerMenuGeometry`, 2c335c48 onward), and `HomeDock` now consumes that shared geometry. The fix here is re-expressed on top of it, which is why the diff is somewhat larger than the original — it follows upstream's frame/content split rather than carrying its own styles.

The underlying problem is unchanged: `ExpoComposeAndroidView` samples the child's size once at composition and never re-measures, so a trigger composed before React Native measures it stays at 0x0 for the rest of its life, leaving the agent and config pickers laid out and tappable but invisible.

chphch:
@bra1nDump — pairing this with #1598; they are the two remaining halves of the regression #1582 came from.

1677fadc reworked `AgentInput.tsx` and `HomeDock.tsx` together. You merged #1582 (send button) on Aug 3; #1598 covers the session composer's gear and model/effort controls, and this one covers the same controls in the home dock, where the agent and config pickers are laid out and tappable but paint nothing on Android.

Status: rebased onto current `main`, mergeable, typecheck green, `happy-app` suite 909/909. Now expressed on top of upstream's shared `resolveMobileComposerMenuGeometry` frame/content split rather than carrying its own styles. Running on my fork's Android build since late July.

Root cause is the same as #1598 and sits below the styling: `ExpoComposeAndroidView` samples the child's size once at composition and never re-measures, so anything composed before React Native measures it is stuck at 0x0 permanently.

Same offer as on #1598 — happy to split or rescope if that helps.

chphch:
Closing — this landed upstream as 85b42b229, with ce636c5a9 following up.

I tagged you on this yesterday asking for a look. That ask is moot now, and I'd rather retract it myself than leave it sitting in your queue.

85b42b229 fixes it by the same mechanism this PR proposed: gate the native menus off on Android (`shouldUseNativeHomeDockMenus` returning `platform !== 'android'`, `homeDockInteraction.ts:34-36`), make every trigger a plain `Pressable`, and drive the six pickers from a sheet inside the focus modal (`renderSettingsSheet`, `HomeDock.tsx:1217`). It also covers three things this PR did not — disabled-option handling (`isHomeDockOptionSelectable`), Android hardware-Back unwinding (`resolveHomeDockPickerBackAction`), and unit tests. ce636c5a9 then restored the worktree row on Android, which 85b42b229 had excluded.

The one behavioural difference left is the shape of the gate: this PR used `Platform.OS === 'ios'`, so web also got the sheet, where upstream uses `platform !== 'android'` and web keeps the native menus. That is unobservable — `HomeDock` is mounted only in the non-web branch of `MainView.tsx:426-441`, so the dock never renders on web at all. Nothing here worth reproposing.

Thanks for fixing it.

Issue #1450, opened 2026-06-26
New sessions don't persist permission mode, model, or effort defaults (consolidates #648, #625)


ay18:
## Summary

Per-session preferences — **permission mode (YOLO), model, and reasoning effort** — are not persisted as **defaults for new sessions**. Every new session reverts to baseline (permission: prompts unless YOLO is re-toggled; effort: `medium`; model resets), so a user who wants a consistent default (e.g. YOLO + `xhigh` + Opus) has to reconfigure it **every single time** they start a session.

This appears to be one underlying cause behind a few existing reports, plus a third field (effort) that isn't tracked yet:
- #648 — YOLO preference not persisted across new sessions
- #625 — model preference / permission mode reset to `default` for new sessions
- (new) **reasoning effort** resets to `medium` even when set to `xhigh`

Filing a consolidated issue because the fix is likely shared: persist a set of new-session defaults rather than fixing each field one at a time.

## Environment

- happy (CLI) **v1.1.10**
- Claude Code **2.1.178**
- macOS (darwin, arm64)
- Sessions originated **from the mobile app** (daemon-spawned).

## Reproduction

1. In the mobile app's **Start New Session** screen, set model = Opus, effort = `xhigh`, YOLO = on.
2. Start the session — it honors those settings.
3. Start **another** new session.
4. Effort is back to `medium`; YOLO is off again; model preference doesn't stick.

## Evidence

- The spawned Claude child process is launched with `--effort medium` even though `xhigh` was selected for prior sessions.
- The in-app YOLO toggle **does** work *within* a session on v1.1.10, but is not remembered for the next new session (matches #648).
- There is **no local lever** to set these defaults:
  - The daemon reads env vars like `HAPPY_SERVER_URL`, `HAPPY_HOME_DIR`, and reconnect/fork vars, but **none** for permission mode, model, or effort.
  - `~/.happy/settings.json` (schema v2) accepts `sandboxConfig` but has **no** `defaultPermissionMode` / `defaultModel` / `defaultEffort` field.
  - The code constant `DEFAULT_CLAUDE_PERMISSION_MODE = "yolo"` exists only as a fallback and is overridden by the per-session metadata the app sends.
- The daemon's new-session spawn builds args as roughly `claude --happy-starting-mode remote --started-by daemon` with **no** permission/effort/model flag (related to #1396), so the effective settings come solely from app/server-supplied session metadata — which is where the non-persistence lives.

## Request

[...]

Issue #648, opened 2026-02-16
YOLO preference not persisted across new sessions


omachala:
## Issue
When creating a new session, the YOLO (approve all) preference is not remembered from previous sessions. This requires manually setting YOLO mode every time a new session is started.

## Expected Behavior
YOLO preference should be persisted and applied automatically to new sessions, similar to how other user preferences are maintained.

## Current Behavior
Every new session defaults to non-YOLO mode, requiring manual intervention to enable it each time.

## Steps to Reproduce
1. Start a session and enable YOLO mode
2. Complete the session
3. Start a new session
4. YOLO mode is disabled, needs to be manually re-enabled

## Impact
This creates friction in the workflow, especially for users who prefer to run in YOLO mode by default and start multiple sessions throughout the day.

## Suggested Solution
- Add YOLO preference to user settings/profile
- Auto-apply saved YOLO preference when spawning new sessions
- Allow per-session override if needed

ashavolian:
seconding this. every time i create a new session i have to send a message, abort, change permissions, then send another message. very annoying. there should definitely be a default permission setting so i can start a new session with yolo-opus everytime

zc402:
I also need this feature.

bra1nDump:
Shipped: Settings > Agents > Agent Defaults sets the permission mode new sessions start with, and per-session override already exists. Note it's an explicit setting, flipping YOLO inside a running session doesn't silently become the default. Covers what you, @ashavolian, and @zc402 asked for; broader defaults work is tracked in #1450. Can you update/reload and verify? Thanks for requesting this!

PR #1606, opened 2026-07-29
fix(app): stop an empty settings queue from resetting agent defaults

chphch:
**Two other open PRs already target this bug, and I'd rather one of them land than this one.** #1477 (@StefanGilligan) reaches the same root cause independently:

> "`partial()` makes fields optional but doesn't strip their Zod `.default()`, so the parse **re-injects** defaulted fields (`schemaVersion`, `agentDefaultOverrides`, `dismissedCLIWarnings`) that were never pending."

and #1395 (@jlixfeld) reports the symptom:

> "Configurable Agent Defaults … revert on app restart. The user sets a default, it appears to persist while the app is open, but a restart reverts it."

If either of those merges, I'll close this. Flagging the overlap up front so it isn't three PRs competing for one review.

**Reported by 6 users** — issues #648 (open), #625, #1450 (open):

> "Every new session reverts to baseline … so a user who wants a consistent default (e.g. YOLO + `xhigh` + Opus) has to reconfigure it **every single time**"
> — ay18, #1450, 2026-06-26

<details>
<summary>All reports</summary>

|---|---|---|---|
| omachala | 2026-02-16 | #648 (open) | "the YOLO (approve all) preference is not remembered from previous sessions" |
| ashavolian | 2026-03-24 | #648 | "every time i create a new session i have to send a message, abort, change permissions, then send another message. very annoying" |
| ashavolian | 2026-03-25 | #625 | "both the permissions setting and model preference get reset to defaults even though i had already configured them" |
| MrPhobos | 2026-04-20 | #625 | "+1, still reproducing on Happy CLI v1.1.6" |
| ay18 | 2026-06-26 | #1450 (open) | see above |
| ay18 | 2026-07-07 | #1450 | "agent defaults shows effort xhigh but every spawned session still runs at medium" |

</details>

---

## The bug

Pick a default in **Settings → Agent Defaults** (model, effort or permission). Reload the app. The pick is gone.

It is not a UI glitch — the value is destroyed on disk *and on the server*, so it also disappears from every other device on the account.

## Root cause

`loadPendingSettings()` parses the persisted outbound queue with `SettingsSchema.partial()`:

```ts
const parsed = JSON.parse(pending);
return SettingsSchema.partial().parse(parsed);
```

`.partial()` makes each key optional but leaves the `.default()` wrappers intact, and three fields carry one — `schemaVersion`, `agentDefaultOverrides` and `dismissedCLIWarnings`. So parsing the **empty** queue `{}` (the normal state right after a successful sync) does not return `{}`; it returns:

```js
{ schemaVersion: 2, agentDefaultOverrides: {}, dismissedCLIWarnings: { perMachine: {}, global: {} } }
```

"nothing is pending" has become "reset these three fields".

On the next app start `syncSettings()` layers that phantom delta over the real settings and POSTs the result to `/v1/account/settings`, so the reset is written locally **and** pushed to the server. Instrumented run, immediately after a reload:

```
POST pending={"schemaVersion":2,"agentDefaultOverrides":{},"dismissedCLIWarnings":{…}}
     storageOverrides={"claude":{"modelMode":"claude-opus-5"}}   ← the real value is still on disk
     payloadOverrides=undefined                                  ← but it is stripped from the payload
```

Two consequences beyond the visible one:

- `dismissedCLIWarnings` is wiped the same way, so dismissed CLI warnings come back.
- Every cold start issues a settings write nobody asked for, so `settingsVersion` climbs by one per app launch forever (1370 on my account).

## The fix

Keep only the keys that were actually queued. Everything else in the parse result is a zod-injected default and must not travel as a pending change.

A deliberate reset still round-trips: `savePendingSettings({ agentDefaultOverrides: {} })` writes that key into the JSON, so it survives the filter.

## Verification

Local harness (standalone server + Expo web, fresh account), same build, only `persistence.ts` swapped between the two runs:

```
before: picked=[Model → opus 5]  afterReload=[Model → Default (opus 4.8)]   ← upstream/main
after : picked=[Model → opus 5]  afterReload=[Model → opus 5]               ← this PR
```

The phantom `POST /v1/account/settings` at startup disappears entirely with the fix.

- `pnpm typecheck` clean
- `vitest` 785 passing, including 5 new cases in `sources/sync/persistence.spec.ts` covering the empty queue, a real queued value, a deliberate empty-overrides reset, and malformed JSON

### DIFF
--- a/packages/happy-app/sources/sync/persistence.spec.ts
+++ b/packages/happy-app/sources/sync/persistence.spec.ts
@@ -1,0 +1,51 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+// react-native-mmkv needs a native/web backend; back it with a plain in-memory map.
+const store = new Map<string, string>();
+vi.mock('react-native-mmkv', () => ({
+    MMKV: class {
+        getString(key: string) { return store.get(key); }
+        set(key: string, value: string) { store.set(key, value); }
+        delete(key: string) { store.delete(key); }
+        getNumber(key: string) { const v = store.get(key); return v === undefined ? undefined : Number(v); }
+        clearAll() { store.clear(); }
+    },
+}));
+
+const { loadPendingSettings, savePendingSettings } = await import('./persistence');
+
+describe('loadPendingSettings', () => {
+    beforeEach(() => store.clear());
+
+    it('returns nothing pending for an empty queue', () => {
+        // Regression: `SettingsSchema.partial().parse({})` re-injects the fields that
+        // carry a zod `.default()`, so an empty queue used to come back as
+        // "reset schemaVersion / agentDefaultOverrides / dismissedCLIWarnings".
+        // That delta then overwrote real settings locally and on the server, which
+        // reset the agent model default on every refresh.
+        savePendingSettings({});
+        expect(loadPendingSettings()).toEqual({});
+    });
+
+    it('keeps only the keys that were actually queued', () => {
+        savePendingSettings({ viewInline: true });
+        expect(loadPendingSettings()).toEqual({ viewInline: true });
+    });
+
+    it('preserves a genuinely queued agentDefaultOverrides value', () => {
+        savePendingSettings({ agentDefaultOverrides: { claude: { modelMode: 'claude-opus-5' } } });
+        expect(loadPendingSettings()).toEqual({
+            agentDefaultOverrides: { claude: { modelMode: 'claude-opus-5' } },
+        });
+    });
+
+    it('keeps a deliberate reset to empty overrides', () => {
+        savePendingSettings({ agentDefaultOverrides: {} });
+        expect(loadPendingSettings()).toEqual({ agentDefaultOverrides: {} });
+    });
+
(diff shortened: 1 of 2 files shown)


chphch:
### Proof — matched pair, one build

Both panels are the **same screen after a reload**, captured from one running bundle with only `sources/sync/persistence.ts` swapped between the two runs (upstream/main version vs this PR's). Same fresh account, same standalone server, same click path: Settings → Agent Defaults → Claude Code → Model → `opus 5`.

agent default reset before/after

Left: the pick is gone and `Clear Overrides` has gone back to disabled — the override was destroyed, not just mis-rendered. Right: it survives.

Console/network trace for the same two runs:

```
before: picked=[Model → opus 5]  afterReload=[Model → Default (opus 4.8)]
after : picked=[Model → opus 5]  afterReload=[Model → opus 5]
```

On `upstream/main` the reload fires a `POST /v1/account/settings` with `expectedVersion=1` that carries `agentDefaultOverrides: {}` even though the real value is still in local storage. With the fix that POST is not issued at all, because the pending queue is genuinely empty.

PR #1608, opened 2026-07-29
fix(app): let the mobile new-session composer take newlines again

chphch:
On native mobile the new-session composer opted out of `MultiTextInput`'s multiline defaults — `multiline={false}`, `returnKeyType="done"`, `submitBehavior="blurAndSubmit"` and a one-line `maxHeight` — so the keyboard offered a Done key that dismissed the field instead of typing a line break, which made the first message of a session the only one that could not contain a newline (the in-session composer, `AgentInput`, has always been multiline). This drops those four overrides so the shared defaults apply again, and removes `flex: 1` from `mobileInputField` so the row can size to its content: inside that auto-height column the shorthand resolves to a zero flex-basis with no free space to grow into, which pinned the field to its 44pt `minHeight` and clipped the text to two lines even once multiline was back. The field now grows to the same 120pt cap as the in-session composer and scrolls past it, and an empty composer keeps exactly the height it has today.

## Proof

Captured on an Android emulator (Pixel 5, API 30, arm64) running the app from Metro, since this composer is gated to native — `isDesktop = Platform.OS === 'web' || isRunningOnMac()`, so no browser width reaches it. Both panels are the same screen in the same state, with only the JS bundle swapped.

### DIFF
--- a/packages/happy-app/sources/app/(app)/new/index.tsx
+++ b/packages/happy-app/sources/app/(app)/new/index.tsx
@@ -124,6 +124,9 @@ const COMPOSER_INPUT_VERTICAL_PADDING = Platform.OS === 'web' ? 10 : 8;
 // Taller composer on web/desktop where vertical space is plentiful; keep the
 // compact cap on native mobile so the input doesn't dominate the screen.
 const COMPOSER_INPUT_MAX_HEIGHT = Platform.OS === 'web' ? 480 : 240;
+// The compact (native mobile) composer grows to the same cap as the in-session
+// composer — see AgentInput's `maxHeight` — instead of a single fixed line.
+const COMPACT_COMPOSER_INPUT_MAX_HEIGHT = 120;
 const COMPOSER_SEND_BUTTON_SIZE = 32;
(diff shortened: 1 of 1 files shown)


chphch:
## Proof — the first message can hold a newline again

Captured on an Android emulator (Pixel 5, API 30, arm64) with the app running from Metro, because this composer never renders in a browser — `isDesktop = Platform.OS === 'web' || isRunningOnMac()`, so narrowing the viewport gets you the desktop composer, not this one. Each pair is the same screen in the same state with only the JS bundle swapped.

The keyboard's action key is the whole bug: `returnKeyType="done"` + `submitBehavior="blurAndSubmit"` turned it into a Done key that blurred the field.

keyboard action key, before and after

And the result of pressing it — on the left the field blurs and the text stays on one line; on the right two line breaks land and the field grows to fit them (it caps at the same 120pt as `AgentInput` and scrolls after that).

result of pressing the action key

`pnpm typecheck` is clean and `vitest run` passes 780/780 in `packages/happy-app`.

PR #1610, opened 2026-07-29
fix(app): make the mobile FAB's plus icon visible again

chphch:
The floating action button paints its icon with `fab.icon`, which the theme defines as the inverse of `fab.background` — white on black in the light theme, black on white in the dark one. That pairing only holds against the opaque fill, and the fill is now the web path only: on native the button's interior became a glass surface tinted like the page behind it, so the same token lands white-on-`rgba(255,255,255,0.84)` in light and black-on-`rgba(28,28,28,0.68)` in dark, and the plus is invisible in both themes. This points the native icon at `theme.colors.text` instead, which is what the glass composer already does for its own send arrow. The web path keeps `fab.icon` and its opaque fill unchanged.

## Proof

Captured on an Android emulator (Pixel 5, API 30, arm64), same device, same account, same screen (Artifacts) — the FAB's only render site. The change is a single colour value, so the frames differ only in the JS that produced them.

### DIFF
--- a/packages/happy-app/sources/components/FAB.tsx
+++ b/packages/happy-app/sources/components/FAB.tsx
@@ -69,7 +69,16 @@ export const FAB = React.memo(({ onPress }: { onPress: () => void }) => {
                     <Ionicons name="add" size={24} color={theme.colors.fab.icon} />
                 ) : (
                     <MobileGlassSurface interactive intensity={76} style={styles.glass}>
-                        <Ionicons name="add" size={24} color={theme.colors.fab.icon} />
+                        {/*
+                          `fab.icon` is the inverse of `fab.background` (white on
+                          black in light, black on white in dark), so it only reads
+                          against that opaque fill — which is the web path. Native
+                          replaces the fill with a glass surface tinted like the page
+                          behind it, where the same token lands white-on-near-white
+                          in light and black-on-near-black in dark. Follow the text
+                          color there instead, as the glass composer already does.
+                        */}
+                        <Ionicons name="add" size={24} color={theme.colors.text} />
                     </MobileGlassSurface>
                 )}
             </Pressable>


chphch:
## Proof — the plus is visible again in both themes

Android emulator (Pixel 5, API 30, arm64), same device and account throughout, on the Artifacts screen — the FAB's only render site (`app/(app)/artifacts/index.tsx`). The **before** frames are the emulator running the unpatched code; the **after** frames are a release build carrying this commit. Since the change is one colour value, that is the only difference between the columns.

FAB plus icon, before and after, light and dark

The dark row is the clearer read: `fab.icon` is `#000000` there, and the glass interior is `rgba(28, 28, 28, 0.68)`.

`pnpm typecheck` is clean and `vitest run` passes 780/780 in `packages/happy-app`.

chphch:
@bra1nDump — same invisible-control class as #1582 and #1598, both of which you merged: a native control that is laid out and tappable but paints nothing, so it reads as missing rather than broken.

The mechanism is #1582's exactly. `fab.icon` is defined as the inverse of `fab.background`, so it only reads against that opaque fill — and `FAB.tsx:26` now takes that fill on the web path only (`Platform.select({ web: fab.background, default: 'transparent' })`). Native renders the icon *inside* `MobileGlassSurface`, still coloured `fab.icon`, so the plus lands white-on-near-white in light and black-on-near-black in dark. Same shape as the composer's idle glass state discarding its computed colour, which @arozanov independently hit on a physical device in #1597.

Still live on current `main` — `FAB.tsx:26` against `:71`.

## Proof — full screen, so the control's place on the page is visible

FAB plus icon, before and after

Android emulator (Pixel 5, API 30, arm64), Artifacts screen — the FAB's only render site. Same account, same screen, same scale; the red circle marks the same coordinates in both frames. The two builds differ by exactly this PR's diff, and both frames were shot on the APK's embedded bundle (`1.7.0+chphch.*`, verified on the Settings version row) rather than an OTA.

The fix points the native icon at `theme.colors.text`, which is what the glass composer already does for its own send arrow; web keeps `fab.icon` and its opaque fill untouched. One file, +10/−1. Rebased onto current `main`: mergeable, CI green, `happy-app` suite 1066/1066.

Happy to narrow the scope if you would rather handle the glass tokens differently.

PR #1612, opened 2026-07-30
feat(cli,app): show when a session still has background work running

chphch:
Implements #1628.

A session drops to "waiting" the moment its turn ends, so a twenty-minute build, a background subagent or a workflow — the whole reason the session is still alive — is invisible from the session list. `thinking` cannot express this because it only describes the current turn, and from the list there is no way to tell "this finished" from "this handed off to something that is still running", which is exactly when you want to leave it alone. This adds a `background` session state fed by the background work Claude Code already reports.

## How

Every `Stop` hook payload carries `background_tasks` — one entry per in-flight item, `{ id, type, status, description, command? }`, where `type` is `shell`, `subagent`, `workflow`, `monitor`, or the raw discriminant for kinds added later. happy-cli already registers a `SessionStart` hook to track session ids, so this registers `Stop` and `SessionEnd` alongside it, folds the array into counts, and publishes them as `metadata.activity`.

Reading it from the hook rather than from the SDK stream is what makes it work in both launch modes: the `claude` binary runs its hooks whether happy-cli drives it through the PTY (`claudeLocal`) or through the SDK (`claudeRemote`).

`metadata.activity` is a shape the app already parses — `MetadataSchema` defines it and the session info screen already renders it for any client. The only thing missing was a producer; on the app side this un-gates the list-level indicators (they were restricted to Rig sessions) and adds the dot state:

- `background` — a lighter blue than `thinking`, so it reads as "still working, but not on your turn", with the existing activity summary spelling out what is running.
- `SessionEnd` clears the counts, so a session that exits mid-build leaves nothing stale behind.
- Identical summaries are not re-sent, so idle turns cost no metadata writes.

Server and wire format are untouched — `activity` rides inside the existing encrypted metadata.

## Proof

Captured live end-to-end: a real `happy-cli` daemon against a standalone `happy-server`, a real Claude session that starts a background shell, screenshotted after the turn ended. Before/after is the same session with the same shell still running — only the app bundle differs. GIF in the first comment.

happy-cli: `typecheck` clean, 804 unit tests pass (12 new, covering the summarizer's type routing, queued-vs-running split, unknown task kinds and the dedupe comparison). happy-app: `typecheck` clean, 780 tests pass. The hook pipeline itself was additionally exercised against a real `claude` process — `SessionStart` still routes as before, `Stop` delivers the running shell, and `SessionEnd` clears it.

## Relationship to #1587

#1587 addresses the same underlying annoyance from the other end: it defers `updateThinking(false)` in `claudeRemote.ts` until the SDK's background-task set empties, so the turn is not reported idle while a subagent runs. This PR does not touch that file and does not change when `thinking` flips — it adds the counts and a separate visual state for the idle-but-still-busy case, and covers the local PTY path too. They compose: with both, a subagent keeps the session `thinking`, and anything still running past turn end shows as `background`. Happy to rebase behind #1587 if it lands first.

### DIFF
--- a/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
+++ b/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
@@ -27,6 +27,8 @@ import { ProviderIcon } from './ProviderIcon';
 const STATUS_CONFIG: Record<SessionState, { color: string; dotColor: string; isPulsing: boolean; isConnected: boolean }> = {
     disconnected: { color: '#999', dotColor: '#999', isPulsing: false, isConnected: false },
     thinking: { color: '#007AFF', dotColor: '#007AFF', isPulsing: true, isConnected: true },
+    // Lighter blue than `thinking`: still working, but not on your turn.
+    background: { color: '#5AC8FA', dotColor: '#5AC8FA', isPulsing: true, isConnected: true },
     waiting: { color: '#34C759', dotColor: '#34C759', isPulsing: false, isConnected: true },
     permission_required: { color: '#FF9500', dotColor: '#FF9500', isPulsing: true, isConnected: true },
 };
@@ -281,7 +283,7 @@ const CompactSessionRow = React.memo(({ session, selected, showBorder }: { sessi
                     color={theme.colors.textSecondary}
                 />
             );
-        } else if (session.state === 'permission_required' || session.state === 'thinking') {
+        } else if (session.state === 'permission_required' || session.state === 'thinking' || session.state === 'background') {
             indicator = <StatusDot color={status.dotColor} isPulsing={status.isPulsing} />;
         } else if (session.state === 'waiting') {
             indicator = <StatusDot color={theme.colors.textSecondary} isPulsing={false} />;
--- a/packages/happy-app/sources/components/SessionsList.tsx
+++ b/packages/happy-app/sources/components/SessionsList.tsx
@@ -415,6 +415,8 @@ export function SessionsList({
 const STATUS_CONFIG: Record<SessionState, { color: string; dotColor: string; isPulsing: boolean; isConnected: boolean }> = {
     disconnected: { color: '#999', dotColor: '#999', isPulsing: false, isConnected: false },
     thinking: { color: '#007AFF', dotColor: '#007AFF', isPulsing: true, isConnected: true },
+    // Lighter blue than `thinking`: still working, but not on your turn.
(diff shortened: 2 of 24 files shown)


chphch:
## Proof — a session that handed off to a background shell no longer reads as idle

Captured live end-to-end: a real `happy-cli` daemon paired to a standalone `happy-server` (PGlite), a real Claude session told to start one background shell (`sleep 600`) and then stop. The screenshots are taken **after the turn ended** — the agent is idle and the shell is still running.

Before and after are the **same session with the same shell still running**; the only thing that changes between the two rows is the app bundle, so the difference is this diff and nothing else.

background work surfaced in the session status and list dot

The `1 processes` line is the existing activity summary — the app already knew how to render it, it just had no producer and was gated to Rig sessions. Worth noting the session **info** screen already rendered `metadata.activity` un-gated on `main` (`app/(app)/session/[id]/info.tsx`), so it lights up from the CLI half of this PR alone; the list dot and status line are the part this PR adds.

<details>
<summary>Hook pipeline verification (separate from the UI capture)</summary>

Ran against a real `claude` process with the generated settings file, asserting on what the hook server actually received:

```
PASS  SessionStart hook still routes to /hook/session-start — 1 post(s)
PASS  Stop hook reached /hook/stop — 2 post(s)
PASS  a Stop payload carried a running background task
      {"processes":{"running":1},"subagents":{"running":0,"queued":0,"total":0}, ...}
PASS  background shell counted as a process — processes.running=1
PASS  SessionEnd hook reached /hook/session-end — 1 post(s)
PASS  SessionEnd yields an empty summary (clears stale activity)
```

`packages/happy-cli`: typecheck clean, 804 unit tests pass (12 new). `packages/happy-app`: typecheck clean, 780 tests pass.
</details>

chphch:
Closing in favour of #1587 (@arozanov) — it solves this from a better signal than mine does.

We both key off the same underlying fact, that Claude Code reports its background work. But #1587 listens to `background_tasks_changed` on the SDK stream, whereas this PR reads `background_tasks` off the `Stop` hook payload. The stream carries the full running set every time it changes, so it also catches work that starts mid-turn; the `Stop` hook only fires when the turn ends, so mine can't. His signal can express everything mine can, and the reverse isn't true — and he does it in one file against my twenty-four.

His timeline is also the evidence I didn't have: fourteen minutes of the app showing a session as idle, with a "done" push sent, while a subagent was still working.

I've left #1628 open for the part #1587 doesn't cover — surfacing background work in the session list itself, not just holding back the idle report. If `background_tasks_changed` lands, that's the signal to build it on.

Issue #1595, opened 2026-07-27
Aborting a turn silently reverts the session model and effort to launch defaults


charliezong18:
Aborting a running turn resets the CLI's tracked model and reasoning effort back to the session launch defaults (`opus` / `medium`), while the app's picker still shows the user's selection — so every subsequent turn silently runs the wrong model with no indication in the UI. The app only sends `meta.model` / `meta.effort` when the user changes the picker, so the desync persists until the user happens to re-touch it.

Repro (mobile/web app + CLI session):

1. In an active session, pick a non-default model (e.g. Sonnet; launch default is opus).
2. Send a message — it correctly runs on the picked model.
3. Tap stop to abort a turn.
4. Send another message without touching the picker.
5. The turn runs on `opus` (`claude-opus-4-8`); the picker still shows the picked model. Same for effort (reverts to `medium`).

Evidence from an end-to-end run (real happy-server + real CLI, `HAPPY_HOME_DIR` sandbox):

```
[loop] Model updated from user message: claude-sonnet-4-6
[claudeRemote] Result received
[remote]: doAbort
[loop] Reset current mode defaults after abort
[loop] User message received with no model override, using current: opus
```

Turn 2's SDK subprocess was spawned with `--model opus` and its assistant stream reports `"model": "claude-opus-4-8"`.

Cause: `resetCurrentModeDefaults()` in `packages/happy-cli/src/claude/runClaude.ts` runs on abort and resets `currentModel`/`currentEffort` to `options.model ?? DEFAULT_CLAUDE_MODEL` — `options.model` is undefined for app-driven sessions, so it lands on the `'opus'` constant. `packages/happy-cli/src/codex/runCodex.ts` has the identical reset.

Related but distinct from #1450 (that one is about *new-session* defaults not persisting; this one is mid-session state being destroyed by abort).

PR with a fix incoming.

PR #1636, opened 2026-08-02
fix(app): stop the agent default from re-modeling a running session

chphch:
The per-agent defaults in Settings are read live on every send — `resolveMessageModeMeta()` falls back to `agentDefaultOverrides` whenever a session has no explicit pick of its own — so the default is not a *new-session* default but a standing remote control over every unconfigured session. Editing the default model in Settings therefore switches the model of sessions that are already running, on their very next message, with nothing in the chat to say so. The CLI already assumes the opposite: `resetCurrentModeDefaults()` in `runClaude.ts` / `runCodex.ts` deliberately keeps the model and effort across an abort on the stated grounds that *"the app sends them only when the user changes the picker"* (#1595) — today the app also sends them when the user changes an unrelated global setting. This keeps the default doing its job (seeding a session that has no pick of its own) but scopes it to the moment the session first uses it: `resolveAgentDefaultPin()` freezes those values onto the session on first use, and `handleAbort()` no longer erases the pick — it now clears only the permission mode, which is the one field both agents actually reset on abort.

## Proof

Live end-to-end run against a standalone `happy-server` (PGlite) + Expo web + a paired `happy-cli` daemon, driven headlessly with Playwright. Identical script on both builds: send turn 1, set **Settings → Agents → Claude Code → Model = sonnet 4.6**, send turn 2, change it to **fable 5**, send turn 3. The verdict is the CLI session log, which records the model each turn actually ran with.

**Before**

```
[loop] User message received with no model override, using current: opus
[loop] Model updated from user message: sonnet
[loop] Model updated from user message: fable      ← the running session followed the setting
```

**After**

```
[loop] User message received with no model override, using current: opus
[loop] Model updated from user message: sonnet
[loop] Model updated from user message: sonnet     ← the session keeps the model it was using
```

Turn 2 is identical on both builds — the default still seeds a session that has no pick, which is the behaviour `resolveMessageModeMeta`'s settings fallback was added for. Only turn 3 differs.

Reproducing this on `main` needs an agent default that survives long enough to send a message; on current `main` the override is dropped again before the next turn (the bug #1606 / #1395 / #1477 address), which masks this one. The run above therefore has #1606's persistence fix applied on top as a harness prerequisite — it is not part of this diff.

`packages/happy-app`: `pnpm typecheck` clean, `pnpm vitest run` 817/817 across 67 files, including six new cases covering `resolveAgentDefaultPin` (pins the fallback, leaves explicit picks alone, pins only missing fields, no-op without overrides, no-op for another agent, no-op for Rig sessions).

### DIFF
--- a/packages/happy-app/sources/-session/SessionView.tsx
+++ b/packages/happy-app/sources/-session/SessionView.tsx
@@ -849,8 +849,14 @@ export function SessionViewLoaded({
     }, [sessionId, expImageUpload, selectedImages, clearImages, pendingCommunications]);
 
     const handleAbort = React.useCallback(() => {
-        // Stop cancels only the active turn. Permission, model, and effort are
-        // session choices and must remain sticky for the next message.
+        // Stop cancels only the active turn, so model and effort stay sticky for
+        // the next message. The permission mode cannot: both agents reset it to
+        // the launch value on abort (runClaude.ts / runCodex.ts
+        // `resetCurrentModeDefaults`), so leaving the app's synced copy in place
+        // desyncs the picker from what the next turn actually runs (#1492/#1595).
+        if (!isRig) {
+            sessionSetAgentModes(sessionId, { permissionMode: null });
+        }
         sessionAbort(sessionId);
     }, [sessionId]);
 
--- a/packages/happy-app/sources/sync/messageMeta.test.ts
+++ b/packages/happy-app/sources/sync/messageMeta.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { resolveMessageModeMeta, UnsupportedPermissionModeError } from './messageMeta';
+import { resolveAgentDefaultPin, resolveMessageModeMeta, UnsupportedPermissionModeError } from './messageMeta';
 import { rigMetadataFixture } from './__testdata__/rigMetadata';
 
 describe('resolveMessageModeMeta', () => {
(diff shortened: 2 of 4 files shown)


chphch:
## Proof — the running session stops following the setting

The setting really was changed under the live session — this is Settings → Agents after step 3, with **fable 5** selected for Claude Code:

agent defaults set to fable 5

With that setting in place, the next message in the session already running on `sonnet` produces:

| turn | setting at send time | before | after |
|---|---|---|---|
| 1 | *(no override)* | `no model override, using current: opus` | `no model override, using current: opus` |
| 2 | sonnet 4.6 | `Model updated from user message: sonnet` | `Model updated from user message: sonnet` |
| 3 | fable 5 | `Model updated from user message: fable` | `Model updated from user message: sonnet` |

Same harness, same script, same account shape — only the app build differs. Turn 2 is unchanged by design: the default still seeds a session that never made a pick. Turn 3 is the regression this fixes.

<details>
<summary>How it was run</summary>

Standalone `happy-server` (PGlite) + Expo web + a paired `happy-cli` daemon on an isolated port set, authenticated via the `getDevWebQueryCredentials()` dev-token URL bypass and driven headlessly with Playwright. Between turns the script navigates to `/settings/agents` and back via SPA history, never a reload. Per-turn model comes from the CLI session log's `[loop]` lines.

`main` alone cannot host this repro: the agent default is wiped before the next message can be sent (#1606 / #1395 / #1477), so that persistence fix was applied on top of both builds as a harness prerequisite. It is not part of this diff.
</details>

Issue #1359, opened 2026-06-04
file-watcher loops on ENOENT indefinitely when ghost session_id ≠ Claude's actual session_id


internexio:
## Symptom

Happy's file-watcher logs the following ~1×/s indefinitely:

```
FILE_WATCHER Watch error: ENOENT: no such file or directory, watch '<project_dir>/<ghost_uuid>.jsonl', restarting watcher in a second
```

Retry counts climb into the hundreds of thousands over multi-day sessions (verified: 178,887 retries on a ~2-day session). User-visible effect: the in-pane Claude output never visually completes; new prompts intermittently surface "Process exited unexpectedly". Typing anything forces a state recheck that flushes the UI queue.

## Root cause (hypothesis, confidence ~0.9)

The Happy launcher generates a fresh session UUID at tmux-creation time and tells the file-watcher to bind to that path. The inner Claude process mints its own session_id (or uses `--resume <id>`) and writes its transcript to a *different* `.jsonl` in the same project dir. The two never reconcile, so the watcher's target file genuinely never exists.

## Repro

1. Start a Happy session, let it run long enough that Claude makes its first transcript write
2. `LATEST=$(ls -t ~/.happy/logs/*pid*.log | head -1)`
3. `grep -c 'FILE_WATCHER.*ENOENT.*restarting' "$LATEST"`
4. Compare: the watched path Happy chose vs the newest `.jsonl` actually written by Claude in the same project dir — they differ.

## Workaround we shipped

A 5-min launchd check that detects looping watchers (mtime <30s + ≥10 recent ENOENTs) and symlinks the ghost path to the real `.jsonl`:

```bash
ln -sfn "$REAL_ABSOLUTE_PATH" "$GHOST_PATH"
```

Key gotcha: target MUST be the absolute path. node's `fs.watch` resolves relative symlinks differently than `ln -s` does — a same-directory relative link leaves the watcher still throwing ENOENT.

## Proposed fix

After N consecutive ENOENT retries (suggest 10–20), re-resolve the watched session_id from Claude's current transcript_path (the newest `.jsonl` in the project dir is a reasonable heuristic) rather than the launcher-resolved one.

Happy to provide a log sample on request.

WhymustIhaveaname:
+1, same pattern on Linux (Ubuntu 24.04). Saw this after Happy ran continuously for 57 days:

[...]

Issue #1111, opened 2026-04-16
Terminal-started session (happy claude -w): web shows title but no messages


AlseinX:
### Environment
- Happy CLI v1.1.6
- Linux x64, Node v25.8.2
- Auth method: dataKey (Mobile App login)

### Reproduction Steps

1. Run `happy claude -w` from terminal inside a git repo (e.g. `/home/yhxiang/repos/vnext`)
2. Session starts successfully, worktree is created
3. Open Happy Coder web client — session appears and **title updates are visible**
4. **Messages are NOT displayed** — the web shows no conversation content

### Evidence from Logs

**Terminal session log** (`happy claude -w`, PID 1305963):
- Session created successfully: `cmo1i0xlbc2xkwm0u9fnct35n`
- Socket connected successfully
- Messages ARE encrypted and queued (`"t": "encrypted"` entries present)
- Title updates work (sent via `update-metadata` socket event)

**File watcher stuck in infinite error loop** (every 1 second):
```
[FILE_WATCHER] Watch error: ENOENT: no such file or directory, 
watch '/home/yhxiang/.claude/projects/-home-yhxiang-repos-vnext/fee85f6a-....jsonl', 
restarting watcher in a second
```
The transcript path uses project directory `-home-yhxiang-repos-vnext` but the session runs in a worktree directory `wild-gathering-lecun`. The file watcher never recovers.

**Comparison with daemon-spawned session** (PID 1289949):
- Daemon-spawned session has 749 `[SOCKET]` events and messages display correctly
- Terminal-started session has only 3 `[SOCKET]` events (all incoming updates, no outgoing messages)
- No `flushOutbox` or HTTP POST to `/v3/sessions/{id}/messages` is logged for terminal session

### Analysis

The file watcher's infinite ENOENT retry loop may be blocking or interfering with the `InvalidateSync` mechanism that drives the outbox flush. Messages are encrypted and queued in `pendingOutbox` but never sent to the server via HTTP POST.

The transcript path mismatch occurs because the `-w` flag creates a worktree, but the transcript path is derived from the original project directory, not the worktree's actual project directory.

### Expected Behavior
- Terminal-started sessions should send messages to the server the same way daemon-spawned sessions do
- The file watcher should use the correct transcript path for worktree sessions

### Impact
Users who start sessions from terminal with `happy claude -w` cannot see messages in the web client, making remote monitoring of terminal-started sessions impossible.

AlseinX:
## Root Cause Analysis

[...]

Issue #846, opened 2026-03-13
Happy sessions are visible on Web/iOS, but message content is blank; session API returns metadata = null


alexsongsg:
## Summary

On Windows, Happy sessions are visible in Happy Web / mobile, but opening the session shows no message content.

I verified that:

- the machine is authenticated
- the daemon is running
- sessions are created successfully
- messages are being sent successfully
- session `seq` increases after message injection

However, when reading the session back through Happy's session API, `metadata` becomes `null`, and the Web/mobile client shows a blank session body.

This appears to be a server-side session metadata / decrypt / render issue rather than a local terminal startup issue.

## Environment

- OS: Windows
- Host: `AlexSong`
- Machine ID: `9ac7cb08-7d7d-4296-8205-09fd8c968f4a`
- Happy CLI version: `0.13.0`

## Symptoms

- Session is visible in Happy Web and mobile
- Opening the session shows no text / blank body
- Push notifications are received
- Session exists and can be selected
- But the conversation content is not rendered

## What I verified locally

I verified all of the following on the machine:

- Happy auth is valid
- machine registration is valid
- daemon is running
- session creation succeeds
- websocket connects successfully
- Codex MCP connects successfully
- Windows-specific `os error 193` MCP bridge issue was fixed locally
- test messages were injected into live sessions
- session sequence increments after message injection

So the local Happy/Codex startup path is working.

### 1. Broken sessions come back with `metadata = null`

For an existing visible session:

- session id: `cmmoon17c2g1zxe14s7p8g6v7`

Reading it through the Happy API returns:

- `metadata = null`

### 2. Fresh test session can be created and message writes succeed

I created a fresh test session:

- session id: `cmmooxd074e85xe14igftofaf`

I then injected a plain text assistant message into it.

After that, the session reads back with:

- `seq = 2`

So the server accepted the writes.

But even then, reading the session again still returns:

- `metadataIsNull = true`

This strongly suggests the problem is not “message never sent”, but that the session metadata/content becomes unreadable/unrenderable when fetched back.

## Reproduction notes

1. Authenticate Happy on Windows
2. Start daemon
3. Start a session from desktop / terminal
4. Open the same account in Happy Web or mobile
5. Session is visible
6. Open it
7. Body is blank / no messages shown

## Additional notes

I also observed that:

- older affected sessions and freshly created test sessions both hit the same pattern
- session existence is fine
- message writes are accepted
- but returned session metadata becomes `null`

[...]

PR #1638, opened 2026-08-02
fix(cli): session never syncs if first prompt comes >60s after launch

### LINKED ISSUE #1538: File watcher permanently drops session sync if the first prompt comes >60s after launch (regression from ENOENT-loop fixes; refs #1111, #1359)
## Summary

When `happy claude` is started from the terminal and the user does **not** send their first prompt within ~60 seconds, the session's transcript watcher gives up permanently. The session still registers with the server (it appears in the Web/mobile session list, title updates work, RPC like git status works), but **no message content is ever uploaded** - the session body is blank forever, even though the local Claude session is healthy and the transcript file exists and is actively written.

This looks like a regression introduced by the fix for the infinite ENOENT retry loops reported in #1111 and #1359. The giveup that stopped those loops (`missingFileTimeoutMs`, default 60s) doesn't account for the normal case where the transcript legitimately doesn't exist yet: Claude Code only creates `<uuid>.jsonl` when the first user message is sent. Launch the session, get distracted for a minute before typing, and sync is silently and unrecoverably dead for that session's entire lifetime.

Note this is distinct from #846: there the message writes succeed (`seq` increments server-side) and the session API returns `metadata = null`. Here the messages are never read or sent at all - the failure is entirely local, in the watcher.

## Environment

- Happy CLI: 1.2.0 (latest at time of writing)
- Claude Code: 2.1.207
- Node: v24.16.0
- OS: Linux (Pop!_OS, kernel 6.18)
- Started from terminal (`happy claude`), daemon running

## Root cause

In `dist/index-*.mjs` (file watcher started by the session scanner):

```js
const missingFileTimeoutMs = options.missingFileTimeoutMs ?? 6e4;
...
if (isMissing) {
  const now = Date.now();
  if (missingSince === null) missingSince = now;
  const absentMs = now - missingSince;
  if (absentMs >= missingFileTimeoutMs) {
    logger.debug(`[FILE_WATCHER] Giving up on ${file}: never appeared after ${Math.round(absentMs / 1e3)}s`);
    options.onGaveUp?.();
    return;   // <-- permanent; nothing ever re-arms this watcher
  }
}
```

`onGaveUp` propagates to the session scanner, which drops the session:

```
[SESSION_SCANNER] Session <uuid> transcript never appeared — dropping it
```

The timeout races against the user's first prompt, because Claude Code creates the transcript jsonl lazily on first message. There is no retry, no re-arm on the `SessionStart` hook (which already delivers `transcript_path` - same observation as the analysis in #1111), and no recovery when the file subsequently appears.

## Log evidence

Session started 10:11:18; watcher gave up ~62s later; the transcript was created shortly after and has been actively written for 3 days since - with zero uploads:

```
[10:12:16.917] [SESSION_SCANNER] Session file not found: ~/.claude/projects/<project>/<uuid>.jsonl
[10:12:19.916] [SESSION_SCANNER] Reading session file: ~/.claude/projects/<project>/<uuid>.jsonl
[10:12:20.671] [SESSION_SCANNER] Session <uuid> transcript never appeared — dropping it
```

Meanwhile the same process stays fully connected and functional:

```
[01:02:18.578] Socket connected successfully
[15:00:44.216] [RPC] Sending encrypted response { method: '<happy-session-id>:bash', ... }   // web UI git-status RPC works
```

`grep -l "transcript never appeared" ~/.happy/logs/*.log` shows **14 sessions** silently hit by this on one machine over ~6 days. Sessions where the first prompt happened to land inside the 60s window sync fine - which makes the failure look random/intermittent from the app side.

## Reproduction

1. `happy claude` from a terminal
2. Do nothing for 65 seconds
3. Send a prompt and have a conversation
4. Open the session in Happy Web/mobile: it's listed (title works, git status works) but the body is empty, and stays empty forever

## Expected behavior

The watcher should tolerate the transcript not existing until the first message, and should recover if the file appears after a giveup.

## Suggested fix

Any of (in order of robustness):

1. Re-arm the watcher from the `SessionStart` hook - the hook payload already includes `transcript_path`, so the scanner can (re)bind to the real file whenever the hook fires, which also helps the path-mismatch cases in #1111/#1359.
2. Don't treat ENOENT-before-first-write as a giveup condition at all: watch the project directory for the file's creation instead of the file itself, then switch to watching the file.
3. Minimally: make `missingFileTimeoutMs` much larger (hours, not seconds) - the ENOENT loops in #1111/#1359 were pathological path mismatches; a 60s ceiling punishes the normal lazy-creation case far more often than it saves resources.

Happy to provide full logs on request.

leikaiwei:
## Summary

Claude Code creates a session transcript lazily — `~/.claude/projects/<slug>/<uuid>.jsonl` does not exist at launch, only once the user submits their first prompt. `startFileWatcher` arms `fs.watch()` immediately, gets `ENOENT`, backs off, and after 60s calls `onGaveUp`; `sessionScanner` then adds the session to `deadSessions` and never reads it again. The session stays registered server-side, so the phone shows the entry, the title updates and RPC works — but no message body is ever uploaded, and it cannot recover for the rest of that session's lifetime. Anyone who starts Happy, reads some code, and types their first prompt more than 60 seconds later hits this. This PR stops bounding "file has not appeared yet" by time and instead waits for the file to be created, narrowing the give-up condition to what it should always have been: the parent directory itself being unwatchable.

Fixes #1538.

## Why the timeout is the wrong tool here

"Not created yet" and "never going to be created" are indistinguishable by elapsed time, so no value of `missingFileTimeoutMs` separates them. The bound arrived in 0efede47 (part of #1332), which records no rationale for it — the commit message and PR description do not mention the watcher, and the code comments describe only what the timeout does. Whatever it was reaching for, it cannot tell a permanently wrong path from a transcript that simply has not been written yet, and it resolves both by discarding the session.

The fix removes the need for the bound rather than tuning it. When the file is absent, watch its **parent directory** and block until the entry appears. That is event driven, so the per-second spin that motivated the timeout is gone on its own.

Events alone are not sufficient. NFS, SMB and some container bind mounts deliver no `inotify` events at all, so a pure event wait would block forever — the same bug in a quieter form, with the phone online and a title showing while nothing syncs. The directory watch therefore **races a 2s existence poll**: events are the fast path, the poll is the safety net.

`missingFileTimeoutMs` keeps its name and default but now bounds only one thing: the parent directory staying *continuously* unwatchable (missing, `EACCES`). That is the condition the original timeout was actually guarding.

## Proof

Before, from a machine running Happy through a normal working day. **9 distinct sessions** were dropped this way in that single day:

```
[11:05:59.296] [FILE_WATCHER] Giving up on /Users/kelvin/.claude/projects/-Users-kelvin-code-hoperun-litellm--/d674e55c-….jsonl: never appeared after 60s
[11:05:59.296] [SESSION_SCANNER] Session d674e55c-… transcript never appeared — dropping it
```

Each of those sessions was live and healthy in the terminal; the phone showed it online, with its title, and synced nothing for the rest of its life.

After, same machine, `Iteration with mode: local`. The transcript took **3 minutes 22 seconds** to appear — the user was reading code before typing anything — and the watcher picked it up cleanly:

```
[22:15:20.213] [SESSION_SCANNER] Starting watcher for session: c4f6d2ff-…
[22:15:20.213] [FILE_WATCHER] Waiting for /Users/kelvin/.claude/projects/-Users-kelvin-github-happy/c4f6d2ff-….jsonl to be created
[22:18:42.426] [FILE_WATCHER] File appeared: /Users/kelvin/.claude/projects/-Users-kelvin-github-happy/c4f6d2ff-….jsonl
[22:18:42.427] [FILE_WATCHER] Starting watcher for /Users/kelvin/.claude/projects/-Users-kelvin-github-happy/c4f6d2ff-….jsonl
```

On the old code that session was already blacklisted at 22:16:20, 2m22s before its transcript existed. A second run in the same session shows the shorter, more common case (57s — still under the old limit only by luck):

```
[21:07:37.103] [FILE_WATCHER] Waiting for …/e351cc52-….jsonl to be created
[21:08:34.113] [FILE_WATCHER] File appeared: …/e351cc52-….jsonl
```

I have been running this patch as my daily driver since, across both local and remote sessions, with no dropped session and no recurrence of the give-up path.

`794 tests passed (83 files)`, `tsc --noEmit` and `pnpm --filter happy build` all clean.

## Also in this change

Each of these is a consequence of the watcher now waiting indefinitely; they are not drive-by cleanups:

- Distinguish "directory iterator ended normally" from "directory is unwatchable". Only the latter arms the give-up countdown — reporting `unwatchable` on a guess would reintroduce the bug through the back door.
- Reset `failureCount` only after an event is actually delivered. Resetting before arming pins a persistently failing watch (`EACCES`, `EMFILE`) at the 1s floor forever.
- Clear `directoryUnwatchableSince` once the directory is watchable again, so an early transient failure cannot make a much later, unrelated failure trip the deadline with zero grace.
- Release the directory watcher when the existence re-check wins the race and the iterator is never entered, which otherwise leaks one OS watch per session.
- Re-arm if the file is deleted and later recreated.
- Flush and stop the sync **before** disposing watchers in `cleanup()`. `invalidateAndAwait` really does run the sync body once more, and that body re-creates a watcher for every session lacking one; the old order let that final sync resurrect them all with no remaining path to dispose them. Harmless while watchers self-terminated after 60s, a leak until process exit now that they wait indefinitely.
- Isolate `CLAUDE_CONFIG_DIR` in `sessionScanner.test.ts`, which until now created directories under the developer's real `~/.claude/projects` and `rm -rf`'d them in `afterEach`.

## Relationship to nearby issues and PRs

Same file or same symptom, different root causes — this PR does not subsume or conflict with any of them:

- **#1359** (watcher loops on `ENOENT` for a ghost session id) and **#1111** (`-w` worktree deriving the wrong transcript path) are both about a path that is *permanently* wrong — the case a bounded give-up genuinely serves, which is why this PR keeps one. Worth noting that #1359 was filed two weeks *after* 0efede47, so the existing bound did not resolve it. This PR addresses the opposite case: a path that is correct but simply not written yet.
- **#1528** (draft) touches `startFileWatcher.ts` but only threads `initialRetryDelayMs`/`maxRetryDelayMs` through for test determinism; it does not change the give-up branch. Its substance is in `sessionScanner.ts`, recovering ghost ids from an explicit `SessionStart` `transcript_path`. That cannot fix this case: at `SessionStart` the file does not exist yet, the hook's path equals the one already being watched, and `SessionStart` fires once — before the 60s deadline, with no later chance to re-attach.
- **#1586** shares the symptom (blank session, shown online) but the cause is a 3 GB transcript hitting `ERR_STRING_TOO_LONG`, surfaced as `Session file not found`. Unrelated mechanism.
- **#846** is server-side `metadata = null` on Windows; #1538 already notes the two are distinct.

## Scope

`happy-cli` only. No protocol, encryption, server or app changes; no change to what gets uploaded or how. The sync engine's behaviour is untouched apart from when a watcher decides to stop trying.

### DIFF
--- a/packages/happy-cli/src/claude/claudeRemote.ts
+++ b/packages/happy-cli/src/claude/claudeRemote.ts
@@ -295,13 +295,14 @@ export async function claudeRemote(opts: {
                     const found = await awaitFileExist(join(projectDir, `${systemInit.session_id}.jsonl`), 30000);
                     logger.debug(`[claudeRemote] Session file found: ${systemInit.session_id} ${found}`);
                     if (!found) {
-                        // The transcript never landed on disk within the grace
-                        // window. We still register the id so the (now
-                        // bounded) scanner watcher can pick it up if it shows
-                        // up late and otherwise drops it cleanly instead of
-                        // wedging — but surface the anomaly so a stuck remote
-                        // launch is visible in the app rather than a silent
-                        // "dead instance".
+                        // The transcript has not landed on disk yet. We still
+                        // register the id: the scanner's watcher waits for the
+                        // file indefinitely (it no longer gives up on a
+                        // deadline — see startFileWatcher), so a late
+                        // transcript is still picked up cleanly. This wait is
+                        // therefore only about surfacing the anomaly, so a
+                        // genuinely stuck remote launch is visible in the app
+                        // rather than a silent "dead instance".
                         logger.debug(`[claudeRemote] WARNING: session transcript ${systemInit.session_id} never appeared after 30s`);
                         opts.onCompletionEvent?.('⚠️ Claude session did not produce a transcript — the agent may be unresponsive. Try sending your message again.');
                     }
--- a/packages/happy-cli/src/claude/utils/sessionScanner.test.ts
+++ b/packages/happy-cli/src/claude/utils/sessionScanner.test.ts
@@ -10,14 +10,25 @@ import { getProjectPath } from './path'
 
 describe('sessionScanner', () => {
   let testDir: string
+  let configDir: string
   let projectDir: string
   let collectedMessages: RawJSONLines[]
   let collectedTranscriptEvents: ClaudeGoalStatusTranscriptEvent[]
   let scanner: Awaited<ReturnType<typeof createSessionScanner>> | null = null
-  
+  let prevConfigDir: string | undefined
+
   beforeEach(async () => {
-    testDir = join(tmpdir(), `scanner-test-${Date.now()}`)
+    const unique = `${Date.now()}-${Math.random().toString(36).slice(2)}`
+    testDir = join(tmpdir(), `scanner-test-${unique}`)
     await mkdir(testDir, { recursive: true })
+
+    // Isolate CLAUDE_CONFIG_DIR *before* deriving projectDir, since
+    // getProjectPath() reads it. Otherwise these tests create directories
+    // under the developer's real ~/.claude/projects and rm -rf them in
+    // afterEach.
+    configDir = join(tmpdir(), `scanner-config-${unique}`)
+    prevConfigDir = process.env.CLAUDE_CONFIG_DIR
+    process.env.CLAUDE_CONFIG_DIR = configDir
 
     // Use the same path calculation as the scanner to ensure paths match
     projectDir = getProjectPath(testDir)
@@ -26,19 +37,25 @@ describe('sessionScanner', () => {
     collectedMessages = []
     collectedTranscriptEvents = []
   })
-  
+
   afterEach(async () => {
     // Clean up scanner
     if (scanner) {
       await scanner.cleanup()
       scanner = null
     }
-    
+
+    if (prevConfigDir === undefined) {
+      delete process.env.CLAUDE_CONFIG_DIR
+    } else {
+      process.env.CLAUDE_CONFIG_DIR = prevConfigDir
+    }
+
     if (existsSync(testDir)) {
       await rm(testDir, { recursive: true, force: true })
     }
-    if (existsSync(projectDir)) {
-      await rm(projectDir, { recursive: true, force: true })
+    if (existsSync(configDir)) {
+      await rm(configDir, { recursive: true, force: true })
     }
   })
   
@@ -319,11 +336,14 @@ describe('sessionScanner', () => {
     // }
   })
 
-  it('drops a phantom session whose transcript never appears and keeps serving real ones', async () => {
-    // Reproduces the "dead Happy instance" bug: a session id is announced
-    // (e.g. a remote launch) but its .jsonl is never written. The scanner
-    // must give up on it instead of spinning forever, and must still process
-    // a real session that arrives afterwards.
+  it('keeps serving real sessions while a phantom session has no transcript', async () => {
+    // A session id is announced (e.g. a remote launch) but its .jsonl is
+    // never written. The watcher now waits for that file indefinitely via
+    // the parent directory instead of giving up on a deadline — a transcript
+    // Claude Code creates lazily must never be abandoned (that dropped real
+    // sessions whose first prompt came late). The phantom therefore stays
+    // watched, costing one idle directory watch and no CPU, and must not
+    // leak messages or disturb a real session arriving afterwards.
     scanner = await createSessionScanner({
       sessionId: null,
       workingDirectory: testDir,
--- a/packages/happy-cli/src/claude/utils/sessionScanner.ts
+++ b/packages/happy-cli/src/claude/utils/sessionScanner.ts
@@ -30,9 +30,14 @@ export async function createSessionScanner(opts: {
     onMessage: (message: RawJSONLines) => void
     onTranscriptEvent?: (event: ScannerTranscriptEvent) => void
     /**
-     * How long a session transcript may stay absent before its watcher gives
-     * up and the session is dropped. Defaults to the startFileWatcher default
-     * (60s). Exposed mainly so tests can exercise the drop path quickly.
+     * How long the transcript's parent directory may stay continuously
+     * unwatchable before the watcher gives up and the session is dropped.
+     * Defaults to the startFileWatcher default (60s). Exposed mainly so tests
+     * can exercise the drop path quickly.
+     *
+     * Note this no longer bounds how long we wait for the transcript *file* —
+     * that wait is unbounded on purpose, because Claude Code creates it lazily
+     * on the first prompt. See startFileWatcher for the full rationale.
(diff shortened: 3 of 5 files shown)


chphch:
Independent confirmation from another machine (macOS, self-hosted server), plus a controlled before/after.

**It reproduces exactly as described.** One of my sessions last night, from the CLI log:

```
23:42:38.651  [SESSION_SCANNER] Starting watcher for session: f125a14c-…
23:43:38.677  [FILE_WATCHER]    Giving up on …/f125a14c-….jsonl: never appeared after 60s
23:43:38.678  [SESSION_SCANNER] Session f125a14c-… transcript never appeared — dropping it
```

and the transcript's birth time on disk:

```
23:43:59.400  f125a14c-….jsonl created      ← 21 s after the watcher gave up
```

That session then ran for a long working stretch and did real work — it even set its own title through the MCP tool, so the CLI↔server link was alive the whole time. The phone showed the session online, with that title, and **zero message bodies**, permanently. Server-side it stayed at `seq = 0` for its entire lifetime.

**The 60 s boundary is the whole discriminator.** Same machine, same CLI build per row; the only thing varied is how long I waited before the first prompt:

| CLI | first prompt after launch | messages on server |
|---|---|---|
| current `main` | 39 s | 9 ✅ |
| current `main` | > 60 s (the session above) | **0** ❌ |
| **this PR** | **128 s** | 9 ✅ |

With the PR applied the watcher logged no give-up at all, and the first message reached the server 7 s after the prompt. `startFileWatcher` + `sessionScanner` suites pass (15/15), `tsc --noEmit` clean. I've since deployed it to my daily-driver instance and re-verified the same 128 s case against the installed build rather than a local one.

One detail worth keeping in the final version: the directory watch races a low-frequency existence poll. I would have been tempted to drop the poll as redundant, but a purely event-driven wait is the same bug in a quieter form wherever watch events never arrive.

@bra1nDump this is still `MERGEABLE` / `CLEAN` against `main` with no review in the month since it was opened, and it silently loses whole sessions for anyone who reads the screen before typing. Would you take a look? Happy to re-verify on my side after a rebase if that helps.

leikaiwei:
Thanks for reproducing this — the 21-second gap, and a session sitting at `seq = 0` while still setting its own title, say it better than my write-up did.

The poll stays. You're right that it looks redundant and right about why it isn't; the reasoning is pinned in a comment above `CREATION_POLL_INTERVAL_MS`.

Still `MERGEABLE` / `CLEAN`, and it's been my daily driver for the past month. Happy to rebase whenever a reviewer wants it — I'll ping you here if the diff changes enough to invalidate your run.

PR #1660, opened 2026-08-06
feat(cli,app): surface the model-scoped weekly limit in the usage popup

charliezong18:
Since #1556 the usage row reads plan rate limits from `get_usage`. That response still carries the top-level `five_hour` / `seven_day` buckets, but the legacy per-model buckets (`seven_day_opus`, …) now come back `null` — the backend reports model-scoped allotments (currently the Fable weekly cap) as entries in a `limits` array instead. Nothing reads that array, so a user who has burned through their Fable weekly allowance sees no sign of it anywhere in Happy.

**Before:** with a `weekly_scoped` entry at 17% in the payload, the usage popup lists Session and Week only.
**After:** it lists a third row under the backend's own display name — `Fable · 17%`, with its reset time.

**CLI** — `windowsFromGetUsage` maps scoped `limits[]` entries into windows keyed group + scope (`weekly_fable`), using the scope's display name as the label. Unscoped kinds (`session`, `weekly_all`) duplicate the top-level buckets and are skipped, so nothing double-counts.

**App** — `getUsageLimitRows` now reports whether the backend named a window. `label` is assigned in exactly one place in the CLI (the scoped-limits mapper), so a name is a reliable "a human asked for this allotment" signal. The popup lists named rows after Session and Week. That keeps provider-internal windows (`nimbus_quill` and friends) out — they arrive unnamed — without an id whitelist every future scope would have to be added to.

**Rebased onto current `main`.** This PR predates c15eeee9. Its original app half added a chip to the status bar; that status bar is gone, so the app half now targets the usage popup instead. `CHIP_WINDOW_LABELS` is left exactly as `main` has it.

**Disclosure — the id is derived, not declared.** The payload gives no stable key for a scoped limit, so the id is built from the scope's display name (`"Fable"` → `weekly_fable`). Rendering is label-driven, so if that display name changes the id changes with it but the row still appears under the new name; only cross-snapshot id continuity is lost. A backend-declared id would be better; this is what the payload offers today.

## Validation

- `happy-app` typecheck clean
- `happy-cli` typecheck: no new errors relative to `main` (same check run on a pristine `main` checkout for comparison)
- `usageLimits` 17/17, `sessionStatusBar` 13/13
- 3 tests: CLI scoped-limit mapping, CLI malformed/scope-less entries, app named-vs-unnamed row classification

The two screenshots that were here showed the status-bar chip, which `main` no longer has, so they're removed rather than left misleading. Happy to attach a fresh popup capture if that helps review.

### DIFF
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -895,13 +895,15 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
     const contextStatus = props.usageData?.contextSize
         ? getContextStatus(props.usageData.contextSize, props.alwaysShowContextSize ?? false, theme, props.usageData.contextWindow)
         : null;
-    // Only Session and Week are user-meaningful; provider-internal windows
-    // (nimbus_quill and friends) stay out of the popup.
+    // Session, Week, and any allotment the backend bothered to name (the
+    // model-scoped weekly caps) are user-meaningful; provider-internal windows
+    // (nimbus_quill and friends) arrive unnamed and stay out of the popup.
     const usageRows = React.useMemo(() => {
         const rows = getUsageLimitRows(props.sessionStatusUsageLimits ?? null);
         const session = rows.find((row) => row.id === 'five_hour') ?? null;
         const week = rows.find((row) => row.id === 'seven_day') ?? null;
-        return { session, week };
+        const scoped = rows.filter((row) => row.scoped);
+        return { session, week, scoped };
     }, [props.sessionStatusUsageLimits]);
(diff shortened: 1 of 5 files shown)


charliezong18:
## Visual proof

Running my own account against this branch. My Fable weekly allowance happens to be fully spent right now, which is exactly the case this PR is about — on `main` that state is invisible in Happy.

**Status bar** — the `fable` chip sits alongside `5h` and `7d`, in the rejected color because the allowance is exhausted:

status bar with fable chip

**Limits popover** — the scoped window renders with the display name the backend sent (`Fable`), next to the two top-level windows:

limits popover listing the Fable window

(The bar is in "remaining" mode and a non-English locale, hence "剩余 0%" / "重置" — that's the existing display toggle, not something this PR adds.)

Without this change the `limits[]` entry is dropped in the CLI, so neither the chip nor the popover row exists and a spent Fable allowance shows up nowhere.

PR #1661, opened 2026-08-06
feat(app): long-press option chips to pick several and run them in sequence

charliezong18:
Option chips send one option per tap, which is right most of the time. But agents routinely offer options that aren't mutually exclusive — "run the tests / update the docs / open a PR" — and picking three today means three round trips, waiting out a full turn between each.

Long-press a chip to enter selection mode: checkmarks appear, tap to toggle, and **Run in sequence** composes the picked options into one numbered message so the agent works through them in order. A single tap outside selection mode behaves exactly as before.

The scope is deliberately small: selection is `useState` local to the one options block. Nothing is persisted, nothing is synced, and it clears on send, cancel, or unmount. No wire, storage, or settings change.

On web the chips suppress the browser's native context menu so a right-click or mouse-hold enters selection mode instead of popping the OS menu — the same approach already used in `SessionsList`.

Both new strings go through `t()` (`markdown.optionsRunInSequence`, `markdown.optionsSequencePrompt`) and are filled in for all 10 locales.

## Visual proof

Captured from this branch in the web app, via the app's own `/dev/messages-demo` route with one temporary `<options>` message in the demo data (that edit is **not** part of this PR — it was reverted after the screenshots). The component, press handlers and send path are the real ones.

Before — single tap untouched:

option chips, normal state

Long-press (right-click on web) enters selection mode; two of three picked:

two options selected with Run in sequence

After sending, selection clears:

selection cleared after sending

## Validation

- `happy-app` typecheck ✅
- app suite 909/909

### DIFF
--- a/packages/happy-app/sources/components/markdown/MarkdownView.tsx
+++ b/packages/happy-app/sources/components/markdown/MarkdownView.tsx
@@ -1,6 +1,7 @@
 import { MarkdownSpan, parseMarkdown } from './parseMarkdown';
 import * as React from 'react';
 import { Image, Pressable, View, Platform } from 'react-native';
+import { Ionicons } from '@expo/vector-icons';
 import { HorizontalScrollView } from '../HorizontalScrollView';
 import { Gesture, GestureDetector } from 'react-native-gesture-handler';
 import { StyleSheet } from 'react-native-unistyles';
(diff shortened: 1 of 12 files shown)


charliezong18:
## Visual proof

Captured from this branch running in the web app. To get a reproducible options block I temporarily added one `<options>` message to `sources/app/(app)/dev/messages-demo-data.ts` and used the app's own `/dev/messages-demo` route — that edit is **not** part of this PR, it was reverted after the screenshots. The component, the press handlers and the send path are the real ones (`MessageView` always wires `onOptionPress`).

**Before** — single tap is untouched:

option chips, normal state

**Selection mode** — long-press (right-click on web) enters it; two of the three are picked and the action row appears:

two options selected with Run in sequence

**After sending** — selection clears back to normal:

selection cleared after sending

One useful detail from that last step: the only console output is `Session demo-messages-session not found after sync`, i.e. the composed message did reach `sync.sendMessage` and only failed because the demo route has no real session behind it. The composed body is one message:

```
Please do the following in order:
1. Run the test suite
2. Open a pull request
```

Issue #1475, opened 2026-07-03
Mermaid diagrams clipped — no height reporting or scroll in WebView renderer


kemier:
## Bug

Mermaid diagrams rendered in chat are cut off with no scrollbar to see the full content.

## Affected platform

- Native (Android/iOS) — the WebView-based renderer in `MermaidRenderer.tsx`
- Web platform is unaffected (uses direct SVG rendering)

## Root cause

Two issues in `packages/happy-app/sources/components/markdown/MermaidRenderer.tsx`:

### 1. No height reporting from WebView → React Native

The WebView HTML template never calls `window.ReactNativeWebView.postMessage()` to report the rendered diagram height. The `onMessage` handler in the React Native component expects dimension updates:

```tsx
onMessage={(event) => {
    const data = JSON.parse(event.nativeEvent.data);
    if (data.type === 'dimensions') {
        setDimensions(prev => ({
            ...prev,
            height: Math.max(prev.height, data.height)
        }));
    }
}}
```

But the HTML inside the WebView has no corresponding `postMessage` call. The container stays at the initial `height: 200`, clipping any diagram taller than 200px.

### 2. Scrolling disabled

```tsx
<WebView
    source={{ html }}
    style={{ flex: 1 }}
    scrollEnabled={false}   // ← no scrolling
    ...
/>
```

Even if the diagram is taller than the container, the user cannot scroll to see the rest.

## Proposed fix

1. **Add height reporting** — After Mermaid renders the SVG, call `postMessage` with the actual content height:
   ```js
   function reportHeight() {
       const h = Math.max(document.body.scrollHeight, document.documentElement.scrollHeight);
       window.ReactNativeWebView && window.ReactNativeWebView.postMessage(JSON.stringify({ type: 'dimensions', height: h }));
   }
   // Call after render + delayed retry for async SVG sizing
   requestAnimationFrame(() => { reportHeight(); setTimeout(reportHeight, 200); });
   ```

2. **Enable scrolling** — Set `scrollEnabled={true}` so the WebView can scroll internally for tall diagrams.

3. **Cap max height** — Set container height to `Math.min(dimensions.height, 600)` so tall diagrams scroll within a 600px container instead of taking over the entire chat screen.

## Environment

- happy-app: current main
- Platform: Android (PICO 4 Ultra VR headset)
- Render path: `MermaidRenderer.tsx` → WebView with mermaid@11 from CDN

## Files

[...]

PR #1662, opened 2026-08-07
fix(app): stop cropping tall mermaid diagrams on native

chphch:
A mermaid diagram taller than 600dp is silently cropped on native: `MermaidRenderer` sizes its container to `Math.min(dimensions.height, MAX_DIAGRAM_HEIGHT)` with `MAX_DIAGRAM_HEIGHT = 600`, so everything past that line is simply not shown, with no scrollbar, fade, or any other indication that the diagram continues. The constant's comment says tall diagrams "scroll inside a capped container", and the `WebView` does set `scrollEnabled` — but on Android the chat list takes the vertical gesture, so dragging over the diagram scrolls the message list and never the diagram. The cropped part is unreachable in practice. This raises the ceiling to 8000 and restates it as what it should be: a guard against a bad height measurement producing an absurd box, rather than a layout decision about how tall a diagram is allowed to be.

Two things make this more than a constant that was tuned too low:

- **Web was never capped.** `webStyle` sets no height, so the same diagram renders in full in the browser and cropped on a phone. The platforms disagreed about the content, not just the presentation.
- **Sizing the container to its content removes the gesture conflict rather than relying on it.** Once the container is as tall as the diagram there is nothing to scroll inside the `WebView`, so the nested-scroll ambiguity that made the capped version unusable does not arise.

The measurement this feeds is already bounded by real content — `reportHeight()` reads the rendered SVG's `getBoundingClientRect().height` rather than `document.body.scrollHeight`, which is what an earlier fix in this file corrected — so the ceiling is only there for the pathological case.

## Proof

Reported and confirmed on a physical Android device (Samsung, 1080×2340) running a self-hosted build of this app. The diagram was a 15-node `flowchart TD` laying out to roughly 650dp at the message column width: before the change its bottom two nodes and their incoming edges were missing, and the fullscreen viewer — a fork-only affordance that does not exist on this code path upstream — showed them present, which is what identified the crop rather than a rendering failure. After the change the same diagram renders complete. `pnpm typecheck` is clean and `pnpm exec vitest run` passes 909/909 on this branch.

I did not attach a capture: the behaviour is native-only (web has no cap, so a browser recording shows the same thing before and after), and I don't have an Android capture rig that reaches the chat surface. Happy to produce one on an emulator if that would help review.

### DIFF
--- a/packages/happy-app/sources/components/markdown/MermaidRenderer.tsx
+++ b/packages/happy-app/sources/components/markdown/MermaidRenderer.tsx
@@ -5,8 +5,13 @@ import { StyleSheet, useUnistyles } from 'react-native-unistyles';
 import { Typography } from '@/constants/Typography';
 import { t } from '@/text';
 
-// Tall diagrams scroll inside a capped container instead of taking over the chat
-const MAX_DIAGRAM_HEIGHT = 600;
+// Runaway guard on the WebView-reported height, NOT a design cap. It used to be
+// 600, which silently cropped any diagram taller than that: the container is a
+// fixed-height WebView, so the excess was only reachable by scrolling *inside*
+// it, and the chat list swallows that gesture on Android. Web has never capped
+// the diagram (see webStyle below), so the two platforms disagreed as well.
+// Keep a ceiling only so a bad measurement cannot produce an absurd box.
+const MAX_DIAGRAM_HEIGHT = 8000;
 
 // Style for Web platform
 const webStyle: any = {


chphch:

**#1475** (Jul 3) reports this independently, and @kemier went as far as locating the cause: `MermaidRenderer.tsx` never reports its height back from the WebView, so a tall diagram is clipped with nothing to scroll. Native only; web renders SVG directly and is unaffected.

That is exactly the file this PR touches, and the only one — 9 lines.

@bra1nDump — small and self-contained if you want an easy one to land.

PR #1663, opened 2026-08-07
fix(ci): record @slopus/happy-wire under devDependencies in the lockfile

chphch:
## main is red at the install step

`pnpm install --frozen-lockfile` fails on `main` as of 38767705, so CI stops before any check runs — on main itself and on every PR opened against it.

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because
pnpm-lock.yaml is not up to date with <ROOT>/packages/happy-cli/package.json

Failure reason:
  "devDependencies" in the lockfile ({"@eslint/compat":…}) doesn't match the
  same field in package.json ({"@eslint/compat":"^1",
  "@slopus/happy-wire":"workspace:*", …})
```

## Cause

e7e0ff6b moved `@slopus/happy-wire` from `dependencies` to `devDependencies` in `packages/happy-cli/package.json` — which is the right fix for the DOA `1.2.1-beta.0`, since `dependencies` is pkgroll's externals policy. The lockfile still records it under `dependencies` for that importer, and pnpm compares the two field by field.

## The change

The three lockfile lines move from `dependencies` to `devDependencies`. Nothing else changes — resolution is `link:../happy-wire` either way; only the field recording it moves.

A plain `pnpm install --lockfile-only` regeneration also clears the mismatch, but it drags in unrelated churn from re-resolved floating ranges — +76/-96 on my machine, including a `typescript` resolution moving 5.9.3 → 6.0.3 under `@prisma/client`. That did not belong in a CI unblock, so the entry is moved by hand.

## Verification

Clean checkout, pinned pnpm 10.11.0 (the `packageManager` version):

- `pnpm install --frozen-lockfile` on 38767705 → `ERR_PNPM_OUTDATED_LOCKFILE`
- same command on this branch → `Done in 13.4s`, exit 0
- the install leaves `pnpm-lock.yaml` byte-identical afterwards (no drift hiding in it)
- `packages/happy-wire` and `packages/happy-cli` both build clean
- `packages/happy-cli/dist` carries no `import` or `require` of `@slopus/happy-wire` — the only occurrences are the bundled copy of package.json's own text — so the inlining e7e0ff6b was after still holds with the entry in `devDependencies`

For cross-reference: `main`'s own CLI Smoke Test is failing on 38767705 and passing on 9ac022f1, which brackets the change.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -927,9 +927,6 @@ importers:
       '@paralleldrive/cuid2':
         specifier: ^2.2.2
         version: 2.3.1
-      '@slopus/happy-wire':
-        specifier: workspace:*
-        version: link:../happy-wire
       '@stablelib/base64':
         specifier: ^2.0.1
         version: 2.0.1
@@ -1018,6 +1015,9 @@ importers:
       '@eslint/compat':
         specifier: ^1
         version: 1.4.1(eslint@9.39.2(jiti@2.6.1))
+      '@slopus/happy-wire':
+        specifier: workspace:*
+        version: link:../happy-wire
       '@types/inquirer':
         specifier: ^9.0.9
         version: 9.0.9


chphch:
@bra1nDump heads-up, since this one blocks you rather than me: `main` is red at the install step, so it lands before any check gets to run.

`e7e0ff6b` moved `@slopus/happy-wire` to `devDependencies` in `packages/happy-cli/package.json` — right call, `dependencies` is pkgroll's externals policy — but `pnpm-lock.yaml` still records it under `dependencies` for that importer, and pnpm compares the two field by field. Repro on a clean checkout of 38767705:

```
pnpm install --frozen-lockfile
→ ERR_PNPM_OUTDATED_LOCKFILE  ... "devDependencies" in the lockfile doesn't
  match the same field in package.json
```

`main`'s own CLI Smoke Test went from green on 9ac022f1 to red on 38767705, and every PR opened against main dies at "Install dependencies" — #1581 included.

This PR moves the three lockfile lines and changes nothing else; resolution stays `link:../happy-wire` either way. I did it by hand rather than regenerating, because `pnpm install --lockfile-only` also clears the mismatch but drags in +76/-96 of unrelated churn from re-resolved floating ranges, including a `typescript` resolution moving 5.9.3 → 6.0.3 under `@prisma/client` — not something to smuggle into a CI unblock.

Verified with the pinned pnpm 10.11.0: fails on 38767705, `Done in 13.4s` on this branch, install leaves the lockfile untouched afterwards, and `happy-cli/dist` still carries no `import`/`require` of `@slopus/happy-wire`, so the inlining you were after holds with the entry in `devDependencies`. All three checks green here.

No ask beyond this one — happy to close it if you'd rather regenerate the lockfile yourself.

chphch:
Closing — this landed independently as 5b852aa2 ("fix(cli): sync happy-wire lockfile importer", Aug 10). That commit makes the identical change: it moves `@slopus/happy-wire` out of the `dependencies` block and into `devDependencies` in the `packages/happy-cli` importer of `pnpm-lock.yaml`. The patch here no longer applies to `main`.

For anyone arriving here from a red CI run: while the lockfile was out of sync, `pnpm install --frozen-lockfile` failed with `ERR_PNPM_OUTDATED_LOCKFILE` ("devDependencies in the lockfile … doesn't match the same field in package.json") on any branch cut from an older `main`. It surfaced as a `typecheck` and/or `smoke-test` failure that had nothing to do with the PR's own diff. Rebasing onto current `main` clears it.

Thanks — glad the fix is in either way.

PR #1664, opened 2026-08-07
fix(app): surface the home dock's custom path entry above the path list

chphch:
The session list's dock composer lets you pick where a new session runs, and typing a directory that has no session yet is the `Enter custom path` row in its PROJECT picker — but that row is appended *after* every distinct path the selected machine has ever run a session in. That list has no bound: on my machine it is 18 entries, the sheet shows four rows at a time, and the only way to type a directory ends up five swipes down. It reads as absent, which is how I hit this — I went looking for a way to set the cwd from the dock and concluded there wasn't one.

Moving the action to the top of the options list is the whole change. It costs nothing when the list is short, and it makes the two new-session entry points agree: the `/new` screen already puts its path field above the same recent-path suggestions (`PathPickerContent`), so only the dock ordered them the other way round.

Both platforms read the same array — iOS renders it as a `NativeOptionsPicker` menu, Android as the in-modal sheet — so the one change covers both.

Device capture on an Android emulator to follow in a comment.

### DIFF
--- a/packages/happy-app/sources/components/HomeDock.tsx
+++ b/packages/happy-app/sources/components/HomeDock.tsx
@@ -889,11 +889,16 @@ export const HomeDock = React.memo(({
             return {
                 title: 'Project',
                 options: [
-                    ...projectOptions,
+                    // First, not last: projectOptions grows with every distinct
+                    // path the machine has ever run a session in, so appending
+                    // the only way to type a path buries it below a list with no
+                    // bound. The /new screen already puts its path field above
+                    // the same suggestions.
                     {
                         key: CUSTOM_PROJECT_PATH_KEY,
                         name: t('machineLauncher.enterCustomPath'),
                     },
+                    ...projectOptions,
                 ],
                 selectedKey: currentProject?.key,
                 onSelect: (key) => {


chphch:
### Device capture — Android emulator, matched pair

before/after

Pixel 5 AVD (arm64, 1080×2340 @440dpi = 393dp), API 30. Same emulator, same install, same account — the two frames differ only in the JS bundle: **BEFORE** is the APK's embedded bundle (`1.7.0+chphch.2059`, built before this change; pinned by launching once with networking off so the update channel could not replace it), **AFTER** is the same APK after it picked up a bundle carrying this commit. Both read off the Settings → Version row.

This account is a throwaway with no paired machine, so its PROJECT list is the minimum case — one path plus the action. That is the honest limit of what the capture shows: it proves the reorder, not the burial. The burial is the measurement in the PR body — the machine that prompted this has **18 distinct session paths**, and the sheet is capped at 270 dp, so it renders four rows at a time and `Enter custom path` was the 19th.

Rendering is the Android in-modal sheet here. iOS reads the same array through `NativeOptionsPicker`, so the ordering change lands there too, but I have no iOS device to shoot.

PR #1665, opened 2026-08-07
fix(app): keep the home dock's config rows legible over the focus scrim

chphch:
Tapping the session list's dock composer opens focus mode, which covers the screen with a `rgba(0, 0, 0, 0.88)` scrim. The machine / project / worktree rows sit above that scrim, and they are the only part of focus mode drawn with no surface of their own — the composer and the options sheet each render on `MobileGlassSurface`. Their text and icons still take `theme.colors.text`, which is `#000000` in the light theme, so they come out black on near-black and read as though the picker had been dimmed along with the background. That is how it was reported to me: "the part where I choose the cwd gets dimmed too."

The scrim is the same colour in both themes, so content painted directly on it cannot take its colour from the theme. Colouring those rows for the scrim fixes the light theme and leaves dark mode byte-identical — its `theme.colors.text` was already `#FFFFFF`.

Scoped to the compact (focus-mode) rows; the non-compact branch, which renders on light glass, keeps the theme colour.

Device capture to follow in a comment.

### DIFF
--- a/packages/happy-app/sources/components/HomeDock.tsx
+++ b/packages/happy-app/sources/components/HomeDock.tsx
@@ -65,6 +65,14 @@ type AgentSetting = 'agent' | 'model' | 'permission' | 'effort';
 
 const CUSTOM_PROJECT_PATH_KEY = '__custom_project_path__';
 
+// Focus mode dims the screen behind a fixed dark scrim (see `focusBackdrop`) in
+// BOTH themes, and the machine / project / worktree rows are drawn straight onto
+// it with no surface of their own — unlike the composer and the options sheet,
+// which each sit on light glass. So those rows have to be coloured for the
+// scrim, not for the theme: `theme.colors.text` is #000000 in the light theme,
+// which makes them black-on-black and reads as "the picker got dimmed too".
+const FOCUS_SCRIM_TEXT = '#FFFFFF';
+
 const AGENTS: Array<{ key: NewSessionAgentType; name: string }> = [
     { key: 'rig', name: 'Rig' },
     { key: 'claude', name: 'Claude Code' },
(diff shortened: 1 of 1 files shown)


chphch:
### Device capture — Android emulator, light theme

before/after

Pixel 5 AVD (arm64, 393 dp), API 30, light theme, same emulator and same APK — the frames differ only in the JS bundle, `d0af673` before and `b3d70d9` (this commit) after, each read off Settings → Version. Crop is the three focus-mode rows sitting on the scrim; the composer below them is unchanged.

In the BEFORE frame the row labels are `#000000` on the `rgba(0, 0, 0, 0.88)` scrim, which is why the "Open Camera" button of the dimmed screen behind them is about as legible as the rows themselves.

Dark mode needs no capture: `theme.colors.text` is already `#FFFFFF` there, so the rendered colour is identical before and after.

chphch:
Closing — and this one has to be closed rather than rebased, because the patch would now *cause* the bug it was written to fix.

The premise was that the focus scrim is the same colour in both themes, so content painted straight onto it cannot take its colour from the theme. That was true at this PR's base — `HomeDock.tsx` had `focusBackdrop: { backgroundColor: 'rgba(0, 0, 0, 0.88)' }`, unconditionally.

ce636c5a9 fixed it at the actual cause instead: the scrim is now theme-aware (`HomeDock.tsx:300-302`, `theme.dark ? 'rgba(0, 0, 0, 0.88)' : 'rgba(255, 255, 255, 0.88)'`), which makes `focusConfigValue` keeping `color: theme.colors.text` (`HomeDock.tsx:338-344`) correct in both themes.

So this PR's `FOCUS_SCRIM_TEXT = '#FFFFFF'` is now the wrong direction — it would paint white text and white icons onto the new white light-mode scrim. Better fix than mine; closing in its favour.

PR #1672, opened 2026-08-08
feat(app): start a session in a project straight from its card

chphch:
Every project card in the session list already knows which machine and directory its sessions run in, and shows that directory as its title. Starting another session there still meant opening `/new` and picking the same machine and the same path by hand — for a folder the list was displaying at that moment.

This gives each card the `+` the active-sessions header already has. It seeds the new-session draft from one of the card's own sessions and opens `/new` with the machine and path filled in, leaving only the prompt to type.

The seeding moves into one function both call. It was written inline in the header, and a second copy would have been the second place that has to remember two non-obvious things: `setMachineId` clears the path (so the path must be set after it), and a worktree session seeds the repository *plus* the worktree rather than the worktree path as a plain directory.

Cards are seeded from a session rather than from a path field on the group, so this also covers Rig cards, which carry a durable project identity instead of one working directory. A card whose sessions have no path shows no `+`.

proof

Captured on a Pixel 5 AVD (arm64, 393 dp, API 30) paired to a real `happy` daemon with two sessions in two directories. The star next to the `+` is from #1581 and is not part of this PR.

### DIFF
--- a/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
+++ b/packages/happy-app/sources/components/ActiveSessionsGroupCompact.tsx
@@ -18,7 +18,7 @@ import { SessionActionsAnchor, SessionActionsPopover } from './SessionActionsPop
 import { useSessionActionAlert } from '@/hooks/useSessionQuickActions';
 import { sessionKill } from '@/sync/ops';
 import { isWorktreePath, getRepoPath, getWorktreeName } from '@/utils/worktree';
-import { useNewSessionDraft } from '@/hooks/useNewSessionDraft';
+import { seedNewSessionDraftFrom } from '@/utils/newSessionFromProject';
 import { useRouter } from 'expo-router';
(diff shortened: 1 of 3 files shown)


chphch:
Closing — this capability shipped upstream in c63c80ff3 + 3cf9ce70f.

c63c80ff3 rebuilt the home list around projects and worktrees and put a new-session button on each card; 3cf9ce70f gave that button the seeding this PR was for. `ProjectGroup.tsx` now fills the draft from one of the card's own sessions and opens the prefilled home dock composer via `requestHomeDockFocus()`, falling back to `/new` where the dock is not mounted — the same fallback shape this PR needed.

Upstream's is broader than mine rather than narrower: the `+` sits on every worktree section instead of once per project group, so a repo with several worktrees gets one button per worktree. And the file this PR patched no longer exists in that shape (collapsible per-project header → per-worktree section), so rebasing would mean rewriting a feature that has already shipped.

PR #1673, opened 2026-08-08
fix(app): stop the composer showing a mic while it holds text

chphch:
The mobile composer folds send, stop and the mic into one button, and which one it draws comes from `hasText` — state the keystroke handler sets inside `React.startTransition`. React may defer that update for as long as higher-priority work keeps arriving, so the composer can hold a typed message while the button still shows a mic. A session streaming a reply is exactly that kind of work, which is why it surfaces mid-conversation and is hard to attribute to any one action; I only found it because a user reported "the send button shows up as the voice button" and could not say when.

What makes it worth fixing rather than tolerating is that the button's *behaviour* is already correct — `handleMobilePrimaryPress` and `handleSendPress` both read the live text and send. The comment above the first one says as much. So the stale icon is not a brief cosmetic lag; it is the button advertising dictation while it would in fact send. The user confirmed that: tapping the mic-looking button sent the message.

`hasText` only ever changes at the empty/non-empty boundary, so the handler now compares against a ref and calls setState only on the flip. Ordinary keystrokes hand React nothing at all — strictly less work than the transition they used to schedule, which was the reason the transition was there — and the one update that matters is urgent.

Every path that changes the text funnels through `onChangeText`, including the imperative `setTextAndSelection` on both the native and web `MultiTextInput` (autocomplete insert, Escape, and SessionView's post-send clear), so the ref cannot drift out of step with the field.

### DIFF
--- a/packages/happy-app/sources/components/AgentInput.tsx
+++ b/packages/happy-app/sources/components/AgentInput.tsx
@@ -701,10 +701,12 @@ export const AgentInput = React.memo(React.forwardRef<MultiTextInputHandle, Agen
     const activeSendIconColor = compactMobileComposer ? theme.colors.text : theme.colors.button.primary.tint;
     const isSendBlocked = props.blockSend ?? false;
 
-    // `hasText` drives only the send-button appearance/enabled state. It's
-    // updated via startTransition from the keystroke handler so a busy reducer
-    // never blocks the next character from landing in the textarea.
+    // `hasText` drives only the send-button appearance/enabled state, and it
+    // only ever changes at the empty/non-empty boundary — so the keystroke
+    // handler compares against this ref and calls setState at most twice per
+    // message, instead of handing React an update per character.
     const [hasText, setHasText] = React.useState(() => props.initialValue.trim().length > 0);
+    const hasTextRef = React.useRef(hasText);
     const hasImages = (props.selectedImages?.length ?? 0) > 0;
     const hasComposerContent = hasText || hasImages;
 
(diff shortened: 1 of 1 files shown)


chphch:
@bra1nDump — after the August 22 batch, here are six more of the same shape, in one comment rather than six pings.

Each one repairs a regression I traced to a specific commit, each is still reproducible on `3cf9ce70f` (I re-checked every line today), and **all six merge clean as they stand** — no rebase needed.

| PR | Broken by | What it does on current `main` | Size |
|---|---|---|---|
| [#1610](https://github.com/slopus/happy/pull/1610) | `1677fadc` | FAB's `+` glyph is invisible — the commit made the native fill transparent under `MobileGlassSurface` but kept `color={theme.colors.fab.icon}` (`FAB.tsx:69,72`), so it paints near-white on near-white in light mode and near-black on near-black in dark. Same class as #1582 / #1598 / #1599; this is the one control that sweep missed. | 1 file, +10/-1 |
| [#1673](https://github.com/slopus/happy/pull/1673) | `514ef3f1` + `1072d782` | Composer shows a **mic while it holds text**. `514ef3f1` put `hasText` behind `React.startTransition` (`AgentInput.tsx:897`); `1072d782` then folded the mic into the mobile primary button. While an agent streams, the transition is starved, so the button draws a mic — sending still works, because the press handler reads live text. | 1 file, +17/-6 |
| [#1378](https://github.com/slopus/happy/pull/1378) | `514ef3f1` | Sibling of the above at `AgentInput.tsx:904`: `setInputState` is the only input to the slash/`@` autocomplete, so under the same starvation the menu silently stops appearing mid-session. | 1 file, +8/-3 |
| [#1382](https://github.com/slopus/happy/pull/1382) | `53850917` | Complex Mermaid diagrams render as an **oversized empty box**. The commit added `reportHeight()` but measures `document.body.scrollHeight` (`MermaidRenderer.tsx:146-149`), so a wide diagram reports the page height. Closes #1475 (kemier, open and uncommented since Jul 3). | 1 file, +33/-20 |
| [#1451](https://github.com/slopus/happy/pull/1451) | `98c6ad9e` + `971d6089` | A selection card folds out of view, so the user **cannot answer a prompt they cannot see**. `98c6ad9e` replaced the "text always passes through as a standalone `TextItem`" path with `collectAgentWorkGroups`; `971d6089` made collapse the default. `useGroupedMessages.ts` has no selection-awareness today. Same surface as #1592. | 2 files, +257/-21 |
| [#1636](https://github.com/slopus/happy/pull/1636) | `53850917` + `b042d834` | Aborting a turn **silently re-models a running session**: `53850917` clears `permissionMode`/`modelMode`/`effortLevel` on abort (`SessionView.tsx:842`), and `b042d834` made `resolveMessageModeMeta` fall back live to `agentDefaultOverrides` (`messageMeta.ts:54,63`). Merging #1606 as `517e7492` removed the mask that made this hard to reproduce on main. | 4 files, +143/-5 |

Two things I'd rather say up front than have you find:

- **#1451 is the odd one out on size** (+257). Most of that is splitting the work-group fold around an interleaved selection. If you'd rather not take that, say so and I'll cut it to a narrow carve-out that only exempts the selection card and leaves `collectAgentWorkGroups` alone.
- **#1378 and #1382 are old** (74 and 73 days). I'm not claiming age as an argument — only that I re-verified both defects against `3cf9ce70f` today rather than trusting the original write-up.

No rush on any of these; I know the queue is long. Happy to split, shrink, or close any of them if a different shape suits you better.

PR #1702, opened 2026-08-16
feat(cli): add Hermes agent backend via ACP

im47cn:
## Summary

Adds the [Hermes Agent](https://github.com/) CLI as a first-class Happy backend by following the generic ACP runner principles from `docs/plans/generic-acp-runner.md` — no vendor-specific runner, no credential resolution:

- **`KNOWN_ACP_AGENTS`**: `hermes` → `{ command: 'hermes', args: ['acp'] }`
- **`factories/hermes.ts`**: `createHermesBackend` over `AcpBackend` with `HermesTransport`; Hermes manages its own auth (native login), so the user shell env is inherited untouched
- **`HermesTransport`**: `DefaultTransport` + tool-name patterns for `change_title` / `save_memory` / `think` / `HermesReasoning`, plus `hasChangeTitleInstruction` for title-change auto-approval (same shape as the Gemini factory)
- **`happy hermes`** routes through the existing generic `runAcp` runner with session flavor `hermes`
- **Daemon remote-spawn**: agent command mapping + CLI availability detection
- **`cliAvailability` metadata**: include `agy` (was reported by the daemon but stripped by the zod schema) and optional `hermes`/`crush` fields for older-CLI compatibility

## Testing

- Unit tests: config resolution, factory registration round-trip, transport pattern extraction (14 new assertions, full suite green)
- **End-to-end against Hermes Agent v0.20.1** (`hermes acp`): session creation, ACP handshake, 9 slash commands, **3 permission modes** (default / accept_edits / dont_ask), and **95 models** reported through the standard ACP config events — the mobile app gets mode switching and model selection natively with zero Hermes-specific code.

```console
$ happy hermes --verbose
[12:21] Happy Session ID: cmsvry9v2321dye0utwqz6i3j
[12:23] Outgoing slash commands from hermes (9): ...
[12:23] Outgoing modes from hermes (3), current=default: ...
[12:23] Outgoing models from hermes (95), current=deepseek:deepseek-v4-flash: ...
```

## Notes

- The `crush` entries in the shared type unions (`AgentId`, `cliAvailability`, daemon spawn switch) are forward declarations for the follow-up Crush backend PR stacked on this branch.

### DIFF
--- a/packages/happy-cli/src/agent/acp/acpAgentConfig.test.ts
+++ b/packages/happy-cli/src/agent/acp/acpAgentConfig.test.ts
@@ -2,10 +2,11 @@ import { describe, expect, it } from 'vitest';
 import { KNOWN_ACP_AGENTS, resolveAcpAgentConfig } from './acpAgentConfig';
 
 describe('KNOWN_ACP_AGENTS', () => {
-  it('defines built-in Gemini and OpenCode command mappings', () => {
+  it('defines built-in Gemini, OpenCode, and Hermes command mappings', () => {
     expect(KNOWN_ACP_AGENTS).toEqual({
       gemini: { command: 'gemini', args: ['--experimental-acp'] },
       opencode: { command: 'opencode', args: ['acp'] },
+      hermes: { command: 'hermes', args: ['acp'] },
     });
   });
 });
@@ -24,6 +25,14 @@ describe('resolveAcpAgentConfig', () => {
       agentName: 'opencode',
       command: 'opencode',
       args: ['acp', '--foo'],
+    });
+  });
+
+  it('resolves hermes to the hermes CLI with acp mode', () => {
+    expect(resolveAcpAgentConfig(['hermes'])).toEqual({
+      agentName: 'hermes',
+      command: 'hermes',
+      args: ['acp'],
     });
   });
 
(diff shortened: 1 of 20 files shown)


im47cn:
Thanks for the review — all findings addressed in cea5c163:

- **P1 app-side agent catalog**: landed in the stacked follow-up PR #1703 (mobile picker, spawn RPC union, availability schema) — this PR deliberately keeps to the CLI side so it can merge independently.
- **P1 `--happy-starting-mode` leak**: the ACP arg parser now consumes the flag (and its value) so it never reaches the agent subprocess.
- **P2 tmux spawns**: the tmux branch now derives the command from `options.agent ?? 'claude'` instead of the hardcoded ternary, so hermes/crush (and any future agent) map correctly.
- **P2 control-server schema**: `/spawn-session` zod enum accepts hermes and crush.
- **P2 Hermes transport unused by the real path**: `runAcp` now selects `hermesTransport` for hermes sessions.
- **P2 permission mode keys**: `MessageMetaSchema.permissionMode` widened to string (ACP agents report dynamic mode ids like `accept_edits`); the Claude path guards with a new `isPermissionMode` type guard so unknown ids are ignored rather than dropping the whole user message.
- **P2 determineToolName**: overridden in HermesTransport, resolving generic kinds via tool-call-id patterns (the path `requestPermission` actually calls).

End-to-end re-verified against Hermes Agent v0.20.1 after the fixes (3 modes, 95 models, session reaches idle).

im47cn:
Hi maintainers 👋 — friendly nudge on this one when you have a moment. The automated review findings have all been addressed (see [my summary above](https://github.com/slopus/happy/pull/1702#issuecomment-5312840429)), and the Hermes backend has been verified end-to-end against Hermes Agent v0.20.1. Happy to make any further changes if needed. There are also two follow-up PRs stacked on this branch (#1703 for the Crush backend + mobile picker, #1704 for a small server fix) if you'd like to review them together.

PR #1703, opened 2026-08-16
feat: add Crush agent backend (crush server HTTP API) + mobile agent picker

im47cn:
## Summary

Stacked on #1702 (Hermes). Adds [Crush](https://github.com/charmbracelet/crush) as a Happy backend. Crush does not speak ACP — it exposes a local HTTP + SSE API via `crush server` — so this PR adds one custom `AgentBackend` implementation and a small generalization of the generic runner:

### CLI (`packages/happy-cli`)

- **`CrushServerBackend`**: spawns `crush server` on a per-session Unix socket, creates workspace/session, and maps the `/v1` SSE event envelope to `AgentMessage`s:
  - assistant `message` events carry cumulative text parts → streamed as `textDelta` diffs
  - `tool_call` / `tool_result` parts → `tool-call` / `tool-result`
  - `run_complete` / `agent_event` → turn lifecycle statuses (incl. error details)
  - `permission_request` → `permission-request` (full request preserved as payload)
- **Permission bridge**: `runAcp` gains an `externalPermissions` mode that surfaces pushed permission requests to the mobile app and forwards responses to `POST /permissions/grant` with the `{ permission, action }` body Crush requires
- **`runAcp` generalization**: a new optional `createBackend` factory lets non-ACP backends ride the exact same runner pipeline (session setup, message queue, SessionEnvelope mapping, keepalive, abort/kill RPC). ACP-specific session-config switching no-ops when the backend omits those methods. `happy acp` behavior is unchanged
- **`happy crush`** subcommand with session flavor `crush`; `happy connect hermes|crush` explain that these agents manage their own auth

### Mobile (`packages/happy-app`)

- Hermes and Crush in the new-session agent picker, home dock, session-draft persistence, and spawn RPC union; machines only offer agents their `cliAvailability` reports as installed (new fields optional, older CLIs unaffected)
- Placeholder badge icons — happy to swap in proper brand assets if maintainers prefer

## Wire-protocol notes

Implementation was probed live against **crush v0.80.0** before writing the mapper: all endpoints live under `/v1`, requests require a UUID `client_id`, SSE events use a two-layer envelope (`{type, payload:{type, payload}}`), assistant message events carry cumulative parts, and run completion is signaled by `run_complete` + `agent_event` (`agent_finished`).

## Testing

- Unit tests: SSE envelope → AgentMessage mapping (streaming diffs, tool parts, lifecycle, permissions), factory registration (full suite green: happy-cli 810 unit, happy-app 909)
- **End-to-end against crush v0.80.0**: workspace/session creation, prompt delivery (`202 Accepted`), streamed assistant text, and run completion all verified live

```console
$ happy crush --verbose
[13:04] Happy Session ID: cmsvtgtb05j2uxw0uom0hxxvh
[13:04] Status: starting
[13:04] Status: idle
```

## Design decisions (vs. a naive port)

- SSE `AbortSignal` is actually wired to the stream teardown
- Typed event parsing — no `any` casts
- Windows fails fast with a clear error instead of silently targeting port 0
- One shared runner instead of a copy-pasted per-agent runner


im47cn:
Thanks for the review — findings addressed:

- **tmux spawns / `--happy-starting-mode` leak**: fixed in the base PR #1702 (cea5c163).
- **Duplicate cumulative tool parts**: stream state now tracks emitted `tool_call`/`tool_result` part ids, so each is emitted exactly once across cumulative message updates (regression test added).
- **fullText dropped by the session manager**: `AcpSessionManager.mapMessage` now treats a `fullText`-only model output as authoritative and replaces the accumulated output buffer (test added).
- **Windows crush availability**: detectCLI no longer reports crush on win32 (the backend requires Unix domain sockets).
- **Mode defaults for the new agents**: hermes/crush get their own agent-default entries (default permission, no hardcoded model/effort) instead of falling through to Claude defaults; the hardcoded model/permission pickers offer the minimal generic set until the agents report their dynamic lists; settings labels cover both agents.

Re-verified end-to-end after fixes: `happy hermes` (modes/models reported, idle) and `happy crush` (workspace/session/SSE, idle).

Issue #1721, opened 2026-08-25
Claude 1M-context models are unreachable from Happy: picker has no `[1m]` entry, and DEFAULT_CLAUDE_MODEL = 'opus' pins every remote turn to 200K


tonsiasy:
## Summary

There is no way to run a Claude session at 1M context through Happy. Two independent causes, both of which have to be fixed for the user-visible problem to go away:

1. The app's Claude model list is hardcoded to three keys, none of them a `[1m]` variant, and the "Custom model…" escape hatch is codex-only.
2. happy-cli defaults `currentModel` to the literal string `'opus'` when no `--model` is given, so every remote turn is pinned to the 200K model — overriding whatever the user configured in `~/.claude/settings.json`, and overriding the model a locally-started Claude Code was already running.

Claude Code itself accepts `claude-opus-5[1m]` / `claude-sonnet-5[1m]` (and the short `opus[1m]` / `sonnet[1m]` aliases); nothing downstream of Happy is blocking this.

## Environment

| | |
|---|---|
| happy-cli | 1.2.0 (npm) |
| Claude Code | 2.1.241 |
| OS | macOS, darwin 24.6.0 |
| Source refs | `slopus/happy` @ `8767b2bccff3` (main, 2026-08-25) |

## Impact

Picking "opus 5" on the phone silently gives you the 200K model. On a long-running session driven from mobile — which is Happy's whole point — you hit compaction far earlier than you would in the same session driven from the terminal, with nothing in the UI indicating that the context window differs.

## Cause 1 — the app's Claude model list has no 1M variants

`packages/happy-app/sources/components/modelModeOptions.ts`:

```ts
export function getClaudeModelModes(): ModelMode[] {
    return [
        { key: 'claude-fable-5', name: 'fable 5', description: null },
        { key: 'claude-opus-5', name: 'opus 5', description: null },
        { key: 'claude-sonnet-5', name: 'sonnet 5', description: null },
    ];
}
```

The key is passed to happy-cli verbatim (`packages/happy-cli/src/daemon/spawnModeArgs.ts` pushes `--model <key>`; mid-session changes travel as `meta.model` into the SDK adapter's `QueryOptions.model`), so "opus 5" reaches Claude Code as exactly `claude-opus-5` → 200K.

There is no workaround inside the app:

- `includeConfiguredModel()` in the same file early-returns unless `flavor === 'codex'`, so a persisted non-listed key is not even surfaced for Claude.
[...]

PR #1724, opened 2026-08-25
test(app): expect the Opus 5 1M row the model list actually returns

chphch:
`packages/happy-app`'s unit tests have been red since ac68e914 (`feat(app): add Opus 5 1M and polish changelog`). That commit added a fourth row to `getClaudeModelModes()`:

```ts
{ key: 'claude-opus-5[1m]', name: 'Opus 5 [1M]', description: '1M context' },
```

but the assertion in `modelModeOptions.test.ts` that enumerates the list still names three. The test file's blob is byte-identical at `ac68e914^`, `ac68e914` and `main`, so it was never updated — before or since.

## Reproduce

On `main` (b16b23c5b), nothing else changed:

```
$ cd packages/happy-app && pnpm exec vitest run

 FAIL  sources/components/modelModeOptions.test.ts > modelModeOptions > only offers the current-generation claude models
AssertionError: expected [ 'claude-fable-5', …(3) ] to deeply equal [ 'claude-fable-5', …(2) ]
+   "claude-opus-5[1m]"

 Test Files  1 failed | 102 passed (103)
      Tests  1 failed | 1090 passed (1091)
```

With this patch:

```
 Test Files  103 passed (103)
      Tests  1091 passed (1091)
```

It is the only failing test on `main`.

## Why add the row rather than relax the assertion

The assertion exists to keep alias keys and a `default` row out of the Claude picker — `opus`/`sonnet` resolve to older models than the row would claim. `claude-opus-5[1m]` is a full model ID rather than an alias, and ac68e914's own comment says the `[1m]` suffix is honored by the CLI (#1721), so the row belongs in the list and it was the expectation that went stale. The `default`/alias assertions below it are untouched.

## Why CI didn't catch it

No workflow runs `happy-app`'s tests. `typecheck.yml` runs `pnpm --filter happy-app typecheck` only, and the repo's single test step is `pnpm --filter happy-server … test` in `server.yml` — so a red vitest in the app package is invisible to Actions. Happy to send a follow-up PR adding an app-test job if you'd like one; kept out of here so this stays a two-line change.

### DIFF
--- a/packages/happy-app/sources/components/modelModeOptions.test.ts
+++ b/packages/happy-app/sources/components/modelModeOptions.test.ts
@@ -130,11 +130,13 @@ describe('modelModeOptions', () => {
         expect(models.map((model) => model.key)).toEqual([
             'claude-fable-5',
             'claude-opus-5',
+            'claude-opus-5[1m]',
             'claude-sonnet-5',
         ]);
         expect(models.map((model) => model.name)).toEqual([
             'Fable 5',
             'Opus 5',
+            'Opus 5 [1M]',
             'Sonnet 5',
         ]);
         // No `default model` row, and no alias keys: an alias would silently


chphch:
Closing as already fixed — a755c2ad1 landed the same two-line change on main, and the assertion passes there now (modelModeOptions.test.ts, 32/32 on 7e63b45a). Nothing left for this PR to do.

PR #1738, opened 2026-08-31
fix(cli): preserve daemon ownership across health check races

### LINKED ISSUE #1654: [Bug]: Daemon-spawned sessions can falsely replace a healthy daemon after a 2s control health timeout
## Summary

On Windows, a session spawned by the Happy daemon can falsely decide that its still-running parent daemon is stale when the daemon control endpoint does not respond within 2 seconds.

The child session then deletes `daemon.state.json` and starts a replacement daemon. The old and new daemons subsequently observe each other's PID in the shared state file and shut themselves down. After the last daemon exits, the machine stops sending heartbeats and appears offline in the Happy app.

This was observed twice in succession: first while spawning a remote Claude session, then while spawning a remote Codex session.

## Environment

- Happy CLI: `1.2.0` (current npm `latest`)
- Platform: Windows, `win32 x64`
- Node.js: `v24.13.0`
- Server: default production server
- Authentication: valid; the daemon connected successfully and started its 20-second keep-alive before the failure

## Business flow that triggers the bug

`spawn-happy-session`
→ daemon launches `happy claude` or `happy codex` with `--started-by daemon`
→ the child session unconditionally calls `ensureDaemonRunning()`
→ `checkIfDaemonRunningAndCleanupStaleState()` probes the parent daemon's local `POST /list` endpoint
→ one 2-second timeout is treated as proof that the state is stale
→ the child deletes `daemon.state.json` and starts another daemon
→ daemon ownership conflict causes both daemons to exit
→ no daemon remains, so the machine becomes offline

## Observed timeline

The following entries are from local Happy logs with personal paths and machine identifiers removed.

### First occurrence: remote Claude session

Parent daemon:

```text
[21:14:09.143] [DAEMON RUN] Spawned process for a remote Claude session
[21:14:09.146] [DAEMON RUN] Waiting for session webhook
```

Daemon-spawned child session:

```text
[21:14:24.087] Ensuring Happy background service is running & matches our version...
[21:14:26.138] [DAEMON RUN] PID <parent> is alive but HTTP health check failed on port <port>, cleaning up stale state
[21:14:26.156] [DAEMON RUN] Daemon state file removed
[21:14:26.166] Starting Happy background service...
```

The original daemon processed the control request later:

```text
[21:14:30.091] [CONTROL SERVER] Listing 2 sessions
```

This shows that the original daemon was alive and its control server eventually handled the request. The 2-second health check produced a false negative.

After the replacement daemon wrote its state, the original daemon exited:

```text
[21:15:36.342] [DAEMON RUN] Somehow a different daemon was started without killing us. We should kill ourselves.
[21:15:36.828] [DAEMON RUN] Daemon state file removed
[21:15:36.840] [DAEMON RUN] Process exiting with code: 0
```

### Second occurrence: remote Codex session

The replacement daemon later spawned a Codex session. The same sequence repeated:

```text
[21:18:02.197] Ensuring Happy background service is running & matches our version...
[21:18:04.242] [DAEMON RUN] PID <parent> is alive but HTTP health check failed on port <port>, cleaning up stale state
[21:18:04.248] [DAEMON RUN] Daemon state file removed
[21:18:04.252] Starting Happy background service...
```

The final replacement daemon connected successfully and started keep-alive, but then also exited after detecting another PID in the shared state:

```text
[21:18:21.992] [API MACHINE] Connected to server
[21:18:31.710] [API MACHINE] Keep-alive started (20s interval)
[21:19:20.713] [DAEMON RUN] Somehow a different daemon was started without killing us. We should kill ourselves.
[21:19:21.157] [DAEMON RUN] Daemon state file removed
[21:19:21.163] [DAEMON RUN] Process exiting with code: 0
```

At that point, `happy daemon status` reported:

```text
Daemon is not running
```

and `daemon.state.json` no longer existed. A daemon-spawned Codex session remained alive as an orphan, but it did not maintain machine presence.

### 1. Daemon-spawned Codex sessions still ensure the daemon

`handleCodexCommand()` parses `--started-by daemon`, but calls `ensureDaemonRunning()` unconditionally:

https://github.com/slopus/happy/blob/main/packages/happy-cli/src/commands/codexCommand.ts

The equivalent startup behavior was also observed for a daemon-spawned Claude session.

### 2. A single 2-second HTTP timeout deletes daemon state

`checkIfDaemonRunningAndCleanupStaleState()` first proves the PID is alive, then performs `POST /list` with:

```ts
signal: AbortSignal.timeout(2000)
```

Any exception immediately calls `cleanupDaemonState()`:

https://github.com/slopus/happy/blob/main/packages/happy-cli/src/daemon/controlClient.ts

A transient timeout is therefore treated the same as a reused PID or a dead daemon.

### 3. Ownership-conflict shutdown still writes state

In the daemon heartbeat, detecting another PID calls `requestShutdown()`, but execution continues into `writeDaemonState(updatedState)`:

https://github.com/slopus/happy/blob/main/packages/happy-cli/src/daemon/run.ts

This can write the old daemon's PID back during shutdown, causing the replacement daemon to detect the reverse conflict on its next heartbeat and exit as well.

## Expected behavior

- A session launched with `--started-by daemon` should not attempt to replace the daemon that just spawned it.
- A live PID plus a single transient HTTP timeout should not be enough to delete daemon ownership state.
- Once an ownership conflict requests shutdown, the losing daemon should not write its PID back into the shared state file.
- Starting a remote Claude or Codex session should not make the machine go offline.

## Actual behavior

A transient control-server timeout causes a daemon-spawned child session to replace its live parent daemon. The resulting state-file race shuts down all daemons and leaves the machine offline.

## Suggested fixes

Any combination of the following would prevent this failure:

1. Skip `ensureDaemonRunning()` when `startedBy === 'daemon'`.
2. Retry the local HTTP health check with backoff before clearing state.
3. Distinguish timeout from `ECONNREFUSED`; do not treat every exception as proof of PID reuse.
4. On Windows, validate the process command line or use a daemon-specific ownership token instead of relying on one HTTP request.
5. Return immediately after requesting shutdown because another daemon owns the state; do not execute the heartbeat state write.
6. Add an integration test where a daemon-spawned session starts while `/list` responds after more than 2 seconds, and assert that no replacement daemon is started.

## Regression context

The behavior appears to result from the interaction of:

- #975, which added unconditional daemon ensure/start behavior to Codex session startup.
- #809, which added the 2-second local HTTP check and stale-state cleanup for Windows PID reuse.

Both changes are individually reasonable, but daemon-spawned child sessions plus a transiently slow control endpoint expose a destructive race.

myartings:
## Summary
- Skip daemon ensure/replacement for sessions that were started by the daemon.
- Preserve live daemon state after a transient control timeout while retaining stale-PID cleanup for non-timeout failures.
- Prevent a losing daemon from rewriting heartbeat state and scope state/lock cleanup to the current owner.

Fixes #1654

## Testing
- Focused unit tests: pnpm --filter happy exec vitest run --project unit src/daemon/ensureDaemonRunning.test.ts src/daemon/controlClient.test.ts src/persistence.test.ts src/commands/codexCommand.test.ts (25 passed)
- pnpm --filter happy typecheck
- pnpm --filter happy build

## Notes
- The authenticated daemon integration test was not run because it stops and restarts the local daemon.
- The full Windows unit suite reported 794 passed and 34 failures across six untouched platform/tooling-sensitive test files.

### DIFF
--- a/packages/happy-cli/src/commands/codexCommand.test.ts
+++ b/packages/happy-cli/src/commands/codexCommand.test.ts
@@ -51,7 +51,7 @@ describe('handleCodexCommand', () => {
   it('ensures the daemon is running before starting a codex session', async () => {
     await handleCodexCommand(['--started-by', 'terminal'])
 
-    expect(mocks.mockEnsureDaemonRunning).toHaveBeenCalledTimes(1)
+    expect(mocks.mockEnsureDaemonRunning).toHaveBeenCalledWith('terminal')
     expect(mocks.mockRunCodex).toHaveBeenCalledWith({
       credentials: { token: 'token' },
       startedBy: 'terminal',
@@ -78,6 +78,7 @@ describe('handleCodexCommand', () => {
 
     await handleCodexCommand(['--no-sandbox', '--resume', 'thread-123', '--started-by', 'daemon'])
 
+    expect(mocks.mockEnsureDaemonRunning).toHaveBeenCalledWith('daemon')
     expect(mocks.mockRunCodex).toHaveBeenCalledWith({
       credentials: { token: 'token' },
       startedBy: 'daemon',
--- a/packages/happy-cli/src/commands/codexCommand.ts
+++ b/packages/happy-cli/src/commands/codexCommand.ts
@@ -29,7 +29,7 @@ export async function handleCodexCommand(args: string[]): Promise<void> {
   }
 
   const { credentials } = await authAndSetupMachineIfNeeded()
-  await ensureDaemonRunning()
+  await ensureDaemonRunning(startedBy)
 
   await runCodex({
     credentials,
--- a/packages/happy-cli/src/daemon/controlClient.test.ts
+++ b/packages/happy-cli/src/daemon/controlClient.test.ts
@@ -148,24 +148,72 @@ describe('checkIfDaemonRunningAndCleanupStaleState', () => {
     )
   })
 
-  it('refuses to stop a legacy snapshot that has no generation token', async () => {
+  it('gracefully stops a live legacy owner after stable HTTP identity evidence', async () => {
     vi.useFakeTimers()
-    const fetchMock = vi.fn().mockResolvedValue({
-      ok: true,
-      json: vi.fn().mockResolvedValue({ status: 'stopping' }),
-    })
+    const fetchMock = vi.fn()
+      .mockResolvedValueOnce({
+        ok: true,
+        json: vi.fn().mockResolvedValue({ children: [] }),
+      })
+      .mockResolvedValueOnce({
+        ok: true,
+        json: vi.fn().mockResolvedValue({ status: 'stopping' }),
+      })
     vi.stubGlobal('fetch', fetchMock)
-
-    const stopPromise = stopDaemon({
+    const legacyState = {
       pid: 1111,
       httpPort: 4111,
       startTime: 'legacy',
       startedWithCliVersion: '1.0.0',
-    })
+    }
+    mocks.mockReadDaemonState.mockResolvedValue(legacyState)
+
+    const stopPromise = stopDaemon(legacyState)
     await vi.advanceTimersByTimeAsync(2100)
     await stopPromise
 
+    expect(fetchMock).toHaveBeenNthCalledWith(1, 'http://127.0.0.1:4111/list', expect.anything())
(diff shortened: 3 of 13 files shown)


myartings:
Hi maintainers — the CLI Smoke Test workflow for this fork PR is currently blocked with action_required before any jobs are created: https://github.com/slopus/happy/actions/runs/33344371435

Could someone approve the workflow run when convenient? Thank you!

Issue #914, opened 2026-03-25
iOS App: reply body text never displayed (title updates work, mobile Safari works)


MILKMILKKING:
## Summary

iPhone Happy App shows conversation titles updating correctly but **never renders the reply body text**. The same account on mobile Safari (`app.happy.engineering`) displays replies normally — confirming the issue is in the iOS App's message rendering layer, not the backend.

## Environment

- Happy CLI: `1.1.0-2`
- Mac: Mac mini, macOS Darwin arm64
- iPhone: iOS (latest)
- Mac hostname: `lyrikdeMac-mini.local`

## Steps to Reproduce

1. Authenticate and start daemon on Mac (`happy daemon start`)
2. Open Happy iOS App, connect to the Mac device
3. Start a new conversation from the iPhone App
4. Send any message (e.g., "test")
5. Observe: title updates but **reply body is not displayed**
6. Open `https://app.happy.engineering` in **mobile Safari on the same iPhone**
7. Send the same message → reply body displays correctly

## Expected Behavior

Reply body text should be displayed in the iOS App, same as it appears in the web interface.

## Actual Behavior

- Title/metadata updates: **visible** in iOS App
- Reply body text: **never displayed** in iOS App
- Push notifications: received successfully
- Mobile Safari (same device, same account): **works correctly**

## Diagnostic Evidence

Mac-side logs confirm the full pipeline succeeds every time:

```
[10:39:43] Title changed (happyMCP title update succeeded)
[10:39:44] Final answer generated (phase: final_answer)
[10:39:45] Push notifications sent successfully (1 token)
[10:39:46] new-message (t: encrypted) seq 3-5 sent via socket
[10:39:47] ev.text + turn-end + ready events all delivered to server
```

No errors, disconnects, or failures in any daemon or session log.

## What We've Already Ruled Out

| Hypothesis | Action Taken | Result |
|---|---|---|
| Stale daemon / zombie processes | Killed 87 old processes, clean restart | Same issue |
| Credential / machine ID corruption | `happy auth login --force` x3 (3 different machine IDs) | Same issue |
| Old `happy-coder` vs new `happy` conflict | Removed all old version processes | Same issue |
| `onboardingCompleted: false` | Fixed to `true` | Same issue |
| WebSocket disconnection | Clean daemon, verified stable connection | Same issue |
| Server-side message delivery | Logs show encrypted messages + events fully delivered | Confirmed working |
[...]

PR #1750, opened 2026-09-04
fix(cli): trim prompts before appPrompt dedupe match

xiaojingyu812:
## Problem

In remote (daemon-spawned) sessions, every prompt sent from the app is persisted **twice** on the server: once by the app itself, and once again by the remote-mode JSONL scanner re-forwarding it.

Observed on live sessions (see my comment on #914 for the full investigation): decrypting the stored history shows two consecutive `role: 'user'` messages with identical text — e.g. `seq=9 "回复ok"` (sentFrom: ios) and `seq=14 "回复ok"` (sentFrom: cli).

## Root cause

`runClaude.ts` dedupes app-sent prompts against the scanner by **exact** content match:

- `recordAppPrompt()` receives the app-delivered text: `"回复ok"`
- `consumeAppPrompt()` receives the JSONL-echoed text the SDK wrote: `"回复ok\n"` (trailing newline)

The exact match misses, the scanner treats the prompt as terminal-typed, and forwards it again.

## Fix

Extract the ring buffer into `appPromptDedupe.ts` (behavior otherwise unchanged: 5-minute window, consume-once semantics, stale entries roll off) and **trim on both record and consume** so the comparison ignores the SDK's trailing newline. The existing empty-line guard in the scanner already uses trim semantics, so this is consistent with how "same prompt" is defined there.

## Tests

Six unit tests in `appPromptDedupe.test.ts`, including the trailing-newline regression case and the 5-minute expiry window (`vitest run src/claude/appPromptDedupe.test.ts` → 6/6 passing).

### DIFF
--- a/packages/happy-cli/src/claude/appPromptDedupe.test.ts
+++ b/packages/happy-cli/src/claude/appPromptDedupe.test.ts
@@ -1,0 +1,56 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { createAppPromptDedupe } from './appPromptDedupe';
+
+describe('createAppPromptDedupe', () => {
+    afterEach(() => {
+        vi.useRealTimers();
+    });
+
+    it('consumes an exact match exactly once', () => {
+        const dedupe = createAppPromptDedupe();
+        dedupe.record('hello');
+        expect(dedupe.consume('hello')).toBe(true);
+        // Consumed entries are removed, so an identical terminal-typed
+        // prompt still gets forwarded afterwards.
+        expect(dedupe.consume('hello')).toBe(false);
+    });
+
+    it('matches when the SDK echoes the prompt with a trailing newline', () => {
+        // Regression: the app sends "fix it" and the JSONL scanner sees
+        // "fix it\n" — the exact-match dedupe used to miss this and the
+        // app-sent prompt was persisted twice on the server.
+        const dedupe = createAppPromptDedupe();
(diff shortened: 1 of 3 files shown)


xiaojingyu812:
Gentle bump — this has been open for ~11 days with no review comments, and I noticed CI never actually ran on it (`status: pending`, 0 checks reported), which is why the PR shows `mergeable_state: unstable`. My guess is the workflow needs a maintainer's approval before it will run for a fork PR.

If someone could approve the workflow run, the six vitest cases should execute and confirm it goes green (`vitest run src/claude/appPromptDedupe.test.ts`).

Summary for whoever picks this up:

- In remote (daemon-spawned) sessions every app-sent prompt is persisted twice. `recordAppPrompt()` receives the text as delivered by the app (`"回复ok"`), while `consumeAppPrompt()` receives the JSONL-scanner echo the SDK wrote (`"回复ok\n"`). The exact-content match misses, so the scanner treats the prompt as terminal-typed and forwards it again.
- Fix: extract the ring buffer into `appPromptDedupe.ts` and trim on both record and consume. Buffer semantics (5-minute window, consume-once, stale entries rolling off) are otherwise unchanged, and this matches the trim semantics the scanner's existing empty-line guard already uses.
- Six unit tests included, among them the trailing-newline regression case and the 5-minute expiry window — 6/6 passing locally.

Happy to rebase or adjust anything reviewers prefer.

Issue #1765, opened 2026-09-07
Web client: main thread saturated by synchronous JS (native mobile instant on same account/server)


f-liva:
### Summary

On the **web client**, the UI becomes very sluggish (session list navigation and opening a session), while the **native mobile app on the same account and the same self-hosted server is instant**. A Chrome Performance profile shows the main thread is saturated by synchronous JS, not by rendering, network, or the server.

### Environment

- Self-hosted server (`happy-server`), web client served over HTTPS.
- Chrome (latest) on Windows desktop.
- Native mobile app: same account, same server → **fast**. So the bottleneck is client-side in the web build, not the backend or the link.

### Profile (Chrome DevTools → Performance, ~36 s capture: session-list scroll + opening one session)

Summary:
- **Scripting 23,374 ms**
- Rendering 111 ms
- Painting 79 ms
- System 4,283 ms
- Total 36,350 ms

→ Rendering/painting are negligible; it's ~64% Scripting. The main thread is blocked by JS, so the tab (and even DevTools itself) stutters.

Bottom-Up (Self time):
- `evaluate` — **16,039 ms (61.3%)**
- Major GC — 1,312 ms (5.0%)  ← heavy allocation churn
- Function call `index-0ed1099…js:1026:546` — 707 ms self / **6,439 ms total (24.4%)**  (app hot path, minified)
- `nextNode` — 699 ms

The combination of a dominant `evaluate` self-time plus significant Major GC points at large synchronous work with high allocation — consistent with decrypting/parsing many sessions and re-deriving/re-sorting large in-memory structures on the main thread.

### Ruling things out

- **Extensions are not the cause**: repeated in a clean **Incognito** window (extensions disabled) — identical slowness. In the profiled run, extension main-thread time was tiny (Bitwarden ~640 ms, everything else <100 ms), ~3% of Scripting.
- Not the server / network: the native app hits the same server and is instant.

### Likely related

- #1453 — background prefetch loads the entire session history into memory, degrading long sessions on web.
- #1285 — prefetch latest messages for non-archived sessions.
- #1170 — laggy cross-device session switching.

These describe specific prefetch paths; this report is about the **web client's overall main-thread saturation** (session list + session open), which may share the same root cause (whole datasets decrypted/parsed/sorted synchronously on the main thread).

### Questions

[...]

Issue #1453, opened 2026-06-27
Background history prefetch loads the entire session into memory, degrading long sessions on web


chaehyun2:
### Summary

After a session's first open, `prefetchOlderMessagesInBackground` ([sync.ts#L1885](https://github.com/slopus/happy/blob/main/packages/happy-app/sources/sync/sync.ts#L1885)) walks backward through **every** older page (100 msgs each, ~250 ms apart) until the whole history is in memory. `fetchInitialLatestPage` already loads the latest page for fast first paint, so this background loop is purely a prefetch.

### Impact (web)

For a long session this grows the in-memory message list without bound. Because the chat list re-derives its grouped items and the message map is re-sorted on updates, a fully-prefetched long session becomes progressively slower to re-enter and re-render — a page reload (which resets to the latest page only) temporarily restores responsiveness. Re-opening such a session from the list feels sluggish.

### Note

`loadOlderMessages` is already wired to `FlatList.onEndReached` (`ChatList.tsx`), so history loads lazily on scroll-up regardless of this prefetch.

The loop's only stop conditions are: no more older history (`hasMoreOlder` false), the session being unloaded, `seq <= 1`, the encryption key being gone, or an error in `loadOlderMessages` — there is no page/count bound.

### Question

Would you be open to bounding the background prefetch to a small page budget (keeping a scroll buffer ready while leaving the rest to on-demand `loadOlderMessages`), or making it configurable? Happy to send a PR if you'd like — I have a small change that's been working well for us.

Eiji-Kudo:
I can reproduce the same user-facing behavior in the native iOS app while using Happy to access a remote computer.

Opening a long conversation takes a very long time, and the client appears to load/decrypt the full conversation history instead of keeping the initial load bounded and fetching older messages only when needed. The primary symptom I can confirm is the long loading time; I am not reporting confirmed data loss.

This suggests the impact may not be limited to the web client. I can provide the app version, iOS/device version, approximate conversation size, and loading time if those measurements would help.

PR #1766, opened 2026-09-07
perf(sync): bound background history prefetch to a small page budget

f-liva:
## What

Bound the background history prefetch to a small page budget instead of pulling **all** older messages into memory.

## Why

On the web client, opening a session and returning to it becomes very sluggish on long sessions. A Chrome Performance profile (~36 s capture) shows the main thread saturated by synchronous JS, not rendering:

- Scripting **23,374 ms** / 36,350 ms (~64%); Rendering 111 ms, Painting 79 ms (negligible)
- Bottom-Up self time: `evaluate` **16,039 ms** (61%), **Major GC 1,312 ms** (heavy allocation churn)

Root cause is O(n²):

- `fetchOlderMessagesInBackground` (`sources/sync/sync.ts`) ran an unbounded `while (true)` loop pulling every older page for each visible session.
- Each applied page calls `applyMessages` (`sources/sync/storage.ts`), which rebuilds the whole map (`{ ...messagesMap }`) and re-sorts every message (`Object.values().sort()`) — O(n) **per page**.
- Unbounded pages × O(n)-per-page = O(n²) + constant Major GC → main-thread saturation on web (native is fine: native crypto, no JS `evaluate` bundle churn).

## Change

`fetchOlderMessagesInBackground`: replace the unbounded `while (true)` with a bounded `for` loop (`MAX_PAGES = 5`, ~600 messages of buffer). The loop still stops early on `!hasMoreOlder`, missing session/key, or `oldestSeq <= 1`. The rest of the history keeps loading on demand via `loadOlderMessages` (already wired to `FlatList.onEndReached`), so nothing is lost — it just isn't all decrypted/sorted up front.

Single-file change, `pnpm typecheck` clean.

## Verified

Built and deployed to a self-hosted instance; the open/return freeze on long sessions is gone in day-to-day web use.

## Related

- Refs #1765 (this profile / main-thread saturation report)
- Refs #1453 (background prefetch loads the entire history into memory)

## Follow-ups (not in this PR)

- `applyMessages` is still O(n) per call (immutable map copy + full re-sort); this PR reduces how often it is called rather than its per-call cost.
- Session list load (`fetchSessions`) decrypts all sessions' metadata + agentState synchronously on the main thread — a separate hotspot.


f-liva:
Superseded by #1797, which reimplements the same 5-page prefetch bound on current `main` (upstream refactored the prefetch loop into `fetchOlderMessagesInBackground` since this PR was written against the older tree).

PR #1774, opened 2026-09-08
fix(cli): keep externally-started sessions resumable after they exit

### LINKED ISSUE #1694: resume-happy-session always fails with "Session … is not tracked by this daemon" after the daemon did not start the session
**Environment**
- happy CLI (production install): `happy` npm package, dist `types-CV0guBiJ.mjs` (July 2026 build)
- macOS (mac mini, Apple Silicon), daemon long-running (pid 3710, started 2026-07-28 22:50 local)
- Client: happy Android app → machine RPC `resume-happy-session`

**What happened**

Resuming an existing session from the app failed 8 out of 8 attempts. Every attempt reached the daemon (routing fine) and was rejected by the handler:

```
[08:21:15.052] [RPC] Calling handler { method: 'fff77c54-b4fb-438a-a7d9-e32913587c1e:resume-happy-session' }
[08:21:15.056] [RPC] [ERROR] Error handling request {
  error: Error: Session cms4t96891ypywc0uryxdimcd is not tracked by this daemon. It may have been started before the daemon or on another machine.
      at file:///Users/bambo/.npm-global/lib/node_modules/happy/dist/types-CV0guBiJ.mjs:4652:21
      at async RpcHandlerManager.handleRequest (file:///Users/bambo/.npm-global/lib/node_modules/happy/dist/types-CV0guBiJ.mjs:832:22)
}
```

The identical error fired 8 times across the retry window (log timestamps 08:21–08:2x, same session id, same machine id). The session itself existed and was listed by the app; the machine was online and answering other RPCs.

**Expected**

`resume-happy-session` resumes a session that the server/account plainly knows about — or, if the daemon cannot resume sessions it didn't spawn, the app should not offer the action (or should fall back to spawning with `--resume <claudeSessionId>`).

**Actual**

The daemon only "tracks" sessions it spawned itself, in memory. For any session that predates the daemon process (started before a daemon restart, or started by a different client), resume is guaranteed to fail — 8/8 in our capture. The error message even names the scenario ("may have been started before the daemon") but nothing recovers from it.

**Why this matters**

Daemon restarts are routine (upgrades, crash recovery, machine reboot). After any restart, every pre-existing session becomes non-resumable from mobile — which is exactly the moment a user most wants to re-attach to ongoing work.

**Suggestions (any of)**

1. Persist the daemon's session registry (it already writes session files under `~/.happy`) and reload on startup, so "tracked" survives restarts.
2. On resume of an untracked session, fall back to spawning a new process with the agent's own resume flag (e.g. `claude --resume <id>`) using the session metadata the daemon can read from disk.
3. If neither is acceptable, surface the limitation to the app so the UI can hide/disable resume for untracked sessions instead of failing at call time.

**Notes**

- This capture is from our fork's instrumentation, but the failing code path is unmodified upstream code (`RpcHandlerManager.handleRequest` → session tracking check).
- Happy to provide the full log excerpt (8 occurrences) privately; the lines above are verbatim.

f-liva:
## Problem

Resuming a session from the app fails with `Session <id> is not tracked by this daemon. It may have been started before the daemon or on another machine.` whenever the session was started from a terminal (`happy`, `happy --resume ...`) instead of being spawned by the daemon. Fixes #1694.

## Cause

`packages/happy-cli/src/daemon/run.ts` keeps two maps: `pidToTrackedSession` (live) and `sessionIdToFinishedSession` (exited, kept so `resumeSession` can find the encryption data).

- Daemon-spawned sessions: the child `exit` handler calls `onChildExited`, which moves the entry to `sessionIdToFinishedSession`.
- Externally-started sessions: they are registered through the `/session-started` webhook (`Registered externally-started session`), but there is no child handle. When the process dies, the only thing that notices is the heartbeat prune loop, which does a bare `pidToTrackedSession.delete(pid)` and drops the encryption data.

`findTrackedSessionById` then returns nothing and `resumeSession` rejects. Restarting the daemon makes it work again because startup reloads the persisted sessions from disk, which is a good hint that the data was there all along.

Daemon log on the affected session (happy-cli 1.2.2):

```
[09:43:07.818] [DAEMON RUN] Session webhook: cmtsd4bw…, PID: 1131843, started by: terminal, hasEncryption: true
[09:43:07.820] [DAEMON RUN] Registered externally-started session cmtsd4bw…
[11:01:09.739] [DAEMON RUN] Removing stale session with PID 1131843 (process no longer exists)
[15:08:27.062] [RPC] [ERROR] Error handling request { error: Error: Session cmtsd4bw… is not tracked by this daemon. … }
```

## Fix

Route the prune loop through `onChildExited(pid)` so both kinds of session are preserved the same way. One-line change, no behaviour change for daemon-spawned sessions (`onChildExited` is idempotent on the map).

## Verification

- `tsc --noEmit` clean for `run.ts`.
- Reproduced on a self-hosted setup: terminal-started session, killed from the app, resume fails with the error above; after a daemon restart (which reloads persisted sessions) the same resume succeeds, confirming that only the in-memory bookkeeping was lost.


f-liva:
This is now obsolete: #1779 landed the same fix — the heartbeat prune routes a dead PID through `onChildExited` instead of dropping the tracked session — so externally-started sessions already survive on current `main`. Nothing left to rebase here; happy to close this unless you'd rather keep it as the reference for #1694.

f-liva:
Closing: #1779 landed the same fix upstream, so there is nothing left to merge here. Issue #1694 stays covered by that PR.

PR #1776, opened 2026-09-09
perf(hooks): drive useElapsedTime from a single shared ticker

f-liva:
## What

Drive `useElapsedTime` from a single shared 1-second ticker instead of one `setInterval` per hook instance.

## Why

`useElapsedTime` (`sources/hooks/useElapsedTime.ts`) starts its own `setInterval(1000)` on mount. It is used per rendered tool call (`ToolView.tsx`), so a chat with many tool calls runs **dozens of unsynchronized timers**, each firing a `setState` on its own sub-second offset. The result is re-renders scattered across every frame, continuously, for as long as the page stays open.

On the web client this is a measurable, **growing** main-thread cost the longer a session is left open — the hook's interval callback was one of the larger app-side self-time entries in a Performance profile of a long-lived tab, and it grows with the number of mounted elapsed-time consumers.

## Change

- One module-level `setInterval` shared by all consumers via `subscribeToTick`. It starts when the first consumer subscribes and is cleared when the last unmounts.
- All consumers fire on the **same aligned tick**, so React can batch their updates into a single render pass instead of N scattered ones.
- The hook skips the state update when the whole second has not changed for that consumer (`prev === next` guard), avoiding needless re-renders.

Public hook API is unchanged (`useElapsedTime(date) => number`). `subscribeToTick` is exported so the shared-timer lifecycle can be unit-tested without a DOM renderer.

## Tests

`sources/hooks/useElapsedTime.test.ts` (Vitest, fake timers): asserts a single interval is created for multiple subscribers, each fires per tick, removing one keeps the timer alive for the rest, and removing the last clears it. `pnpm typecheck` clean.

## Related

- Refs #1765 (main-thread cost on the web client)


f-liva:
Rebased onto current `main` (37e9479). Also dropped a commit that did not belong here: `40574cc0` ("search sessions by title in the home list") had been pushed onto this branch by mistake and is the subject of #1783. This PR is now the single ticker commit it claims to be.

PR #1778, opened 2026-09-10
fix(cli): stop cutting off codex turns that run past ten minutes

chphch:
`sendTurnAndWait` caps a codex turn at ten minutes of wall clock, so a turn that legitimately runs longer is declared aborted while Codex is still working on it. The abort is local — the timer resolves the pending promise and returns, and never sends `turn/interrupt` — so Codex keeps streaming into a client that has already moved on: `runCodex`'s `finally` clears `thinking`, the app drops the session's activity indicator, the "ready" push fires, and the message loop returns to `Waiting for messages...`, where the next prompt starts a second turn on top of the first. This makes the timer an **idle window** rather than a cap on turn length: every message the app-server sends re-arms it, and a turn blocked on an approval nobody has answered yet does not count as idle either, so it now fires only when Codex has genuinely gone quiet.

From a session on my machine (`codex-cli 0.153.4`, `happy codex --yolo`), a turn that was working normally the whole time:

```
14:15:31.957  turn/started
14:25:31.840  [WARN] [CodexAppServer] Turn timed out after 600000ms — treating as abort
              … 1317 further app-server notifications arrive …
14:42:12.418  turn/completed        ← the turn actually finished, 26m40s in
```

The longest silence anywhere inside that turn was **61 seconds** — command output streams back as `item/commandExecution/outputDelta`, so a working turn is never quiet for long. A ten-minute idle window keeps it alive with an order of magnitude to spare, while still catching the wedged-turn case the timeout was added for.

## Proof

Two live runs against a standalone `happy-server` + Expo web + a paired `happy-cli` daemon, driving real `codex app-server`. Same prompt both times (a shell loop that ticks for ~2.5 min); the only difference is whether activity re-arms the timer. `TURN_IDLE_TIMEOUT_MS` was shortened to 45 s for the capture so the run takes 3 minutes instead of 25 — same timer, same code path.

before and after

Left is the current behaviour: the session's status goes back to `online` at t=50 s while Codex runs on until t≈171 s. Right is this branch: it stays on the working indicator until the turn really ends.

The CLI logs agree:

```
before   thinking started 15:15:42            after   thinking started 15:19:45
         Turn idle for 45000ms — abort  (+45s)         thinking completed 15:22:37  (+171s)
         turn/completed 15:18:33       (+171s)         idle-timer firings: 0
         (no "thinking completed" — the finally block clears it silently)
```

Two regression tests cover both directions, and both fail on `main`: a turn that keeps producing output survives several idle windows, and a turn that goes silent for a whole window is still aborted. A third covers the approval case — a turn waiting on an unanswered approval is silent by nature, and must not be timed out on the user.


chphch:
Closing — 21a902a0b removes the ten-minute cutoff outright, which covers the bug this PR was opened for. The idle-window design here is no longer needed on top of that. Thanks for taking it.

PR #1784, opened 2026-09-11
feat(resume): resume sessions the daemon never tracked

f-liva:
## Problem

Resume is refused for any session the daemon does not still hold in memory:

```
Session cmsyan259b70xp414ozo9tawc is not tracked by this daemon.
It may have been started before the daemon or on another machine.
```

`findTrackedSessionById` consults `pidToTrackedSession` and `sessionIdToFinishedSession` only. A session started before the daemon — or before its last restart, when no encryption record made it to disk — is simply gone, and restarting the daemon does not bring it back.

There is a second, independent blocker underneath: `fetchServerSessionMetadata` reads `GET /v1/sessions`, which is hardcoded to `take: 150` (`sessionRoutes.ts`). On an account with ~800 sessions an older session is not in that page, so even a daemon that remembered the session could not refresh its metadata.

## Why the daemon cannot fix this alone

Reattaching needs the per-session data encryption key, to pass as `HAPPY_RECONNECT_ENCRYPTION_KEY`. `~/.happy/access.key` holds `encryption.publicKey`, `encryption.machineKey` and `token` — no account master secret. The daemon can encrypt *to* the account but cannot decrypt a session key it did not itself generate.

The client can: it holds the account secret, and it has already decrypted both the key and the metadata to render the session.

## Change

The app attaches the session's state to the `resume-happy-session` RPC:

```ts
{ sessionId, model, permissionMode, fallbackReason,
  fallback: { metadata, metadataVersion, agentStateVersion, seq, encryptionKey, encryptionVariant: 'dataKey' } }
```

- `Encryption` keeps the raw per-session data key (it already decrypts it in `ingestSessions`) and exposes `getSessionDataKey`.
- The daemon prefers its tracked state when it has it and falls back to this payload otherwise, so behaviour for tracked sessions is unchanged.
- The client's key removes the dependency on the `take: 150` page for the one thing that page was needed for, so the second blocker disappears without touching the server.
- The payload is validated with zod at the RPC boundary; a malformed one is ignored rather than failing the call, so the tracked path can still succeed.

Two things surfaced while validating this end to end on a self-hosted instance, and are fixed in the follow-up commits:

**`fallback: undefined` does not survive `JSON.stringify`.** A client that could not build a fallback was therefore indistinguishable on the wire from a client too old to know about the field, and the daemon's error read the same in both cases — a failed resume could not be diagnosed without rebuilding the client. The client now always sends a short `fallbackReason` (`ok` / `client-has-no-session-row` / `client-has-no-data-key`), the daemon echoes it in the error, and when the field is absent the message says so. It is truncated to 64 chars and only ever echoed back, never trusted.

**The metadata refresh must key off the missing field, not off who supplied the metadata.** The agent session ID reaches metadata only once the agent reports it — for Claude, on the `SessionStart` hook — so the webhook snapshot taken at spawn never carries it, and a client row ingested before that update arrived has exactly the same gap. Skipping the refresh whenever a fallback was sent therefore skipped it in the case that needed it, and resume failed with `is missing its Claude session ID` for a session the server could describe fully. The refresh now runs whenever the agent session ID is absent, from either source; it stays best effort, since it reads the 150 most recent sessions and an older session simply keeps the metadata the client sent.

## Security

`apiSocket.machineRPC` encrypts params with the machine encryption key, which the daemon already holds — the session key travels no further than the machine that is about to use it.

Legacy sessions (no per-session `dataEncryptionKey`, encrypted with the account master secret) are deliberately **not** covered: that secret must never leave the client, so they stay resumable only while tracked. They get an explicit error saying so.

## Tests

`packages/happy-cli/src/api/apiMachine.resumeFallback.test.ts` — a well-formed fallback reaches the resume handler; a malformed one is dropped instead of failing the call.

`pnpm typecheck` clean in `happy-app` and `happy-cli`; existing `apiMachine`, `daemon`, `sessionEnvironment` and `ops.codexFork` suites pass.

## Validated end to end

On a self-hosted instance with webapp and daemon both built from this branch:

- a session from weeks ago, never seen by the current daemon and well past the 150-row boot fetch, resumes;
- a session whose metadata had no `claudeSessionId` yet resumes too, after the refresh fix.

```
[SPAWN HAPPY CLI] Spawning: happy claude --happy-starting-mode remote --started-by daemon --resume a659f785-… --permission-mode bypassPermissions
[RPC] Handler returned { method: '…:resume-happy-session', hasResult: true }
```


f-liva:
Validated end-to-end on a self-hosted instance (webapp + daemon both built from this branch): a session started months ago, never seen by the current daemon and well past the 150-row boot fetch, now resumes. Daemon log on success:

```
[SPAWN HAPPY CLI] Spawning: happy claude --happy-starting-mode remote --started-by daemon --resume a659f785-… --permission-mode bypassPermissions
[RPC] Handler returned { method: '…:resume-happy-session', hasResult: true }
```

Pushed one follow-up commit while validating. `fallback: undefined` does not survive `JSON.stringify`, so a client that could not build a fallback was indistinguishable on the wire from a client too old to know about the field — the daemon's error read the same in both cases, which made a failed resume impossible to diagnose without rebuilding the client. The client now always sends a short `fallbackReason` (`ok` / `client-has-no-session-row` / `client-has-no-data-key`) and the daemon echoes it in the error; when the field is absent the message says so explicitly. It is truncated to 64 chars and only ever echoed back, never trusted.

PR #1788, opened 2026-09-14
feat(server): add encrypted session avatar transport

ex3ndr:
## Summary

Server feature only, stacked on migration PR #1790. This diff excludes the database schema, migration, and migration test.

Expose optional encrypted avatar descriptors and avatarVersion on session snapshots and change events. Add owner-authenticated upload, activation, download, and removal routes using existing local/S3 conventions. Preserve the removal revision even when avatar is null, preventing delayed events from restoring removed artwork.

Keep new session-avatar files private through the public files route and clean their blobs when the session is deleted. Retired/incomplete blobs live until session deletion to avoid racing concurrent uploads.

## Why sessions, and why similar code

Conversations can have their own pictures; project sessions leave theirs unset and mobile inherits project art. A fake project per bot would mislead old clients into offering project creation actions. Bot file access does not require that wrapper, and a future direct Agent API can use existing bot images.

Small resource checks and request/error copy follow projects and stay local so messages identify the right resource. Encryption is reused. The entire transport is not a trivial copy: session events, immutable uploads, bounded rate-limit state and deletion races differ. We avoid introducing a generic callback route framework in this change.

The session descriptor, avatar revision and account event sequence commit atomically. Serializable contention deliberately uses the existing bounded transaction retries. The static-file guard is scoped to new session avatars, not an unrelated rewrite of project/attachment transport.

## Rollout and review

1. Review/apply #1790 first.
2. Retarget this PR to main after that migration lands, then deploy the server feature.
3. Enable the separate Happy Agent producer.

Mobile #1789 can land before both. Only that UI PR is authorized to merge now; leave this PR and migration open.

## Verification

Server production typecheck and 24 focused tests passed, including the companion migration regression. Covers local/S3 lifecycle, account isolation, immutable uploads, public-path containment, host fallback, event projection and removal revisions. Opus 5 extra-high reviewed all three implementations; the missing-host fallback finding is fixed with a deterministic test. No production migration or deployment was run.

Full endpoint/encryption and design notes: packages/happy-server/sources/app/api/routes/SESSION_AVATARS.md.


ex3ndr:
Production verification completed against https://api.cluster-fluster.com on 2026-09-16.

The new route first appeared during a mixed-version rollout at 03:58 UTC. By 03:58:54 UTC the direct API smoke test passed using one newly generated temporary account and an encrypted test session:

- Session snapshots expose avatar:null and avatarVersion:0 before upload.
- Unauthenticated avatar access returns 401.
- Upload grant returns 200; encrypted S3 upload returns 204.
- Activation and identical reactivation return 200, retaining revision 1.
- Session snapshots match the activated descriptor.
- Download grant and blob download succeed; ciphertext and decrypted PNG bytes match exactly.
- Removal and repeated removal return 200, retaining null artwork at revision 2; subsequent download requests return 404.
- Health, root, account profile and session list stayed 200 throughout sampled checks.
- A follow-up at 03:59:46 UTC still exposed revision 2 and healthy basic endpoints.

The temporary session was deleted and its absence confirmed. An empty temporary account remains because the public API has no account-deletion endpoint; credentials were kept only in process memory and were discarded at exit. No existing user data was changed.

Before merging, corrected the migration test to use a CommonJS-compatible path without changing the SQL. Full server typecheck and 117 tests passed; the merged-main Server workflow also passed all jobs, including production-container verification.

PR #1800, opened 2026-09-17
Project checkouts with chat tabs and instant new chats

Scoteezy:
Two changes, one per commit.

### `feat(app)`: project checkouts with chat tabs and instant new chats

The grouped home list now lists **checkouts** — a project's own checkout and each of its worktrees — instead of individual chats. A checkout's chats open as tabs inside the session screen, the way the desktop keeps them.

- `ProjectHomeList` / `projectHomeList.ts` — build the checkout hierarchy. Happy Agent reports project and worktree identity directly; CLI sessions get the same shape derived from their machine and repository path.
- `WorktreeTabStrip` / `worktreeTabs.ts` — the tab strip for a checkout, and the rule for what it draws while a chat is being started.
- `pendingChats.ts` / `PendingChatView` / `pendingChatHandover.ts` — starting a session is a round trip to a machine that takes seconds. The strip's `+` puts the tab and the screen on screen immediately and lets the machine's answer catch up underneath. Memory-only, not persisted.
  - The de-duplication is by identity, not timestamps: the server's `new-session` push beats the daemon's reply, so a stand-in is matched against the set of chats that were in the checkout when `+` was pressed. This is what stops the same new chat being drawn twice and then one of them vanishing.
- `newSessionCheckout.ts` — "another chat exactly here": the new chat inherits the *chat's* directory, not the project's, plus its model / permission / effort picks. Happy Agent is handed the catalog project or workspace rather than a path, so it doesn't mint a duplicate project.
- `useComposerModes.ts` — extracted the composer's permission/model/effort resolution, which has to look in three places (session, agent metadata, account default).
- Session actions menus gained archive and a title, and are now reachable from the tab strip.

### `fix(server)`: 409 when a machine id belongs to another account

A machine id is the primary key of the whole `machine` table, not of an account's slice of it, so `POST /machines` could not create an id another account had already registered — and daemons do carry such ids, e.g. one re-paired to a second account keeps the identity it minted under the first. That surfaced as a 500 and an endless retry.

It now answers `409 machine_id_taken`, which is a permanent answer and lets the daemon mint a fresh id instead of retrying a request that can never succeed.

### Testing

- `packages/happy-app`: `tsc --noEmit` clean, 132 test files / 1427 tests pass (new: `pendingChats.test.ts`, `worktreeTabs.test.ts`, `projectHomeList.test.ts`, `newSessionCheckout.test.ts`, `agentSessionPlaces.spec.ts`, extended `useStartSessionFromDraft.test.ts`).
- `packages/happy-server`: `tsc --noEmit` clean, `machinesRoutes.spec.ts` 5/5 pass.


bra1nDump:
Merged by hand rather than through the merge button, so this shows as closed rather than merged. It is on `main` as:

- 215e4c0d `feat(app): open project checkouts with chat tabs and instant new chats`
- 8ce230cb `fix(server): answer 409 when a machine id belongs to another account`
- d8226888 `fix(app): keep pending chat text and show the archive in the project list`

Your authorship is preserved on all three.

What I changed on top:

- Dropped the `+` from the Projects section header. Workspace creation stays on the project card, where the `+` names what it acts on.
- Compared `projectAvatarThumbhash` when diffing project entries, so a decoded avatar repaints instead of holding the placeholder.
- Bot rows now read the session avatar instead of project artwork, so bots show their faces.
- Resolved conflicts in `useStartSessionFromDraft` against the bot-creation composer that landed on main in the meantime.

One gap still open if you want it: subtasks (side chats) do not show up anywhere in the project hierarchy.

Thank you for contributing!

